PR #12, opened 2022-11-11
Re-extract config.json to get correctly bundled LuaScript

Buhallin:
An error in TTSModManager caused the global LuaScript to not unbundle.  Update with the new extraction.


argonui (approved):

Chr1Z93 (approved):
Looks good to me!

PR #15, opened 2022-11-11
Clean up a -0 value

Buhallin:
Manager should handle these correctly from here on, but clean up this instance directly


Chr1Z93 (approved):
Approved

PR #17, opened 2022-11-12
updated doom counters

Chr1Z93:
Clean up for doom counters with new features:
- main counter ("Agenda Deck") gets a reset button
- smaller counter now automatically counts the doom in play (on playermats and big playarea), doom can be excluded via tag for cards like "Elle Rubash"

### DIFF
--- a/src/core/AgendaDeck.ttslua
+++ b/src/core/AgendaDeck.ttslua
@@ -1,132 +1,62 @@
-MIN_VALUE = -99
-MAX_VALUE = 999
+-- Doom Counter with Print
+-- original by:     -
+-- changed by:      Chr1Z
+-- description:     Clickable counter for doom on the agenda, changing the value prints the new value, reset button added
+information = {
+    version = "1.4",
+    last_updated = "12.11.2022"
+}
 
-function onload(saved_data)
-    light_mode = false
-    val = 0
+function onSave() return JSON.encode(val) end
 
+function onLoad(saved_data)
     if saved_data ~= "" then
-        local loaded_data = JSON.decode(saved_data)
-        light_mode = loaded_data[1]
-        val = loaded_data[2]
-    end
-
-    createAll()
-end
-
-function updateSave()
-    local data_to_save = {light_mode, val}
-    saved_data = JSON.encode(data_to_save)
-    self.script_state = saved_data
-end
-
-function createAll()
-    s_color = {0.5, 0.5, 0.5, 95}
-
-    if light_mode then
-        f_color = {1,1,1,95}
+        val = JSON.decode(saved_data)
     else
-        f_color = {0,0,0,100}
+        val = 0
     end
-
-
 
     self.createButton({
-      label=tostring(val),
-      click_function="add_subtract",
-      function_owner=self,
-      position={0,0.05,0},
-      height=600,
-      width=1000,
-      alignment = 3,
-      scale={x=1.5, y=1.5, z=1.5},
-      font_size=600,
-      font_color=f_color,
-      color={0,0,0,0}
-      })
+        label          = tostring(val),
+        click_function = "add_subtract",
+        function_owner = self,
+        position       = { 0, 0.06, 0 },
+        height         = 800,
+        width          = 800,
+        font_size      = 650,
+        scale          = { 1.5, 1.5, 1.5 },
+        font_color     = { 1, 1, 1, 95 },
+        color          = { 0, 0, 0, 0 }
+    })
 
+    self.createButton({
+        label          = "Reset",
+        click_function = "setToZero",
+        function_owner = self,
+        position       = { 0, -0.04, 2.7 },
+        height         = 600,
+        width          = 1250,
+        font_size      = 425
+    })
 
-
-
-    if light_mode then
-        lightButtonText = "[ Set dark ]"
-    else
-        lightButtonText = "[ Set light ]"
-    end
-
+    self.addContextMenuItem("More Information", function()
+        printToAll("------------------------------", "White")
+        printToAll("Doom Counter v" .. information["version"] .. " by Chr1Z", "Orange")
+        printToAll("last updated: " .. information["last_updated"], "White")
+    end)
 end
 
-function removeAll()
-    self.removeInput(0)
-    self.removeInput(1)
-    self.removeButton(0)
-    self.removeButton(1)
-    self.removeButton(2)
-end
-
-function reloadAll()
-    removeAll()
-    createAll()
-
-    updateSave()
-end
-
-function swap_fcolor(_obj, _color, alt_click)
-    light_mode = not light_mode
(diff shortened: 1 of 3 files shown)


src/core/AgendaDeck.ttslua
```diff
@@ -1,132 +1,62 @@
...
 
 function add_subtract(_obj, _color, alt_click)
-    mod = alt_click and -1 or 1
-    new_value = math.min(math.max(val + mod, MIN_VALUE), MAX_VALUE)
-    if val ~= new_value then
-        val = new_value
-      updateVal()
-        updateSave()
-    end
+    new_value = math.min(math.max(val + (alt_click and -1 or 1), 0), 99)
```
Buhallin, line 52: Make this local

src/core/AgendaDeck.ttslua
```diff
@@ -1,132 +1,62 @@
...
-
-    if light_mode then
-        lightButtonText = "[ Set dark ]"
-    else
-        lightButtonText = "[ Set light ]"
-    end
-
-end
+        label          = tostring(val),
+        click_function = "add_subtract",
```
Buhallin, line 21: Nit: Please rename so the style is consistent with other methods (i.e. addSubtract).  I'd probably rename it to addOrSubtract as well

src/core/AgendaDeck.ttslua
```diff
@@ -1,132 +1,62 @@
...
+    end)
 end
 
-function editName(_obj, _string, value)
-    self.setName(value)
-    setTooltips()
-end
+function setToZero() updateVal(0) end
 
 function add_subtract(_obj, _color, alt_click)
```
Buhallin, line 51: Nit/cleanup: Remove leading _ from variables

src/core/AgendaDeck.ttslua
```diff
@@ -1,132 +1,62 @@
...
-        })
-end
-
-function null()
-end
-
-function keepSample(_obj, _string, value)
-    reloadAll()
-end
+function updateSave() end
```
Buhallin, line 62: Does this still need to be here?

src/core/DoomInPlayCounter.ttslua
```diff
@@ -1,132 +1,105 @@
...
+    last_updated = "12.11.2022"
+}
+
+-- common parameters
+local CAST_PARAMETERS        = {}
+CAST_PARAMETERS.direction    = { 0, 1, 0 }
+CAST_PARAMETERS.type         = 3
+CAST_PARAMETERS.max_distance = 0
+
+local ypos       = 1.6
```
Buhallin, line 15: This could use a more descriptive name.  What's it used for?  Same with zone below

src/core/DoomInPlayCounter.ttslua
```diff
@@ -1,132 +1,105 @@
...
+local zone       = getObjectFromGUID("a2f932")
+local doom_url   = "https://i.imgur.com/EoL7yaZ.png"
+local IGNORE_TAG = "DoomCounter_ignore"
```
Buhallin, line 18: Just so we're clear, you know nobody will use this, right? :D  IMHO still better off marking it somehow to show the number may not be accurate (or even turning it off entirely) if Elle is in play.  No action required, just a comment.

src/core/DoomInPlayCounter.ttslua
```diff
@@ -1,132 +1,105 @@
...
-end
+-- searches playermats (num = 1-4) or the scripting zone (num = 5)
+function search(num)
+    local val = 0
+    if num == 5 then
+        for _, obj in ipairs(zone.getObjects()) do
+            val = val + getResult(obj)
+        end
+    else
+        CAST_PARAMETERS.origin = originAndSize[num].origin
```
Buhallin, line 81: UPPER_SNAKE implies a constant, even if Lua doesn't enforce it changing the values is bad form (and can carry over).  Make a new local table that's initialized to the values from CAST_PARAMETERS, and change that as needed.

src/core/DoomInPlayCounter.ttslua
```diff
@@ -1,132 +1,105 @@
...
-        })
-end
-
-function null()
-end
-
-function keepSample(_obj, _string, value)
-    reloadAll()
+-- checks an object for the doom image and gets quantity (for stacks)
+function getResult(obj)
```
Buhallin, line 92: Give this a more descriptive name, maybe isDoom() or countDoom()

src/core/DoomInPlayCounter.ttslua
```diff
@@ -1,132 +1,105 @@
...
+    if (obj.is_face_down and obj.getCustomObject().image_bottom == doom_url) or
+        (obj.name == "Custom_Token" and obj.getCustomObject().image == doom_url) then
+        if not obj.hasTag(IGNORE_TAG) then
+            if removeDoom then
```
Buhallin, line 96: Relying on a global variable to change behavior is going to be error prone, especially with any kind of timed/background operation.  At a minimum it should be a parameter passed here, but better to make removal a separate method and call it when appropriate.

src/core/DoomInPlayCounter.ttslua
```diff
@@ -1,132 +1,105 @@
...
+                obj.destruct()
+                return 0
+            else
+                return math.abs(obj.getQuantity())
```
Buhallin, line 100: Does a single token return 1 from getQuantity()?  I thought you said at some point that this only worked for a stack?

Buhallin (changes requested):

src/core/AgendaDeck.ttslua
Chr1Z93: No, it can go

src/core/DoomInPlayCounter.ttslua
Chr1Z93: Fair point, will think about a different approach for the next update

src/core/DoomInPlayCounter.ttslua
Chr1Z93: Yep, it does return 1 for singles, maybe I was wrong in the past?

src/core/DoomInPlayCounter.ttslua
Chr1Z93: changed capitalization

src/core/DoomInPlayCounter.ttslua
Chr1Z93: loop will be stopped before setting the global variable, then it will be called once and only afterwards the loop will be restarted

Chr1Z93:
commit [6b7eba5](https://github.com/argonui/SCED/pull/17/commits/6b7eba56a52c8f25866754f0ef4010a6459421f3) resolves the comments from @Buhallin

Buhallin (approved):

PR #18, opened 2022-11-12
Extract card spawning methods from the loader

Buhallin:
This refactors the current ArkhamDB loader by moving the card spawning operations to a standalone file which can be included in other objects.

Card list build methods are appropriately updated as well.

### DIFF
--- a/src/arkhamdb/DeckImporterMain.ttslua
+++ b/src/arkhamdb/DeckImporterMain.ttslua
@@ -423,7 +423,7 @@ function loadCards(slots, investigatorId, playerColor, commandManager, configura
           zoneCards,
           deckPos,
           Zones.getDefaultCardRotation(playerColor, zone),
-          true,
+          true, -- Sort deck
           spreadCallback)
 
       coroutine.yield(0)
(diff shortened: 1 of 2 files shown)


src/arkhamdb/DeckImporterMain.ttslua
```diff
@@ -399,50 +398,34 @@ function loadCards(slots, investigatorId, playerColor, commandManager, configura
...
-        rotation = Zones.getDefaultCardRotation(playerColor, zone),
-       callback_function = spreadCallback
-      })
+      Spawner.spawnCards(
+          zoneCards,
+          deckPos,
+          Zones.getDefaultCardRotation(playerColor, zone),
+          true,
+          spreadCallback)
+
```
Chr1Z93, line 428: comment with description of arguments might be useful (so you don't need to open the required file when sth needs to change)

Chr1Z93 (approved):
Looks fine to me. Also loaded it ingame and didn't find an issue - gonna trust you that edge cases are still getting handled correctly :P

src/arkhamdb/DeckImporterMain.ttslua
Buhallin: Added for the sort, but I think the names for the rest of them are clear (cardList, pos, rotation, callback)

PR #20, opened 2022-11-12
updated playarea

Chr1Z93:
- code restructured
- legacy code removed
- comments added
- clue counts above 12 spawn without overlap

### DIFF
--- a/src/core/PlayArea.ttslua
+++ b/src/core/PlayArea.ttslua
@@ -1,80 +1,62 @@
+---------------------------------------------------------
+-- general setup
+---------------------------------------------------------
+
 -- set true to enable debug logging
 DEBUG = false
 
--- we use this to turn off collision handling (for clue spawning)
--- until after load is complete (probably a better way to do this)
+-- we use this to turn off collision handling until onLoad() is complete
 COLLISION_ENABLED = false
 
--- TODO get the log function from global instead
--- log = Global.call('getLogFunction', this)
-function getLogFunction(object)
-  return function (message)
-    if DEBUG then
-      print(message)
-    end
-  end
-end
+local COUNTER = getObjectFromGUID('f182ee')
+local TOKEN_DATA = Global.getTable('TOKEN_DATA')
+local clueData = {
+  thickness    = 0.1,
+  stackable    = true,
+  type         = 2,
+  image        = TOKEN_DATA.clue.image,
+  image_bottom = TOKEN_DATA.doom.image
+}
 
-log = getLogFunction(self)
+---------------------------------------------------------
+-- general code
+---------------------------------------------------------
 
-function onload(save_state)
-  self.interactable = DEBUG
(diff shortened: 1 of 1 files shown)


src/core/PlayArea.ttslua
```diff
@@ -85,130 +67,54 @@ function getClueCount(object, isFaceDown, playerCount)
...
-      }
-    else
-      obj_parameters.position = {
-        collision_info.collision_object.getPosition()[1] + 0.575,
-        collision_info.collision_object.getPosition()[2] + 0.4,
-        collision_info.collision_object.getPosition()[3] - 5.2 + (0.55 * i)}
-    end
-    spawnToken(obj_parameters.position, '3')
-    i = i + 1
+  SPAWNED_LOCATION_GUIDS[object.getGUID()] = 1
```
Buhallin, line 79: Not a mandatory change, but for cases like this I prefer to use a boolean (true) rather than the number (1).  It's basically irrelevant, we're only using it as a Set but it's a bit more correct semantically and the tiniest bit more efficient.

src/core/PlayArea.ttslua
```diff
@@ -85,130 +67,54 @@ function getClueCount(object, isFaceDown, playerCount)
...
+
+  -- spawn clues (starting top right, moving to the next row after 4 clues)
+  local pos = object.getPosition()
+  for i = 1, clueCount do
+    local row = math.floor(1 + (i - 1) / 4)
+    local column = (i - 1) % 4
+    spawnClues({ pos[1] + 1.5 - 0.55 * row, pos[2], pos[3] - 0.825 + 0.55 * column })
```
Buhallin, line 86: pos should have .x, .y, .z values, much clearer to use those instead of the number index

src/core/PlayArea.ttslua
```diff
@@ -85,130 +67,54 @@ function getClueCount(object, isFaceDown, playerCount)
...
   end
 end
 
+function spawnClues(position)
```
Buhallin, line 90: Just spawns one, right?  Rename to spawnClue

src/core/PlayArea.ttslua
```diff
@@ -85,130 +67,54 @@ function getClueCount(object, isFaceDown, playerCount)
...
+  local token = spawnObject({
+    position = position,
+    rotation = { 3.88, -90, 0.24 },
+    type     = 'Custom_Tile'
+  })
+
+  token.setCustomObject(clueData)
```
Buhallin, line 97: I think all of these can be put in the clueData table and use spawnObjectData (not 100% sure there because of Custom_Tile but should work)

Buhallin (changes requested):

src/core/PlayArea.ttslua
Chr1Z93: I am not sure about that one, so I will leave this as is for now

Buhallin (approved):

PR #21, opened 2022-11-12
updated playermats and added InvestigatorSkillTracker

Chr1Z93:
WARNING: very long PR

Features:

StatTracker:
- new tracker for stats instead of the 4 single tokens
image

Playermats:
- code restructuring
- onCollisionEnter for investigators: load base statsStats for statTracker, set action tokens to class
- onCollisionEnter for cards: only spawn uses in actual playarea, not in threat area
- load additional uses for customizable cards if upgradesheets is present in set aside area
- Upkeep: replenish tokens for cards with respective metadata
- uses spawn now rotated for the playmat
- playercards will be discarded to the local discard pile if they have the respective tag


argonui (commented):
No idea about the content. But for a change this big I would recommend leaving it in its own feature branch and letting folks make a different mod if they want to try it out test it

src/playermat/Playmat.ttslua
```diff
@@ -93,293 +70,271 @@ function onload(save_state)
...
+    if obj.tag == "Card" and not obj.is_face_down and not doNotReady(obj) then
+      local notes = JSON.decode(obj.getGMNotes()) or {}
+      if notes.id == "08031" then
+        forcedLearning = true
+      elseif notes.type ~= "Investigator" then
+        obj.setRotation(PLAY_ZONE_ROTATION)
+
+        -- check for cards with 'replenish' in their metadata
+        if notes.uses ~= nil then
+          local count = notes.uses[1].count
```
Buhallin, line 204: Can we make this more robust, specifically by handling the multiple uses?  None of them have replenish right now but it's possible, would be nice to future-proof it

src/playermat/Playmat.ttslua
```diff
@@ -93,293 +70,271 @@ function onload(save_state)
...
-        drawCards(numCards)
-        return
+  -- flip investigator mini-card if found
+  -- flip summoned servitor mini-cards (To-Do: don't flip all of them)
+  if activeInvestigatorId ~= nil then
+    local miniId = string.match(activeInvestigatorId, "%d%d%d%d%d") .. "-m"
+    for _, obj in ipairs(getObjects()) do
+      if obj.tag == "Card" then
+        local notes = JSON.decode(obj.getGMNotes())
+        if obj.is_face_down and notes ~= nil and notes.type == "Minicard" and (notes.id == miniId or notes.id == "09080-m") then
```
Buhallin, line 221: Can move the obj.is_face_down to the same check with tag == Card, will avoid the JSON decode time if the object wouldn't flip anyway

src/playermat/Playmat.ttslua
```diff
@@ -93,293 +70,271 @@ function onload(save_state)
...
-
-function shuffleDiscardIntoDeck()
-    if not discardPile.is_face_down then discardPile.flip() end
-    discardPile.shuffle()
-    discardPile.setPositionSmooth(DRAW_DECK_POSITION, false, false)
-    drawDeck = discardPile
-    discardPile = nil
-end
+  -- gain an additional resource for Jenny Barnes
+  if string.match(activeInvestigatorId, "%d%d%d%d%d") == "02003" then
```
Buhallin, line 232: Need to check for Jenny's promo ID as well

src/playermat/Playmat.ttslua
```diff
@@ -93,293 +70,271 @@ function onload(save_state)
...
+    RESOURCE_COUNTER.call("add_subtract")
+    printToColor("Gaining 2 resources (Jenny)", messageColor)
+  end
 
-function patriceDraw()
+  -- draw a card (with handling for Patrice and Forced Learning)
+  if activeInvestigatorId == "06005" then
```
Buhallin, line 238: Unlikely to happen, but Patrice can actually take Forced Learning.  See the ArkhamDB page for the FAQ - probably better to do nothing if both are present

src/playermat/Playmat.ttslua
```diff
@@ -93,293 +70,271 @@ function onload(save_state)
...
-      type = 'ScriptingTrigger',
-      callback = 'zoneCallback',
-      callback_owner = self,
-      rotation = self.getRotation()
-    }
-    spawnObject(zoneProps)
-end
+-- draw X cards (shuffle discards if necessary)
+function drawCardsWithReshuffle(numCards)
+  if type(numCards) ~= "number" then numCards = 1 end
```
Buhallin, line 260: This feels like overkill, probably safe to remove?

src/playermat/Playmat.ttslua
```diff
@@ -93,293 +70,271 @@ function onload(save_state)
...
-        debug=false --If it displays the sphere when casting.
-    })
-    local decksAndCards = {}
-        for _, obj in ipairs(objList) do
-        if obj.hit_object.tag == "Deck" or obj.hit_object.tag == "Card" then
-            table.insert(decksAndCards, obj.hit_object)
-        end
+-- get the draw deck and discard pile objects
+function getDrawDiscardDecks()
+  drawDeck = nil
```
Buhallin, line 306: Declare these at the top, or better yet make them local and return a table {drawDeck = ..., discardDeck = ...} (assuming you can put object refs in a table, I don't think I've ever tried that.

src/playermat/Playmat.ttslua
```diff
@@ -424,32 +409,54 @@ end
...
+    for _, useInfo in ipairs(cardMetadata.uses) do
       type = useInfo.type
+      token = useInfo.token
       tokenCount = useInfo.count
-      if (activeInvestigatorId == "03004" and useInfo.type == "Charge") then
-        tokenCount = tokenCount + 1
+    end
+
+    -- additional uses for certain customizable cards (by checking the upgradesheets)
+    if customizationsTable[cardMetadata.id] ~= nil then
```
Buhallin, line 437: Let's make sure we remember this is here when we get around to pushing the actual metadata changes

src/playermat/Playmat.ttslua
```diff
@@ -424,32 +409,54 @@ end
...
+      for _, obj in ipairs(searchArea(PLAY_ZONE_POSITION, PLAY_ZONE_SCALE)) do
+        local obj = obj.hit_object
+        if obj.tag == "Card" then
+          local notes = JSON.decode(obj.getGMNotes()) or {}
+          if notes.id == (cardMetadata.id .. "-c") then
+            local pos = customizationsTable[cardMetadata.id][1]
+            local boxes = customizationsTable[cardMetadata.id][2]
+            if obj.getVar("markedBoxes")[pos] == boxes then tokenCount = tokenCount + 2 end
```
Buhallin, line 445: We should try to avoid getVar() whenever possible.  It exposes the internal structure of an object, which makes changing that very problematic.  Instead, create a method in the target object and use .call().  That creates a layer of indirection where you can update the internals.

Wouldn't do it now, PR is already massive as it is, but consider for a near-future enhancement.

src/playermat/Playmat.ttslua
```diff
@@ -424,32 +409,54 @@ end
...
+    if activeInvestigatorId == "03004" and type == "Charge" then tokenCount = tokenCount + 1 end
+
+    log("Spawning tokens for "..object.getName()..'['..object.getDescription()..']: '..tokenCount.."x "..token)
+    spawnTokenGroup(object, token, tokenCount)
   else
     local data = getPlayerCardData(object)
-    if data == nil then
-      error('attempt to spawn tokens for ' .. object.getName() .. ': no token data')
-    end
+    if data == nil then error('attempt to spawn tokens for ' .. object.getName() .. ': no token data') end
```
Buhallin, line 459: nil is expected if there's no tokens for the card, right?  Or am I misreading that?

src/playermat/Playmat.ttslua
```diff
@@ -484,68 +488,102 @@ function unmarkSpawned(guid, force)
...
+      ["Rogue"]    = 3,
+      ["Mystic"]   = 4,
+      ["Survivor"] = 5,
+      ["Neutral"]  = 6
+    }
+    
+    for _, obj in ipairs(search) do
+      local obj = obj.hit_object
+      if obj.getDescription() == "Action Token" and obj.getStateId() > 0 then
+        if self.positionToLocal(obj.getPosition()).x > -0.95 then
```
Buhallin, line 533: Checking this based on position feels fragile.  Use scale instead?

src/playermat/Playmat.ttslua
```diff
@@ -484,68 +488,102 @@ function unmarkSpawned(guid, force)
...
-function updateStatToken(tokenGuid, val)
-  local statToken = getObjectFromGUID(tokenGuid)
-  if (statToken == nil) then
-    return
-  end
-  statToken.call("reset_val")
-  for i = 1, val do
-    statToken.call("add_subtract", { alt_click = false })
-  end
+function objState(obj, stateId)
```
Buhallin, line 563: setActionTokenState (since that's what it's used for) or setObjectState at a minimum

src/playermat/Playmat.ttslua
```diff
@@ -484,68 +488,102 @@ function unmarkSpawned(guid, force)
...
+        end
+      end
+    end
+
+    -- update the small token with special action for certain investigators
+    -- Ursula Downs: Investigate action
+    if activeInvestigatorId == "04002" then
+      objState(small_token, 8)
+    -- Daisy Walker (only for normal front, not parallel): Tome action
+    elseif activeInvestigatorId == "01002" or activeInvestigatorId == "01002-r" or activeInvestigatorId == "01002-pb" then
```
Buhallin, line 546: 01002-r changed to 01502 in the recent metadata update

src/playermat/Playmat.ttslua
```diff
@@ -484,68 +488,102 @@ function unmarkSpawned(guid, force)
...
+function drawEncountercard(_, _, isRightClick)
+  Global.call("drawEncountercard", {self.positionToWorld(DRAWN_ENCOUNTER_CARD_OFFSET), self.getRotation(), isRightClick})
 end
 
 function spawnToken(position, tokenType)
-  Global.call('spawnToken', {position, tokenType})
+  Global.call('spawnToken', {position, tokenType, PLAY_ZONE_ROTATION})
 end
 
 function updatePlayerCards(args)
```
Buhallin, line 583: Document the method, and add what the expected field/value is in the args table

Buhallin (changes requested):

src/playermat/statTracker.ttslua
```diff
@@ -0,0 +1,76 @@
...
+        printToAll("Stat Tracker v" .. information["version"] .. " by Chr1Z", "Orange")
+        printToAll("last updated: " .. information["last_updated"], "White")
+    end)
+    self.addContextMenuItem("Reset to 1s", function() updateStats({ 1, 1, 1, 1 }) end)
+end
+
+-- helper function to carry index
+function attachIndex(click_function, index)
+    local fn_name = click_function .. index
+    _G[fn_name] = function(obj, player_color, isRightClick)
```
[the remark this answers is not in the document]
Chr1Z93: I didn't come up with this snippet myself, but I am open for different suggestions!
Maybe this could be solved with local scope but I am not sure how that would look like.

src/playermat/Playmat.ttslua
```diff
@@ -1,90 +1,67 @@
...
+  {-1.365, 0.1, -0.945},
+  {-0.91, 0.1, -0.945},
+  {-0.455, 0.1, -0.945},
+  {0, 0.1, -0.945},
+  {0.455, 0.1, -0.945},
+  {0.91, 0.1, -0.945}
+}
+
+local PLAY_ZONE_ROTATION = self.getRotation()
+activeInvestigatorId = "00000"
```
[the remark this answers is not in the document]
Chr1Z93: Intentionally global so it can be reset by the cleanup helper

src/playermat/Playmat.ttslua
```diff
@@ -93,293 +70,271 @@ function onload(save_state)
...
-                local r = obj.getRotation()
-                if (r.y - y > 10) or (y - r.y > 10) then
-                    obj.setRotation(PLAY_ZONE_ROTATION)
-                end
-            end
-        elseif obj.getDescription() == "Action Token" then
-            if obj.is_face_down then obj.flip() end
-        end
-    end
+function doNotReady(card) return card.getVar("do_not_ready") or false end
```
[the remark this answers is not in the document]
Chr1Z93: We could flip that but this is how it is currently implemented and might be used among custom content (I didn't touch this function)

src/playermat/statTracker.ttslua
```diff
@@ -0,0 +1,76 @@
...
+        _G[click_function](obj, player_color, isRightClick, index)
```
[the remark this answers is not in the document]
Chr1Z93: Isn't this correctly indented? It's the function that starts in the line before that

src/playermat/Playmat.ttslua
```diff
@@ -93,293 +70,271 @@ function onload(save_state)
...
+  end
+  
+  local forcedLearning = false
+
+  -- unexhaust cards in play zone, flip action tokens and find forcedLearning
+  for _, v in ipairs(searchArea(PLAY_ZONE_POSITION, PLAY_ZONE_SCALE)) do
+    local obj = v.hit_object
+    if obj.tag == "Card" and not obj.is_face_down and not doNotReady(obj) then
+      local notes = JSON.decode(obj.getGMNotes()) or {}
+      if notes.id == "08031" then
```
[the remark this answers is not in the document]
Chr1Z93: Already like it currently is and not an edge case that we should need to account for

src/playermat/Playmat.ttslua
Chr1Z93: I don't see the issue - isn't it "notes.uses" indepedant of the type of uses? Could you provide an example card (hypothetical) that wouldn't work?

src/playermat/Playmat.ttslua
Chr1Z93: Good catch, will change later

src/playermat/Playmat.ttslua
Chr1Z93: This is what the string.match does

src/playermat/Playmat.ttslua
Chr1Z93: Good idea, would need to update the upgradesheets for that (and also use require for them)

src/playermat/Playmat.ttslua
Chr1Z93: Yes, but then this function should not have been called (could get rid of this imo, but I left it there for now)

src/playermat/Playmat.ttslua
Chr1Z93: You are right, but I don't think scale would be less fragile 😅

src/playermat/Playmat.ttslua
Chr1Z93: Thanks for notifying, will change later

objects/StatTracker.af7ed7.ttslua
```diff
@@ -0,0 +1 @@
+require("playermat/statTracker")
```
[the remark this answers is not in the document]
Buhallin: Since the unbundled code is less than 80 characters, ModManager won't extract it to a file.  It'll leave it in the LuaScript field in the .json, causing a change to that file.  It'll also leave this file orphaned.

src/playermat/statTracker.ttslua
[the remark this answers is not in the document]
Buhallin: Ah, you're right...  but man that's some ugly code :P

objects/StatTracker.af7ed7.ttslua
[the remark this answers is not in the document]
Chr1Z93: Understood and will be changed

src/playermat/Playmat.ttslua
Chr1Z93: Will rename

src/playermat/Playmat.ttslua
Chr1Z93: will leave this as is for now since I don't wanna risk breaking anything that calls this

src/playermat/Playmat.ttslua
Chr1Z93: As long as the ID in the metadata doesn't get changed, this should not need any changes in the future
image

src/playermat/Playmat.ttslua
Chr1Z93: implementing scale for now

src/playermat/Playmat.ttslua
Buhallin: Uses is an array, this is only checking the first entry.  Currently Arcane Initiate (3) is the only card that has multiple use types, but if a second (or more) type had a replenish this would miss it

src/playermat/Playmat.ttslua
Buhallin: Ah, I see now.  Got it.

src/playermat/Playmat.ttslua
Buhallin: We'd talked about having the metadata actually changed based on the selected boxes.  If we do that, the modification here would duplicate it.  I think changing the metadata will be a more general solution, otherwise everywhere that uses metadata will have to have these sorts of checks.  But we can change this later when we implement that.

src/playermat/Playmat.ttslua
Chr1Z93: Ah, that makes sense. So we probably wanna loop through all the entries in "uses" and check if there is replenish in one of them?

Buhallin: Probably should just loop through and handle each uses independently.  Agree that it's unlikely, up to you if you want to fix it or not.  But since a lot of places are likely to (ahem) use uses, IMHO it's better to have them comprehensive so that when a new combination comes along we don't have to audit which (ahem) uses of uses handle which combinations.

Buhallin (approved):

src/playermat/Playmat.ttslua
Chr1Z93: Would prefer leaving it like this because this case is not possible atm (and at the very earliest relevant with the nex investigator expansion which will be at least 6 months away)

PR #25, opened 2022-11-12
 updated global script

Chr1Z93:
- restructured code
- major clean up
- legacy code removal
- enable spawning of uses on playermats rotated


src/core/Global.ttslua
```diff
@@ -456,358 +268,281 @@ function drawChaostoken(params)
...
-  if tokenData == nil then
-    error("no token data found for '" .. tokenType .. "'")
-  end
+  if tokenData == nil then error("no token data found for '" .. tokenType .. "'") end
 
   local token = spawnObject({
     type = 'Custom_Token',
     position = position,
-	rotation = {x=0, y=270, z=0}
+	  rotation = rotation
```
Buhallin, line 285: Nit: fix indentation

src/core/Global.ttslua
```diff
@@ -456,358 +268,281 @@ function drawChaostoken(params)
...
-function round(params) -- Parameter (int number, int numberDecimalPlaces)
-  return tonumber(string.format("%." .. (params[2] or 0) .. "f", params[1]))
-end
+---------------------------------------------------------
+-- chaos token stat tracker
+---------------------------------------------------------
 
-function roundposition(params) -- Parameter (Table position)
-  return {round({params[1], 2}),round({params[2], 2}),round({params[3], 2})}
+function take_callback(object_spawned, mat)
```
Buhallin, line 302: Rename to something more specific

src/core/Global.ttslua
```diff
@@ -456,358 +268,281 @@ function drawChaostoken(params)
...
+  local player = mat.getGUID()
+  local image = object_spawned.getCustomObject().image
+  PULLS[image]                = (PULLS[image] or 0) + 1
```
Buhallin, line 305: Declare at top, make local, and change name to not look like a constant

src/core/Global.ttslua
```diff
@@ -456,358 +268,281 @@ function drawChaostoken(params)
...
+  PLAYER_PULLS[player][image] = (PLAYER_PULLS[player][image] or 0) + 1
 end
 
-function isEqual(params) --Parameter (Table table1, Table table2) returns true if the tables are equal
-  if params[1][1] == params[2][1] and params[1][2] == params[2][2] and params[1][3] == params[2][3] then
-    return true
+function printOrResetStats(_, _, isRightClick)
```
Buhallin, line 309: Let's split this up for clarity:
function handleStatTrackerClick(_, _, isRightClick)
  if isRightClick then
    resetStatTracker()
  else
    printStats()
  end
end

Associated functions would take the blocks from the current if

src/core/Global.ttslua
```diff
@@ -456,358 +268,281 @@ function drawChaostoken(params)
...
+    if value ~= 0 then printToAll(IMAGE_TOKEN_MAP[key] .. ': ' .. tostring(value)) end
   end
 end
 
---Difficulty selector script
+---------------------------------------------------------
+-- Difficulty selector script
+---------------------------------------------------------
 
 function createSetupButtons(args)
```
Buhallin, line 356: Setup is very generic, maybe createBagSetupButtons?

Also good practice to document expected values for anything that just takes the single table parameter style for args

src/core/Global.ttslua
```diff
@@ -456,358 +268,281 @@ function drawChaostoken(params)
...
+      buttonParameters.label = "Standalone"
+      buttonParameters.click_function = "standaloneClick"
+      args.object.createButton(buttonParameters)
     end
+  end
 end
 
-function getObjectsInZone(zoneId)
-    local zoneObject = getObjectCache(zoneId)
+function fillContainer(args)
```
Buhallin, line 403: initializeChaosBag?

Especially in Global (which can and does have a wide variety of things it does) generic names have no context to work with.

At some point we should probably split these to independent files based on function, and require them...  but that's another day.

src/core/Global.ttslua
```diff
@@ -1,102 +1,78 @@
...
-
-function onObjectDrop(player, obj)
---  local mat = getObjectFromGUID("dsbd0ff4")
---  log(mat.positionToLocal(obj.getPosition()))
-end
-
-function take_callback(object_spawned, mat)
-    customObject = object_spawned.getCustomObject()
-    local player = mat.getGUID();
+IMAGE_TOKEN_MAP = {
```
Buhallin, line 37: Rename to CHAOS_TOKEN_IMAGE_MAP

src/core/Global.ttslua
```diff
@@ -456,358 +268,281 @@ function drawChaostoken(params)
...
+    })
+    obj.scale {0.81, 1, 0.81}
+    obj.setName(getTokenName({ url=url }))
+    return obj
+  end
+end
 
--- Content Importing
+-- chaos bag needs this!
+function getTokenName(params)
```
Buhallin, line 498: Rename to getChaosTokenName, add doc for paramaters

src/core/Global.ttslua
```diff
@@ -456,358 +268,281 @@ function drawChaostoken(params)
...
+  local name = IMAGE_TOKEN_MAP[params.url]
+    if name == nil then name = "" end
+  return name
+end
 
+function getImageUrl(id)
```
Buhallin, line 504: getChaosTokenImageUrl()

src/core/Global.ttslua
Chr1Z93: this function is called by all the difficulty selectors contained in all scenario setup bags, will not change that for now

src/core/Global.ttslua
Chr1Z93: same as above - not easily changeable because it is referenced by every difficulty selector

src/core/Global.ttslua
Chr1Z93: this is referenced by a lot of cards (mostly the ones that have a context menu with "sealing" options - not changing the name for now

src/core/Global.ttslua
Chr1Z93: sadly this is also referenced by some setup bags (like for example from the custom campaign "Jumanji")

src/core/Global.ttslua
Buhallin: That makes sense.  Can we at least enhance the method doc here then?  Describe (at least at a high level) what it does, where it's used, and what the parameters expected in the args table are?

Chr1Z93: Documentation added

Buhallin (approved):

PR #28, opened 2022-11-12
adding additional information to the README

Chr1Z93:


argonui (changes requested):
I think google docs should be where we plan what we say, and .md files in the repo should be where they live.

Buhallin (commented):
Let's hold off on this for now.  Several of these docs are very WIP and even contradict each other (the SCED in Source Control is based on the code-only approach).

Chr1Z93:
Understandable

PR #36, opened 2022-11-15
Player card updates, Part 1: JSON

Buhallin:
Part 1 of a mass refresh of card data.  This fixes some unicode issues with card names, and standardizes some of the JSON.

This touches a LOT of files, I've verified via loading cards from the card spreads.  Suggest testing rather than reading :)


Chr1Z93 (approved):
I trust you and will not check 683 cards

PR #39, opened 2022-11-15
Metadata update, Part 3

Buhallin:
Add Investigator tag to all investigator cards.  This will support tagging the snap points on playmats.

Set size of all investigators to 1.15.  This puts them over the threshold where TTS will combine them with a standard deck and resize them.


Chr1Z93 (approved):
Nice branch name

PR #41, opened 2022-11-15
Metadata update, Part 4

Buhallin:
Clean up unicode characters which were corrupted by an error in the metadata editor


Buhallin:
This was derived from the previous update, so includes that commit (which was merged without changes).  Charlie's tag and the Unicode should be the only changes.

Chr1Z93 (approved):
LGTM

PR #46, opened 2022-11-16
clearing the cache for the data helper

Chr1Z93:
script state of data helper cleared, because some cards were falsely marked as "true" (and because of that wouldn't get resources - e.g. the default brand of cthugha)


Buhallin (changes requested):
This will show up as a modified file on next decompose.  Move into the .json LuaScriptState field.

Chr1Z93:
@Buhallin moved the scriptstate into the JSON

Buhallin (approved):

PR #47, opened 2022-11-17
Rounding cleanup

Buhallin:
Looks like a lot of files got committed without the value rounding applied.  Not sure when or which PR, but this cleans it up.

This change was generated based on the current main by assembling then decomposing the mod.


Chr1Z93 (approved):
I think this isn't anyone's fault - we just actually didn't do a full bundle & decompose with the newest version of the mod manager yet

PR #48, opened 2022-11-17
Create SpawnBag to replace card memory bags

Buhallin:
This adds a new SpawnBag capability which will replace the various small memory bags around the table.

Initial implementation is for the Bonded container, which is no longer a bag.

### DIFF
--- a/src/playercards/PlayerCardSpawner.ttslua
+++ b/src/playercards/PlayerCardSpawner.ttslua
@@ -2,7 +2,7 @@
 -- Amount to shift for the next card (zShift) or next row of cards (xShift)
 -- Note that the table rotation is weird, and the X axis is vertical while the
 -- Z axis is horizontal
-local zShift = -2.29998
+local SPREAD_Z_SHIFT = -2.3
 
 Spawner = { }
 
@@ -53,7 +53,7 @@ Spawner.spawnCardSpread = function(cardList, startPos, rot, sort, callback)
   local position = { x = startPos.x, y = startPos.y, z = startPos.z }
   for _, card in ipairs(cardList) do
     Spawner.spawn({ card }, position, rot, callback)
-    position.z = position.z + zShift
+    position.z = position.z + SPREAD_Z_SHIFT
   end
 end
 
--- a/src/playercards/PlayerCardSpawner.ttslua
+++ b/src/playercards/PlayerCardSpawner.ttslua
@@ -1,3 +1,9 @@
+
+-- Amount to shift for the next card (zShift) or next row of cards (xShift)
+-- Note that the table rotation is weird, and the X axis is vertical while the
+-- Z axis is horizontal
+local zShift = -2.29998
+
 Spawner = { }
 
 -- Spawns a list of cards at the given position/rotation.  This will separate cards by size -
@@ -37,6 +43,18 @@ Spawner.spawnCards = function(cardList, pos, rot, sort, callback)
 
   position.y = position.y + (#standardCards + #miniCards) * 0.07
   Spawner.spawn(miniCards, position, rot, callback)
+end
+
+Spawner.spawnCardSpread = function(cardList, startPos, rot, sort, callback)
+  if (sort) then
+    table.sort(cardList, Spawner.cardComparator)
+  end
+
+  local position = { x = startPos.x, y = startPos.y, z = startPos.z }
+  for _, card in ipairs(cardList) do
+    Spawner.spawn({ card }, position, rot, callback)
+    position.z = position.z + zShift
+  end
 end
 
 -- Spawn a specific list of cards.  This method is for internal use and should not be called
--- a/src/playercards/spawnbag/BondedBag.ttslua
+++ b/src/playercards/spawnbag/BondedBag.ttslua
@@ -1,0 +1,63 @@
+require("playercards/spawnbag/SpawnBag")
+
+SPAWN_SPEC = {
+  name = "BondedCards",
+  cards = {
+    "05314", -- Soothing Melody
+    "06277", -- Wish Eater
+    "06019", -- Bloodlust
+    "06022", -- Pendant of the Queen
+    "05317", -- Blood-rite
+    "06113", -- Essence of the Dream
+    "06028", -- Stars Are Right
+    "06025", -- Guardian of the Crystallizer
+    "06283", -- Unbound Beast
+    "06032", -- Zeal
+    "06031", -- Hope
+    "06033", -- Augur
+    "06331", -- Dream Parasite
+    "06015a", -- Dream-Gate
+  },
+  globalPos = { x = -33.88, y = 1.5, z = 85.61 },
+  rotation = { x = 0, y = 270, z = 0 },
+  spread = true,
+}
+
+function onLoad(savedData)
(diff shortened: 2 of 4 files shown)


src/playercards/PlayerCardSpawner.ttslua
```diff
@@ -1,3 +1,9 @@
+
+-- Amount to shift for the next card (zShift) or next row of cards (xShift)
+-- Note that the table rotation is weird, and the X axis is vertical while the
+-- Z axis is horizontal
+local zShift = -2.29998
```
Chr1Z93, line 5: Does this need that many decimals?

src/playercards/spawnbag/BondedBag.ttslua
```diff
@@ -0,0 +1,63 @@
...
+    "06113", -- Essence of the Dream
+    "06028", -- Stars Are Right
+    "06025", -- Guardian of the Crystallizer
+    "06283", -- Unbound Beast
+    "06032", -- Zeal
+    "06031", -- Hope
+    "06033", -- Augur
+    "06331", -- Dream Parasite
+    "06015a", -- Dream-Gate
+  },
```
Chr1Z93, line 20: Nit: list could be sorted by ID

Buhallin: This maintains the ordering currently in the bag

src/playercards/spawnbag/SpawnBag.ttslua
```diff
@@ -0,0 +1,202 @@
...
+
+-- Generates a table of save state that can be included in the parent object's onSave
+SpawnBag.getStateForSave = function()
+  local saveState = {
+    placed = placedSpecs,
+    placedObjects = placedObjectGuids,
+    recall = recallZone,
+  }
+
+  return saveState
```
Chr1Z93, line 77: could get rid of line 76+77 and just replace the "local" in line 71 with "return"

src/playercards/spawnbag/BondedBag.ttslua
```diff
@@ -0,0 +1,63 @@
...
+function createActionButtons()
+    self.createButton({
+        label="Place", click_function="buttonClick_place", function_owner=self,
+        position={1,0.1,2.1}, rotation={0,0,0}, height=350, width=800,
+        font_size=250, color={0,0,0}, font_color={1,1,1}
+    })
+    self.createButton({
+        label="Recall", click_function="buttonClick_recall", function_owner=self,
+        position={-1,0.1,2.1}, rotation={0,0,0}, height=350, width=800,
+        font_size=250, color={0,0,0}, font_color={1,1,1}
```
Chr1Z93, line 52: Nit: use one line per argument

src/playercards/spawnbag/BondedBag.ttslua
```diff
@@ -0,0 +1,63 @@
...
+    end
+  end
+  createActionButtons()
+end
+
+function onSave()
+  local saveState = {
+    spawnBagState = SpawnBag.getStateForSave(),
+  }
+  return JSON.encode(saveState)
```
Chr1Z93, line 40: Is it intended that the savestate is a table with one entry instead of that entry directly?
Also I wouldn't declare a new local variable if it's only used once (and place the table directly into the "JSON.encode()")

src/playercards/spawnbag/BondedBag.ttslua
```diff
@@ -0,0 +1,63 @@
...
+        position={-1,0.1,2.1}, rotation={0,0,0}, height=350, width=800,
+        font_size=250, color={0,0,0}, font_color={1,1,1}
+    })
+end
+
+function buttonClick_place()
+  SpawnBag.spawn(SPAWN_SPEC)
+end
+
+--Recalls objects to bag from table
```
Chr1Z93, line 60: Nit: space after "--"

src/playercards/spawnbag/SpawnBag.ttslua
```diff
@@ -0,0 +1,202 @@
...
+    b = 0,
+    a = 0,
+  },
+  Locked = true,
+  Grid = true,
+  Snap = false,
+  Tooltip = false,
+}
+
+
```
Chr1Z93, line 56: Nit: delete one of the two empty lines

src/playercards/spawnbag/SpawnBag.ttslua
```diff
@@ -0,0 +1,202 @@
...
+    recallZone = { }
+    recallZone.upperLeft = { x = pos.x + RECALL_BUFFER_X, z = pos.z + RECALL_BUFFER_Z }
+    recallZone.lowerRight = { x = pos.x - RECALL_BUFFER_X, z = pos.z - RECALL_BUFFER_Z }
+    return
+  else
+    if (pos.x > recallZone.upperLeft.x) then
+      recallZone.upperLeft.x = pos.x + RECALL_BUFFER_X
+    end
+    if (pos.x < recallZone.lowerRight.x) then
+      recallZone.upperLeft.x = pos.x - RECALL_BUFFER_X
```
Chr1Z93, line 152: pretty sure this should contain ".lowerRight" instead of "upperLeft"

src/playercards/spawnbag/SpawnBag.ttslua
```diff
@@ -0,0 +1,202 @@
...
+      },
+      })
+  end
+end
+
+-- Checks to see if the given object is in the current recall zone.  If there isn't a recall zone,
+-- will return true so that everything can be easily cleaned up.
+function isInRecallZone(obj)
+  if (recallZone == nil) then
+    return true
```
Chr1Z93, line 197: Why would you want to return true if there is no zone defined?

Chr1Z93 (changes requested):
A bit of nit-picking and a bit of small stuff

src/playercards/spawnbag/BondedBag.ttslua
Buhallin: It is.  Wrapping it in a save object is redundant when there's one, but minimally so and makes it much easier to see/change if more get added.

The extra local variable is a stylistic preference - I find embedded table declarations hard to read and format well, so generally avoid them.

src/playercards/spawnbag/SpawnBag.ttslua
Buhallin: This is basically an error condition, and fails to "recall everything" rather than "recall nothing".  IMHO is a better default behavior to be overly aggressive than just do nothing.

src/playercards/PlayerCardSpawner.ttslua
Buhallin: Nope.  Was trying to match the Memory Bag layout exactly, but certainly overkill.

src/playercards/spawnbag/SpawnBag.ttslua
Buhallin: Done.

src/playercards/spawnbag/SpawnBag.ttslua
Buhallin: It should, thanks for the catch.  Hazard of test case being only one line of cards.

Chr1Z93 (approved):

PR #50, opened 2022-11-17
Refactor upgrade sheet handling in the deck loader

Buhallin:
- Split upgrade sheets to follow the slots/data pattern
- Factor out methods to set selections on upgrade sheets
- Factor out method to add Summoned Servitor minicard
- Add type 'UpgradeSheet' to appropriate cards
- Remove handleStartsInPlay() method and detect it as a default zone

### DIFF
--- a/src/arkhamdb/DeckImporterMain.ttslua
+++ b/src/arkhamdb/DeckImporterMain.ttslua
@@ -226,6 +226,9 @@ function maybeAddInvestigatorCards(deck, slots)
   end
 end
 
+-- Process the card list looking for the customizable cards, and add their upgrade sheets if needed
+---@param slots: The slot list for cards in this deck. Table key is the cardId, value is the number
+--     of those cards which will be spawned
 function maybeAddCustomizeUpgradeSheets(slots, configuration)
   local allCardsBag = getObjectFromGUID(configuration.card_bag_guid)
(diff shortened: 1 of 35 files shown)


src/arkhamdb/DeckImporterMain.ttslua
```diff
@@ -224,6 +226,27 @@ function maybeAddInvestigatorCards(deck, slots)
...
+      slots[cardId.."-c"] = 1
+    end
+  end
+end
+
+function maybeAddSummonedServitor(slots)
+  for cardId, cardCount in pairs(slots) do
+    -- spawn additional minicard for 'Summoned Servitor'
+    if cardId == "09080" then
+      slots["09080-m"] = cardCount
```
Chr1Z93, line 244: I don't think we need the cardCount because you only ever need one mini-card per player since "Summoned Servitor" is unique (and wow this makes the changing of the upkeep function to flip your own servitor way easier 🤣)

Buhallin: Huh.  I actually changed this to just be 1.  Must have missed it in an update.  And turns out all my doc changes got lost too.  Welll...  boo.

Chr1Z93 (approved):

Chr1Z93 (approved):

PR #51, opened 2022-11-17
Metadata minimization

Buhallin:
Mass change to streamline the metadata JSON.  For boolean fields which are rarely true (e.g. permanent, weakness) the fields are removed and assumed to be false if absenst.

It also removes the traits field if it was empty, which would apply to non-card cards like Minicards and Upgrade Sheets.


Chr1Z93 (approved):
not checked in detail 😅

PR #52, opened 2022-11-17
Add card type tags to playermats

Buhallin:
Tags the main playmat slots as Asset, and the investigator as Investigator.  This matches the tagging on player cards so only the appropriate type of card will snap to those points.


Chr1Z93:
Do we actually want this? Would stop all custom content from snapping (and even campaign assets like the partners from EotE)

Buhallin:
> Do we actually want this? Would stop all custom content from snapping (and even campaign assets like the partners from EotE)

I'll leave the final call up to CZ, but I think we do.  If we're going to do more with card data, at some point it's going to break the custom content in some ways.  Snap points are a convenience (a major one, to be sure) so the lack of snapping isn't the end of the world.

I actually like this as a subtle way to nudge custom content towards the expectations in our brave new world.

BootleggerFinn:
> > Do we actually want this? Would stop all custom content from snapping (and even campaign assets like the partners from EotE)
> 
> I'll leave the final call up to CZ, but I think we do. If we're going to do more with card data, at some point it's going to break the custom content in some ways. Snap points are a convenience (a major one, to be sure) so the lack of snapping isn't the end of the world.
> 
> I actually like this as a subtle way to nudge custom content towards the expectations in our brave new world.

I think we could probably do this in a way that doesn't break custom content snapping, though, and it's not realistic that existing custom content will all get updated.

Suppose the points restrict any cards _with metadata_ to snapping only if they have the right type, but any cards _without_ metadata will just snap indiscriminately.  This would reverse "the problem" so that custom content might snap when it shouldn't, which I think is a bit preferable since snapping can just be turned off for an individual card with a right-click.

Buhallin:
Unfortunately, TTS's way to target snap points is very limited.  It can ONLY be by tags, so we can't do any kind of code check on it (like whether it has metadata).

Chr1Z93:
Closed until we find a less conflicting solution

Buhallin:
I had a tool to do this so I didn't have to click 50 snap points manually.  Maybe adding it to the player contribution bag is a decent choice?

Chr1Z93:
> I had a tool to do this so I didn't have to click 50 snap points manually. Maybe adding it to the player contribution bag is a decent choice?

Yeah, that sounds like a good compromise

argonui:
Looks good

argonui (approved):

Chr1Z93 (changes requested):
There is a branch open for including this as function in the playmats (https://github.com/argonui/SCED/tree/playermat-tags) - this PR should not be merged

Buhallin:
I think it still makes sense to merge this as the base version, since it's going to be the default?

Chr1Z93:
> > There is a branch open for including this as function in the playmats (https://github.com/argonui/SCED/tree/playermat-tags) - this PR should not be merged
> 
> This doesn't make sense. We aren't long term having branches for features stay open. This pr should be merged.

I don't understand what the issue is? Why would we merge this while we are still having a discussion about this? Furthermore, we have a code freeze atm and this is far from a bug-fix.
I agree that this could be the default setting, but imo that should wait until we actually have an option panel to disable it.

What is the point in disabling the snapping for all custom cards right now?

Buhallin:
So first, everyone chill.

This is being held because of the code freeze, no other reason.  I'm planning to submit it once 2.4 releases and we're clear to submit again.

We should submit this as the default.  We have accepted that new features will make some fan content work more erratically.  If we have the options panel in the same release to minimize that then great, but it should not block this PR because it's doing what we all agreed is fine to do.

Chr1Z93 (approved):

PR #54, opened 2022-11-18
Add skill selection capability to Living Ink

Buhallin:
Skill icons can now be clicked on the sheet to circle them


Buhallin:
The UpgradeSheetBag should not be in this commit - somehow I've been getting new files drifting across branches all day (I think Atom is "helpfully" restoring them for me when it sees them disappear).  Removing it's kind of a pain, I know it's bad of me but I'm going to leave it here rather than try and fix the commit.

Chr1Z93 (approved):

PR #57, opened 2022-11-18
raising tokens, combining and updating source scripts

Chr1Z93:
unified the position of the tokens and playmats, combined the 3 required sourcescripts into 1 and updated it (now also includes a function that can be called to set to a specific value)


src/playermat/GenericCounter.ttslua
```diff
@@ -0,0 +1,39 @@
...
+function onLoad(saved_data)
+    if saved_data ~= nil then
+        val = JSON.decode(saved_data)
+    end
+
+    self.createButton({
+        label = tostring(val),
+        click_function = "addOrSubtract",
+        function_owner = self,
+        position = { 0, 0.06, 0 },
```
Buhallin, line 16: The damage trackers all had slightly different positions to keep them in the right place given the different token shapes.  Making this common loses that.  Extract to a parent file for each tracker type?
image

objects/Resources.816d84.json
```diff
@@ -43,7 +43,7 @@
   "Tooltip": false,
   "Transform": {
     "posX": -59.793,
-    "posY": 1.61,
+    "posY": 1.55,
```
Buhallin, line 46: I think these are a little higher than they used to be.  Bump them down a bit, or was the previous height causing issues?
image

Buhallin (changes requested):

src/playermat/GenericCounter.ttslua
Buhallin: Previous centering, for reference
image

objects/Resources.816d84.json
Buhallin: Previous height
image

Chr1Z93: previous height was causing flickering when viewed from directly above, but there is probably a middle ground

src/playermat/GenericCounter.ttslua
```diff
@@ -0,0 +1,39 @@
+MIN_VALUE = -99
```
Buhallin, line 1: I think we can set this to zero.  We don't have anything that tracks negatives, right?

Chr1Z93: probably not needed, but wouldn't hurt to let this in there just in case someone is using that for something?

src/playermat/GenericCounter.ttslua
Chr1Z93: using name to distinguish between positions now

src/playermat/GenericCounter.ttslua
Buhallin: I can only think of one use case even at the fringe (someone who counts health/sanity down rather than up, in Dim Carcosa) but I know I see players in my group often overclicking and going negative when healing or paying for things.  I think we should floor it at zero, but it's a minor annoyance, can leave it for later.

Buhallin (approved):

PR #65, opened 2022-11-24
Fix multiple bugs with uses

Buhallin:
- Saved old cards have a different structure for uses (single vs. array) making the change breaking.  Code was assuming an array, breaking any loaded objects with the old structure
- Uses spawn loop was setting the values only, resulting in only the last item being spawned
- Upkeep checks were embedded incorrectly, leading to logic errors in certain circumstances
- If uses had multiple entries, tokens were spawning on top of each other and exploding, probably due to the yPos shift in playmats.  Added an offset to spawn the groups in different places to avoid this, which is prettier anyway

### DIFF
--- a/src/playermat/Playmat.ttslua
+++ b/src/playermat/Playmat.ttslua
@@ -360,8 +360,8 @@ function spawnTokenOn(object, offsets, tokenType)
   spawnToken(tokenPosition, tokenType)
 end
 
--- spawn a group of tokens of the given type on the object
--- @parm object Object to spawn the tokens on
+-- Spawn a group of tokens of the given type on the object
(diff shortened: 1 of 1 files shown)


src/playermat/Playmat.ttslua
```diff
@@ -351,9 +361,25 @@ function spawnTokenOn(object, offsets, tokenType)
 end
 
 -- spawn a group of tokens of the given type on the object
-function spawnTokenGroup(object, tokenType, tokenCount)
+-- @parm object Object to spawn the tokens on
```
Chr1Z93, line 364: typo (par"a"m)

Chr1Z93 (approved):

PR #68, opened 2022-11-24
Optional clue counter: fixing position, using require

Chr1Z93:
- fixed the position
- fixed the name ("Resources" --> "Clues")
- now uses "require" instead of duplicate code


objects/ClueCounterSwapper.d919d6/Clues.3f22e5.json
```diff
@@ -33,11 +33,11 @@
   "IgnoreFoW": false,
   "LayoutGroupSortIndex": 0,
   "Locked": true,
-  "LuaScriptState": "[true,0]",
-  "LuaScript_path": "ClueCounterSwapper.d919d6/Resources.3f22e5.ttslua",
+  "LuaScript": "require(\"playermat/GenericCounter\")",
+  "LuaScriptState": "0",
```
Buhallin, line 37: Is this right?  The LuaScriptState is typically JSON, will this parse correctly?

Chr1Z93: Yeah, this is correct (it's just a single value that is saved)

Buhallin (approved):

PR #75, opened 2022-11-24
Handle On the Mend when loading from ArkhamDB

Buhallin:
Vincent Lee gets one copy of On the Mend per investigator, and they should start set aside.

### DIFF
--- a/src/arkhamdb/DeckImporterMain.ttslua
+++ b/src/arkhamdb/DeckImporterMain.ttslua
@@ -321,7 +321,7 @@ function getDefaultCardZone(cardMetadata)
   if (cardMetadata.id == "09080-m") then -- Have to check the Servitor before other minicards
     return "SetAside6"
   elseif (cardMetadata.id == "09006") then -- On The Mend is set aside
-      return "SetAside2"
+    return "SetAside2"
   elseif cardMetadata.type == "Investigator" then
     return "Investigator"
   elseif cardMetadata.type == "Minicard" then
--- a/src/arkhamdb/DeckImporterMain.ttslua
+++ b/src/arkhamdb/DeckImporterMain.ttslua
@@ -32,6 +32,8 @@ customizationRowsWithFields["09101"].inputMap[1] = 1
 customizationRowsWithFields["09101"].inputMap[2] = 2
 customizationRowsWithFields["09101"].inputMap[3] = 3
 
+local PLAY_AREA_GUID = "721ba2"
+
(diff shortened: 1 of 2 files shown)


src/core/PlayArea.ttslua
```diff
@@ -8,7 +8,7 @@ DEBUG = false
 -- we use this to turn off collision handling until onLoad() is complete
 COLLISION_ENABLED = false
 
-local COUNTER = getObjectFromGUID('f182ee')
+local investigatorCounter = getObjectFromGUID('f182ee')
```
Chr1Z93, line 11: should move this to "onLoad" to avoid issues with different load orders

src/arkhamdb/DeckImporterMain.ttslua
```diff
@@ -245,11 +248,20 @@ end
...
+
+-- On the Mend should have 1-per-investigator copies set aside, but ArkhamDB always sends 1.  Update
+-- the count based on the investigator count
+---@param slots: The slot list for cards in this deck. Table key is the cardId, value is the number
+--     of those cards which will be spawned
+function maybeAddOnTheMend(slots)
+  if slots["09006"] ~= nil then
+    local playArea = getObjectFromGUID(PLAY_AREA_GUID)
+    if playArea ~= nil then
+      slots["09006"] = playArea.call("getInvestigatorCount")
```
Chr1Z93, line 264: Wouldn't it be better to directly access the playercounter instead of doing the additional step via the playarea?

src/core/PlayArea.ttslua
Buhallin: Bigger refactoring of this coming later, just trying to do a minimal disruption change (apart from the name fix, which I couldn't help myself on).

src/arkhamdb/DeckImporterMain.ttslua
Buhallin: This is a design choice - depends on whether we consider the investigator count as a separate element, or part of the play area.  For now it's part of the area so I'm leaving it there and using the play area as a parent object/interface.  Possibly more changes coming on this, but again trying to be minimally disruptive to make the fix in the freeze window.

src/core/PlayArea.ttslua
Buhallin: Eh, take that back, I'll fix it.  Change incoming.

Buhallin: Done.

Chr1Z93 (approved):

PR #76, opened 2022-11-24
Add startsInHand handling for the deck loader

Buhallin:
Currently this is only used for Diana Stanley's Dark Insight, but it adds generalized handling for drawing cards from the deck on load based on a metadata flag.

### DIFF
--- a/src/arkhamdb/DeckImporterMain.ttslua
+++ b/src/arkhamdb/DeckImporterMain.ttslua
@@ -373,7 +373,7 @@ function loadCards(slots, investigatorId, customizations, playerColor, commandMa
       local deckPos = Zones.getZonePosition(playerColor, zone)
       deckPos.y = 3
 
-      local spreadCallback = nil
+      local callback = nil
       -- If cards are spread too close together TTS groups them weirdly, selecting multiples
       -- when hovering over a single card.  This distance is the minimum to avoid that
       local spreadDistance = 1.15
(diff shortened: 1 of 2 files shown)


src/arkhamdb/DeckImporterMain.ttslua
```diff
@@ -430,6 +432,22 @@ function loadCards(slots, investigatorId, customizations, playerColor, commandMa
...
+-- @param deck Callback-provided spawned deck object
+-- @param playerColor Color of the player to draw the cards to
+function deckSpawned(deck, playerColor)
+  local handPos = Player[playerColor].getHandTransform(1).position -- Only one hand zone per player
+  local deckCards = deck.getData().ContainedObjects
+  -- Process in reverse order so taking cards out doesn't upset the indexing
+  for i = #deckCards, 1, -1 do
+    local cardMetadata = JSON.decode(deckCards[i].GMNotes) or { }
+    if cardMetadata.startsInHand then
+      deck.takeObject({ index = i - 1, position = handPos, flip = true, smooth = true})
```
Chr1Z93, line 446: I assume cards with "startsInHand" are always at the top of the deck?

Buhallin: No, they'll be in the usual order.  We'd have to parse the deck anyway, simpler to just loop the whole thing and not try to add special casing or be dependent on the ordering happening elsewhere.

Chr1Z93 (approved):

BootleggerFinn (approved):

PR #78, opened 2022-11-30
Clean Up Helper: Remove vector lines

Chr1Z93:


objects/Fan-MadeAccessories.aa8b38/CleanUpHelper.26cf4b.ttslua
```diff
@@ -179,8 +179,16 @@ function cleanUp()
 
   getTrauma()
   resetCounters()
+
+  -- bless / curse manager prints removed amounts
   removeBlessCurse()
 
+  -- image swapper prints reset message
+  resetPlayareaImage()
```
Buhallin, line 187: This seems very specific to the "different image for every scenario" play pattern.  If the user just wants to set a different image that they keep, you're wiping that out.  Even for the per-scenario user I don't think this adds much, as there's no reduction in the user flow (whatever the current is, they'll do exactly the same thing to pick the image for the next scenario).

What's the use case you see for this?

Chr1Z93: The clean up helper is meant to reset the table between scenarios during campaign play. imo this should include the image reset but I see that this could sometimes not be wanted. How about having that as toggleable option?

Buhallin: What I'm getting at is that while it might fall into a broad "reset the table" concept, I don't think it provides any user benefit by doing so.  The ideal case doesn't actually help the user, and the others it's actively harmful.

Should be an option at a minimum, but I honestly think it just doesn't need to be done.

Chr1Z93: I am confused why you think it doesn't provide any benefit? The custom playmat images are primarily used with different images per scenario.

Buhallin (approved):

PR #79, opened 2022-11-30
Displacement tool: visual update

Chr1Z93:


Buhallin (commented):
Looks like this PR got munched with the cleanup helper?

Chr1Z93:
> Looks like this PR got munched with the cleanup helper?

Uff yeah, I'll remove that part

objects/Fan-MadeAccessories.aa8b38/DisplacementTool.0f1374.ttslua
```diff
@@ -1,62 +1,70 @@
----
---- Generated by Luanalysis
---- Created by Whimsical.
---- DateTime: 2022-01-21 1:39 p.m.
----
-
-function onload()
-  self:getRotation(Vector(0, 270, 0))
-end
+local zone = getObjectFromGUID("a2f932")
```
Buhallin, line 1: Change to grab zone on demand

objects/Fan-MadeAccessories.aa8b38/DisplacementTool.0f1374.ttslua
```diff
@@ -1,62 +1,70 @@
...
+function shift_left(color) shift(color, "left") end
+
+function shift_right(color) shift(color, "right") end
+
+function shift_up(color) shift(color, "up") end
+
+function shift_down(color) shift(color, "down") end
+
+function shift(color, direction)
   if not zone then
```
Buhallin, line 60: It might be beyond the scope of the effort here, but I don't think we need to fail this.  The scripting zone is a convenience, but not necessary for the operation to succeed.  If the zone isn't found for some reason (which admittedly shouldn't happen) can fall back to a position check to move everything.

objects/Fan-MadeAccessories.aa8b38/DisplacementTool.0f1374.ttslua
```diff
@@ -1,62 +1,70 @@
...
   end
 
-  local adjustment = Vector(0, 0, 0)
-  adjustment[offset.axis] = offset.offset
-
-  ---@param object TTSObject
-  for _, object in ipairs(zone:getObjects()) do
-    if not (excluded[object:getGUID():lower()] or object:hasTag("displacement_excluded")) then
-      object:translate(adjustment)
+  for _, object in ipairs(zone.getObjects()) do
```
Buhallin, line 65: From an allocation of responsibility perspective, I think it would make sense for the move code to be owned by Playarea and accessed via API.

objects/Fan-MadeAccessories.aa8b38/DisplacementTool.0f1374.ttslua
Chr1Z93: That can be added if needed, but I don't think we should account for a core zone missing. There is nothing that should cause a removal of that zone.

objects/Fan-MadeAccessories.aa8b38/DisplacementTool.0f1374.ttslua
Chr1Z93: As talked about, that will be done when you add the PlayareaAPI during the locations connection script.

Buhallin (approved):

PR #81, opened 2022-12-01
Doom Counter Option UI // Clue Counter clean up

Chr1Z93:
Preface: I know that this should have been split into two PRs - please don't hate me for it 😅

1) Doom Counter now has a UI to select options:
image
image
The "Reset" button will take the options into account and call the respective function from the "Doom in Play Counter"

2) Because I saw some flickering for the label of the Master Clue Counter I also took a look at it's script and couldn't stop myself from updating that one. Removed Clues will now be moved into the global trashcan instead of the white trashcan.


src/core/DoomCounter.ttslua
```diff
@@ -0,0 +1,75 @@
...
+function onLoad(savedData)
+  if savedData ~= "" then
+    local loadedData = JSON.decode(savedData)
+    val              = loadedData[1]
+    options          = loadedData[2]
+
+    -- restore state for option panel
+    for key, bool in pairs(options) do
+      if bool then
+        self.UI.setAttribute("option" .. key, "isOn", 0)
```
Buhallin, line 22: Save/load doesn't seem to be working for me.  Haven't tested exhaustively, but after changing a setting, overwriting, and then reloading it resets to all options being chosen.

I think isOn may need a direct true/false here, rather than the converted 0-is-false?

I think you can just set the value of bool to each for all the items as well to simplify a bit

src/core/DoomCounter.ttslua
```diff
@@ -0,0 +1,75 @@
...
+  self.editButton({ index = 0, label = tostring(val) })
+  printToAll("Doom on agenda set to: " .. val)
+end
+
+-- called by "Reset" button to remove doom
+function startReset()
+  if options.Agenda then
+    updateVal(0)
+  end
+  getObjectFromGUID("652ff3").call("removeDoom", options)
```
Buhallin, line 58: Move GUID into a named constant

objects/DoomCounter.85c4c6.xml
```diff
@@ -0,0 +1,37 @@
...
+      </Cell>
+      <Cell>
+        <ToggleButton onClick="toggleOptions" fontSize="55">☰</ToggleButton>
+      </Cell>
+    </Row>
+  </TableLayout>
+</Panel>
+
+<Panel id="Options" offsetXY="0 535" active="false">
+  <VerticalLayout height="300" width="500" spacing="10" childAlignment="MiddleCenter">
```
Buhallin, line 20: Might experiment with showAnimation and hideAnimation on this

Buhallin (approved):

PR #89, opened 2022-12-11
SCED Intro Tour, Part 1

Buhallin:
Creates the basic framework for the SCED introduction/training tour.  As this will be a large feature, this is intended to checkpoint progress and work.  This a pure code submission, with no hooks to the actual mod yet.

Depending on the severity of review comments, issues may be fixed in one of the (many) following PRs rather than directly in this one.

What's here:
- Basic tour script definition
- Script UI creation, camera handling, and flow

What's not here:
- Multiplayer support.  Everything is currently locked to white.
- Final UI.  Most UI is placeholder and will be updated later.
- Extra script capability such as higlighting objects, explicit positioning, etc.

### DIFF
--- a/src/core/tour/TourManager.ttslua
+++ b/src/core/tour/TourManager.ttslua
@@ -1,0 +1,188 @@
+do
+  require("core/tour/TourScript")
+  require("core/tour/TourCard")
+  local TourManager = { }
+  local internal = { }
+
+  -- Base IDs for various tour card UI elements.  Actual IDs will have _[playerColor] appended
+  local cardId = "tourCard"
+  local narratorId = "tourNarratorImage"
+  local textId = "tourText"
+  local nextButtonId = "tourNext"
+  local stopButtonId = "tourStop"
+
+  -- Table centerpoint for the camera hook object.  Camera handling is a bit erratic so it doesn't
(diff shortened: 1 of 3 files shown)


src/core/tour/TourManager.ttslua
```diff
@@ -0,0 +1,188 @@
...
+      object = hook,
+      offset = { x = -20, y = 30, z = 0 }
+    })
+    internal.showCurrentCard()
+  end
+
+  -- Creates an XMLUI entry in Global for a player-specific tour card.  Dynamically creating this
+  -- is somewhat complex, but ensures we can properly handle any player color.
+  -- @param playerColor Player color to create the card for
+  internal.createTourCard = function(playerColor)
```
Chr1Z93, line 164: How about  a handling for playerColor = "All"?

Chr1Z93 (approved):

src/core/tour/TourManager.ttslua
Buhallin: With the camera controls being distinct for each player an All doesn't really work here.  It might theoretically be possible to hook all the active players, but the camera controls are so touchy I'm afraid to try.  Would also be some complicated questions like what happens if a new player joined mid-tour, that just aren't worth dealing with.

PR #93, opened 2022-12-12
Optionpanel: Part 1

Chr1Z93:
This contains the first part of the code for the option panel.


src/core/Global.ttslua
```diff
@@ -127,10 +130,25 @@ local overallStats = {
 -- general code
 ---------------------------------------------------------
 
-function onLoad()
+-- saving state of optionPanel to restore later
+function onSave() return JSON.encode(optionPanel) end
```
Buhallin, line 134: Please put this in a wrapper object with a named key.

src/core/Global.ttslua
```diff
@@ -127,10 +130,25 @@ local overallStats = {
...
+
+function onLoad(savedData)
+  if savedData then
+    optionPanel = JSON.decode(savedData)
+    for id, enabled in pairs(optionPanel) do
```
Buhallin, line 139: Move this into an update method

src/core/Global.ttslua
```diff
@@ -127,10 +130,25 @@ local overallStats = {
...
+      if enabled then self.UI.setAttribute("toggle" .. id, "isOn", true) end
+    end
+  else
+    for i = 1, 8 do
```
Buhallin, line 143: Hardcoding like this will mean it has to change every time an option gets added.  It's also going to get messy if/when we get an option that isn't just a boolean.  Better to have the options handle nil values gracefully.

src/core/Global.ttslua
```diff
@@ -600,41 +618,85 @@ local source_repo = 'https://raw.githubusercontent.com/seth-sced/loadable-object
...
+    state = false
+  end
+
+  self.UI.setAttribute("toggle" .. id, "isOn", state)
+  optionPanel[id] = state
+
+  applyChange(id, state)
+end
+
+local PlayerMatAPI = require("playermat/PlaymatApi")
```
Buhallin, line 660: Better to declare these at the top so it's easy to see the included modules.  Also, lowerCamelCase as it's a variable

src/core/Global.ttslua
```diff
@@ -600,41 +618,85 @@ local source_repo = 'https://raw.githubusercontent.com/seth-sced/loadable-object
...
+function applyChange(id, state)
```
Buhallin, line 661: More specific name since this is in Global

src/core/Global.ttslua
```diff
@@ -600,41 +618,85 @@ local source_repo = 'https://raw.githubusercontent.com/seth-sced/loadable-object
...
+  -- option 1: Snap tags
+  if id == "1" then
+    printToAll("Playermat snap tags " .. (state and "en" or "dis") .."abled.", "White")
```
Buhallin, line 664: I don't think this is necessary.  The user clicked the value.  The value updated.  They can see it :)

src/playermat/Playmat.ttslua
```diff
@@ -45,8 +45,9 @@ local RESOURCE_COUNTER
 
 -- global variable so it can be reset by the Clean Up Helper
 activeInvestigatorId = "00000"
+local drawButton = false
 
-function onSave() return JSON.encode({zoneID = zoneID, playerColor = PLAYER_COLOR, activeInvestigatorId = activeInvestigatorId}) end
+function onSave() return JSON.encode({zoneID = zoneID, playerColor = PLAYER_COLOR, activeInvestigatorId = activeInvestigatorId, drawButton = drawButton}) end
```
Buhallin, line 50: This is way too long to be a single line.  I know we're not strictly enforcing it, but 100 characters is generally a good guideline.

src/playermat/Playmat.ttslua
```diff
@@ -648,11 +652,40 @@ function spawnToken(position, tokenType)
   Global.call('spawnToken', {position, tokenType, PLAY_ZONE_ROTATION})
 end
 
+-- Sets this playermat's draw 1 button to visible
+---@param visibleButton Boolean.  Whether the draw 1 button should be visible
+function showDrawButton(visibleButton)
```
Buhallin, line 657: Nit: visibleButton should probably just be visible.  Plenty of context from the function name

src/playermat/Playmat.ttslua
```diff
@@ -45,8 +45,9 @@ local RESOURCE_COUNTER
 
 -- global variable so it can be reset by the Clean Up Helper
 activeInvestigatorId = "00000"
+local drawButton = false
```
Buhallin, line 48: Nit: showDrawButton

src/playermat/Playmat.ttslua
```diff
@@ -648,11 +652,40 @@ function spawnToken(position, tokenType)
...
+        click_function = "doDrawOne",
+        function_owner = mat,
+        position = { 1.84, 0.1, -0.36 },
+        scale = { 0.12, 0.12, 0.12 },
+        width = 800,
+        height = 280,
+        font_size = 180
+    })
+  else
+    -- remove button with index 9 if 10 buttons are present (because index starts at 0)
```
Buhallin, line 671: This is going to be very fragile - e.g. the change to the larger playmats added a new discard button, which would break this.  use getButtons() and iterate it looking for the matching [something, probably label] to find the index

src/playermat/Playmat.ttslua
```diff
@@ -648,11 +652,40 @@ function spawnToken(position, tokenType)
...
+    if #self.getButtons() == 10 then
+      self.removeButton(9)
+    end
+  end
+end
+
+-- Spawns / destroys a clickable clue counter for this playmat
+---@param clickableCounter Boolean.  Whether the clickable clue counter should be present
+function clickableClues(clickableCounter)
```
Buhallin, line 680: As above - showCounters or something similar so it doesn't sound like you're passing the actual Counter

src/playermat/Playmat.ttslua
```diff
@@ -648,11 +652,40 @@ function spawnToken(position, tokenType)
...
+  print("dummy function for clue counters")
+end
+
 -- Sets this playermat's snap points to limit snapping to matching card types or not.  If matchTypes
 -- is true, the main card slot snap points will only snap assets, while the investigator area point
 -- will only snap Investigators.  If matchTypes is false, snap points will be reset to snap all
 -- cards.
--- @param matchTypes Boolean.  Whether snap points should only snap for the matching card types.
+---@param matchTypes Boolean.  Whether snap points should only snap for the matching card types.
```
Buhallin, line 688: Is this style expected by something in Lua?  Or just added by VSCode?

src/playermat/PlaymatApi.ttslua
```diff
@@ -13,20 +13,40 @@ do
...
+  ---@param visibleButton Boolean.  Whether the draw 1 button should be visible or not
+  ---@param matColor String for one of the active player colors - White, Orange, Green, Red.  Also
+  --     accepts "All" as a special value which will apply the setting to all four mats.
+  PlaymatApi.showDrawButton = function(visibleButton, matColor)
+    for _, mat in ipairs(internal.getMatForColor(matColor)) do
+      mat.call("showDrawButton", visibleButton)
+    end
+  end
+
+  -- Spawns a clickable clue counter for the requested playermat
```
Buhallin, line 36: Nit: Spawns or hides the clickable...

xml/OptionPanel.xml
```diff
@@ -0,0 +1,123 @@
...
+        <Text class="group-header">PLAYERMAT SETTINGS</Text>
+      </Panel>
+
+      <!-- Option 1 -->
+      <HorizontalLayout class="group-content">
+        <VerticalLayout class="text-column">
+          <Text class="option-header">Enable snap tags</Text>
+          <Text class="description">Only cards with the tag "Asset" will snap (official cards are supported by default).&#xA;Disable this if you are having issues with custom content.</Text>
+        </VerticalLayout>
+        <Toggle id="toggle1" onValueChanged="onClick_toggleOption(1)" />
```
Buhallin, line 40: These really should use meaningful IDs rather than generic numbering.  We're not going to be able to swap them out at will as each button is hard-linked to a given option, so the generic IDs just obfuscate what's what.

Buhallin (changes requested):

src/playermat/Playmat.ttslua
Chr1Z93: That style supports hightling in VS Code:
image

xml/OptionPanel.xml
Chr1Z93: Assigning them numbered indexes makes saving/loading options way easier

src/playermat/Playmat.ttslua
Chr1Z93: That's already the name of the function, so not changing that

Buhallin: Can always pick something else :)  drawButtonVisible, isDrawButtonVisible, etc.  The point about what the name implies this is holding still applies.

xml/OptionPanel.xml
Chr1Z93: A lookup map sounds like a good solution!
Would like to defer that until part 2 has made it's way past the merge because this would require changes at a lot of different positions in the code.

Buhallin: Fair enough.

Buhallin (approved):

PR #94, opened 2022-12-12
Optionpanel: Part 2

Chr1Z93:
This adds additional options (mainly the option to swap the clue counting to clickable tokens) and contains some cleanup (still WIP).


objects/Fan-MadeAccessories.aa8b38/CleanUpHelper.26cf4b.ttslua
```diff
@@ -209,10 +208,10 @@ function resetCounters()
   end
 
   -- reset doom on agenda
-  -- local doomcounter = getObjectFromGUID("85c4c6")
-  -- if doomcounter ~= nil then
-  --   doomcounter.call("setToZero")
-  -- end
+  local doomcounter = getObjectFromGUID("85c4c6")
+  if doomcounter ~= nil then
+    doomcounter.call("updateVal")
```
Buhallin, line 213: Nit: Even if 0 ends up as the default for a nil, should pass it here for clarity

src/core/MasterClueCounter.ttslua
```diff
@@ -6,11 +6,27 @@ local clueCounterGUIDS = {
...
+local clueClickers = {}
+local clueClickerGUIDS = {
+  "db85d6",
+  "3f22e5",
+  "891403",
+  "4111de"
+}
+
+count = 0
+clickableCounters = false
```
Buhallin, line 18: Naming: useClickableCounters.  Also, any reason not to make these local?

src/core/MasterClueCounter.ttslua
```diff
@@ -21,26 +37,41 @@ function onLoad()
...
   loopID = Wait.time(sumClues, 2, -1)
 end
 
--- removes all player clues by calling the respective function from the counting bowls
+-- removes all player clues by calling the respective function from the counting bowls / clickers
 function removeAllPlayerClues()
-  for i = 1, 4 do
-    clueCounters[i].call("removeAllClues")
+  printToAll(count .. " clue(s) from playermats removed.", "White")
+  if not clickableCounters then
```
Buhallin, line 52: Reverse this to remove the negative logic.  Same below.

src/playermat/ClueCounter.ttslua
```diff
@@ -9,12 +9,15 @@ local validCountItemList = {
 }
 local trashGUID = "70b9f6"
 exposedValue = 0
+local trashGUID = "70b9f6"
```
Buhallin, line 12: Looks like these lines got duplicated?

Buhallin (commented):
Broad design comment:

I think pretty much all this could be added inside the Playermat, with methods to change the approach, get the clue counts, and reset them exposed via the API object.  That would give ownership of the clues to the object, and make the callers (like the master clue counter) ignorant of the actual current option.

The design would be generally cleaner, but also enable some other possible features like (and I'm not suggesting we do this) allowing different playermats to have different clue options selected.

I'm agnostic on whether this is done - I do think it's better, but also a big redesign from the current.

src/core/MasterClueCounter.ttslua
Chr1Z93: Renaming it with next commit. It's not local because this will be changed when the option is toggled.

src/playermat/ClueCounter.ttslua
Chr1Z93: Good catch!

Chr1Z93:
Comments are resolved, but I found a minor issue that will be solved after this PR:
The clickable clue counters (that are now hidden below the playmat) can be clicked from certain camera angles (this is no issue in itself, but a bit counterintuitive that there is something clickable that's not doing anything).

Chr1Z93:
> Broad design comment:
> 
> I think pretty much all this could be added inside the Playermat, with methods to change the approach, get the clue counts, and reset them exposed via the API object. That would give ownership of the clues to the object, and make the callers (like the master clue counter) ignorant of the actual current option.
> 
> The design would be generally cleaner, but also enable some other possible features like (and I'm not suggesting we do this) allowing different playermats to have different clue options selected.
> 
> I'm agnostic on whether this is done - I do think it's better, but also a big redesign from the current.

I am a bit confused because the clue option is pretty much handled by the playermats (and the API). Do you mean the masterclue counter should talk to the playermats when looking for the clue count?

Buhallin:
> > Broad design comment:
> > I think pretty much all this could be added inside the Playermat, with methods to change the approach, get the clue counts, and reset them exposed via the API object. That would give ownership of the clues to the object, and make the callers (like the master clue counter) ignorant of the actual current option.
> > The design would be generally cleaner, but also enable some other possible features like (and I'm not suggesting we do this) allowing different playermats to have different clue options selected.
> > I'm agnostic on whether this is done - I do think it's better, but also a big redesign from the current.
> 
> I am a bit confused because the clue option is pretty much handled by the playermats (and the API). Do you mean the masterclue counter should talk to the playermats when looking for the clue count?

Sorry, got confused by the MasterClueCounter handling the reset and thought it was the type flip.  You're right that the change is already in the playmat.

But yes, getting and resetting clues should go through the API and the code should live in the player mat.

On the good side that's a far less impactful change, so let's do that :)

Chr1Z93:
Clue handling is moved to playermatAPI

Buhallin (approved):

PR #95, opened 2022-12-12
Datahelper: data for 10 tokens (akachi + brand lvl 4)

Chr1Z93:


src/core/DataHelper.ttslua
```diff
@@ -923,6 +923,18 @@ PLAYER_CARD_TOKEN_OFFSETS = {
...
+    { -0.7, 3, -0.9 },
+    { 0.7, 3, -0.2 },
+    { 0, 3, -0.2 },
+    { -0.7, 3, -0.2 },
+    { 0.7, 3, 0.5 },
+    { 0, 3, 0.5 },
+    { -0.7, 3, 0.5 },
+    { 0, 3, 1.2 }
+  },
   [12] = {
```
argonui, line 938: Seems like we need 11 also

argonui (changes requested):

src/core/DataHelper.ttslua
Chr1Z93: There is no possible way to get 11 at the moment

Buhallin: Worth adding though so it's covered and doesn't get missed if something else happens to hit that mark.

Buhallin (approved):

argonui (approved):

PR #96, opened 2022-12-12
SCED Intro Tour, Part 2

Buhallin:
This updates the Tour to work for each player independently.  There is still some refinement to be done, but the basic implementation is there.

I'm chaining work so this will include some of my merges and the like, but actual changes should be contained.


Chr1Z93 (approved):
LGTM

PR #97, opened 2022-12-13
Add PlayArea API

Buhallin:
Creates an API object for the PlayArea, and moves most references to the PlayArea to use the API instead.

Image swapper is excluded on this, as I'm not completely sure how TTS will handle having an object rebuild itself.

### DIFF
--- a/objects/Fan-MadeAccessories.aa8b38/CleanUpHelper.26cf4b.ttslua
+++ b/objects/Fan-MadeAccessories.aa8b38/CleanUpHelper.26cf4b.ttslua
@@ -5,6 +5,8 @@
 --                  - puts everything on playmats and hands into respective trashcans
 --                  - use the IGNORE_TAG to exclude objects from tidying (default: "CleanUpHelper_Ignore")
 
+local playAreaApi = require("core/PlayAreaApi")
+
 -- enable this for debugging
 local SHOW_RAYS = false
 
@@ -207,10 +209,10 @@ function resetCounters()
   end
 
   -- reset doom on agenda
-  local doomcounter = getObjectFromGUID("85c4c6")
-  if doomcounter ~= nil then
-    doomcounter.call("setToZero")
-  end
+  -- local doomcounter = getObjectFromGUID("85c4c6")
+  -- if doomcounter ~= nil then
+  --   doomcounter.call("setToZero")
+  -- end
 
   for i, guid in ipairs(TRACKER_GUIDS) do
     local obj = getObjectFromGUID(guid)
@@ -300,10 +302,7 @@ function tidyPlaymatCoroutine()
     end
   end
 
-  local PLAYMAT = getObjectFromGUID('721ba2')
-  if PLAYMAT then
-    PLAYMAT.setTable("SPAWNED_LOCATION_GUIDS", {})
-  end
+  playAreaApi.resetSpawnedCards()
 
   printToAll("Tidying playermats and agenda mat...", "White")
   startLuaCoroutine(self, "tidyPlayerMatCoroutine")
--- a/src/arkhamdb/DeckImporterMain.ttslua
+++ b/src/arkhamdb/DeckImporterMain.ttslua
@@ -1,5 +1,6 @@
 require("arkhamdb/LoaderUi")
 require("playercards/PlayerCardSpawner")
+local playAreaApi = require("core/PlayAreaApi")
 
 local zones = require("playermat/Zones")
 
@@ -32,8 +33,6 @@ customizationRowsWithFields["09101"].inputMap = {}
 customizationRowsWithFields["09101"].inputMap[1] = 1
 customizationRowsWithFields["09101"].inputMap[2] = 2
 customizationRowsWithFields["09101"].inputMap[3] = 3
-
-local PLAY_AREA_GUID = "721ba2"
 
 local RANDOM_WEAKNESS_ID = "01000"
 local tags = { configuration = "import_configuration_provider" }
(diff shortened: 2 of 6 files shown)


src/core/PlayAreaApi.ttslua
```diff
@@ -0,0 +1,50 @@
...
+  ---@param playerColor Color of the player requesting the shift.  Used solely to send an error
+  ---    message in the unlikely case that the scripting zone has been deleted
+  PlayAreaApi.shiftContentsLeft = function(playerColor)
+    return getObjectFromGUID(PLAY_AREA_GUID).call("shiftContentsLeft", playerColor)
+  end
+
+  -- Move all contents on the play area (cards, tokens, etc) one column right.  Certain fixed
+  -- objects will be ignored, as will anything the player has tagged with 'displacement_excluded'
+  ---@param playerColor Color of the player requesting the shift.  Used solely to send an error
+  ---    message in the unlikely case that the scripting zone has been deleted
```
Chr1Z93, line 39: I'd combine the documentation for the four shift functions (because they are the same except for the direction).

src/core/PlayAreaApi.ttslua
```diff
@@ -0,0 +1,50 @@
...
+
+  -- Returns the current value of the investigator counter from the playmat
+  ---@return Integer.  Number of investigators currently set on the counter
+  PlayAreaApi.getInvestigatorCount = function()
+    return getObjectFromGUID(PLAY_AREA_GUID).call("getInvestigatorCount")
+  end
+
+  -- Move all contents on the play area (cards, tokens, etc) one row up.  Certain fixed objects will
+  -- be ignored, as will anything the player has tagged with 'displacement_excluded'
+  ---@param playerColor Color of the player requesting the shift.  Used solely to send an error
```
Chr1Z93, line 14: What do you think about removing the playercolor as argument and instead using "broadcastToAll" for the error message?

src/core/PlayAreaApi.ttslua
```diff
@@ -0,0 +1,50 @@
+do
+  local PlayAreaApi = { }
```
Chr1Z93, line 2: nit: remove space between brackets

src/core/PlayArea.ttslua
```diff
@@ -122,7 +136,60 @@ function onCollisionEnter(collision_info)
...
+---@param playerColor Color of the player requesting the shift.  Used solely to send an error
+---    message in the unlikely case that the scripting zone has been deleted
+function shiftContentsLeft(playerColor)
+  shiftContents(playerColor, "left")
+end
+
+-- Move all contents on the play area (cards, tokens, etc) one column right.  Certain fixed
+-- objects will be ignored, as will anything the player has tagged with 'displacement_excluded'
+---@param playerColor Color of the player requesting the shift.  Used solely to send an error
+---    message in the unlikely case that the scripting zone has been deleted
```
Chr1Z93, line 166: Would also combine the documentation for the four shift functions here

src/arkhamdb/DeckImporterMain.ttslua
```diff
@@ -260,9 +259,9 @@ end
 --     of those cards which will be spawned
 function maybeAddOnTheMend(slots)
   if slots["09006"] ~= nil then
-    local playArea = getObjectFromGUID(PLAY_AREA_GUID)
-    if playArea ~= nil then
-      slots["09006"] = playArea.call("getInvestigatorCount")
+    local investigatorCount = playAreaApi.getInvestigatorCount()
+    if investigatorCount ~= nil then
```
Chr1Z93, line 263: If we are checking this anyway, there should probably be a message if it's not found.

objects/Fan-MadeAccessories.aa8b38/CleanUpHelper.26cf4b.ttslua
```diff
@@ -207,10 +209,10 @@ function resetCounters()
   end
 
   -- reset doom on agenda
-  local doomcounter = getObjectFromGUID("85c4c6")
-  if doomcounter ~= nil then
-    doomcounter.call("setToZero")
-  end
+  -- local doomcounter = getObjectFromGUID("85c4c6")
+  -- if doomcounter ~= nil then
+  --   doomcounter.call("setToZero")
```
Chr1Z93, line 214: If this gets changed to `doomcounter.call("updateVal")`  it would work again

Buhallin: Unrelated to this PR, can fix it later.  I probably shouldn't have even commented it out, but it was breaking the testing

src/core/PlayAreaApi.ttslua
Buhallin: This makes me cringe, but in the absence of anything to generate automatic documentation it's better.  Done.

src/core/PlayAreaApi.ttslua
Chr1Z93: haha, that's fair

src/core/PlayAreaApi.ttslua
Buhallin: I can see arguments for either, but I'd opt for less spammy.  In this case, if the script zone isn't found it's basically unrecoverable.  Broadcasting the message to more people who can't do anything about it doesn't seem beneficial.

src/arkhamdb/DeckImporterMain.ttslua
Buhallin: Done, and made it spawn 4 as a backup.  Though if this fails the game's pretty much irredeemably broken so On the Mend may be the only thing they get out of it...

Chr1Z93 (approved):

PR #98, opened 2022-12-14
Major refactoring of ArkhamDB deck importer

Buhallin:
- Splits all ArkhamDB interactions to a module
- Removes undeveloped code for Command Manager
- Removes bespoke logging in favor of standard SCED approach
- Removes mass load test functions
- Function documentation cleanup


src/arkhamdb/DeckImporterMain.ttslua
```diff
@@ -422,27 +166,11 @@ function loadCards(slots, investigatorId, customizations, playerColor, commandMa
...
-              if (adbCardInfo.xp ~= nil and adbCardInfo.xp > 0) then
-                cardName = cardName .. " (" .. adbCardInfo.xp .. ")"
-              end
-              debugPrint("Card not found: " .. cardName .. ", ArkhamDB ID " .. cardId, Priority.ERROR, playerColor)
-            else
-              debugPrint("Card not found in ArkhamDB, ID " .. cardId, Priority.ERROR, playerColor)
-            end
-          end)
+        arkhamDb.logCardNotFound(cardId, playerColor)
       end
```
Chr1Z93, line 170: Instead of using a separate `hadError` variable, the coroutine could end with `return 0` at this point

src/arkhamdb/DeckImporterMain.ttslua
```diff
@@ -422,27 +166,11 @@ function loadCards(slots, investigatorId, customizations, playerColor, commandMa
...
-            if (cardName ~= nil) then
-              if (adbCardInfo.xp ~= nil and adbCardInfo.xp > 0) then
-                cardName = cardName .. " (" .. adbCardInfo.xp .. ")"
-              end
-              debugPrint("Card not found: " .. cardName .. ", ArkhamDB ID " .. cardId, Priority.ERROR, playerColor)
-            else
-              debugPrint("Card not found in ArkhamDB, ID " .. cardId, Priority.ERROR, playerColor)
-            end
-          end)
+        arkhamDb.logCardNotFound(cardId, playerColor)
```
Chr1Z93, line 169: Instead of using a separate `hadError` variable, the coroutine could end with `return 0` at this point

src/arkhamdb/ArkhamDb.ttslua
```diff
@@ -0,0 +1,439 @@
+do
```
Chr1Z93, line 1: Should this file maybe have `Api` in its name to be in line with the other ones?

Chr1Z93 (approved):

src/arkhamdb/DeckImporterMain.ttslua
Chr1Z93: Ah, I missed the loop because it wasn't changed 😅
At the end of the coroutine the returned value should still probably not be `1` if there was an error.

PR #99, opened 2022-12-14
Coordinate cleaning and set aside snap points

Chr1Z93:
Cleaned up the slightly misaligned coordinates of playmats and miscelleaneous objects.
(I've not touched every single token, but most major objects that were misaligned.)

Edit: Snap points not added


argonui:
Looks good! I'll wait till I can load it to check & approve

Thanks for keeping it focused

Buhallin:
Can't test this right now, but how do the set aside snap points line up with the zones used by the deck loader?

Chr1Z93:
> Can't test this right now, but how do the set aside snap points line up with the zones used by the deck loader?

The snap points don't directly relate to the zones: Deck importer uses 3 rows, with a bit more distance between the cards.
These new points are for 4 cards directly next to the playermats.

Buhallin:
> > Can't test this right now, but how do the set aside snap points line up with the zones used by the deck loader?
> 
> The snap points don't directly relate to the zones: Deck importer uses 3 rows, with a bit more distance between the cards. These new points are for 4 cards directly next to the playermats.

Why not line them up with the importer points?  Not sure the extra row is all that big a deal, and it would avoid "Picked this up and snapped it back into a slightly different spot" jank

Chr1Z93:
> > > Can't test this right now, but how do the set aside snap points line up with the zones used by the deck loader?
> > 
> > 
> > The snap points don't directly relate to the zones: Deck importer uses 3 rows, with a bit more distance between the cards. These new points are for 4 cards directly next to the playermats.
> 
> Why not line them up with the importer points? Not sure the extra row is all that big a deal, and it would avoid "Picked this up and snapped it back into a slightly different spot" jank

Main reason is that table space is very limited (and since the deck importer spawns piles I need to lay them each out anyway - let me know if you have a different approach to this).

Buhallin:
> > > > Can't test this right now, but how do the set aside snap points line up with the zones used by the deck loader?
> > > 
> > > 
> > > The snap points don't directly relate to the zones: Deck importer uses 3 rows, with a bit more distance between the cards. These new points are for 4 cards directly next to the playermats.
> > 
> > 
> > Why not line them up with the importer points? Not sure the extra row is all that big a deal, and it would avoid "Picked this up and snapped it back into a slightly different spot" jank
> 
> Main reason is that table space is very limited (and since the deck importer spawns piles I need to lay them each out anyway - let me know if you have a different approach to this).
My main concern is things like this...  This kind of positioning is actually pretty bad, because the snap point claims the card so it can't go back into the deck with the others.  It'll highlight the deck like it's going to merge them, but when you draw it the snap point goes "YOINK!"
image
But this is also MUCH more compact than we typically do for snap points:
image

I don't think table space is THAT limited, and snapping to those points feels very cramped and not in line with the rest of the mod's styling.  Green definitely has it worse, but I think we can change up the victory display to help them out there.  Especially once the counter is in place, a single defined row that either snaps to about 3 piles, or doesn't snap at all, should do the job.  The current display has almost a full card's gap between each of the points.

IMHO it would be better to sync up the loader and snap points, in either direction.  The loader zones were specifically laid out to match the spacing between the card on the playmat snap points, so I think that's good for consistency of style.  Rather than 4 cramped ones, add 7 points matching the zones for the loader and we'll help Green out another way.

Buhallin:
Red gets it bad because of the campaign guides, not 100% sure how to fix that one.

Buhallin:
And, alternately, we could update the loader's zones to match these snap points.  I still think these look really cramped and keeping the spacing consistent is better, but it's an option.

Chr1Z93:
> Red gets it bad because of the campaign guides, not 100% sure how to fix that one.

Yes, exactly. Red and green don't have much room and I wanted to be consistent with these points.
Ideally, they would be the same zones that the deck importer uses, but right now that is not helpful for those two players.

Snap points can be spread out more when those two playermats get moved down quite a bit, but that would be a separate PR.

> My main concern is things like this... This kind of positioning is actually pretty bad, because the snap point claims the card so it can't go back into the deck with the others. It'll highlight the deck like it's going to merge them, but when you draw it the snap point goes "YOINK!"

That's indeed a bit annoying for set-aside decks, but those are quite rare. Stuff like "permanents" will not stay in a deck (at least it shouldn't...)

Chr1Z93:
I think this is a good case for a beta - would be very helpful to know how people handle their set aside cards. From personal experience I can add that these points are highly sought after by my friends.

Buhallin (approved):

PR #104, opened 2022-12-17
Option Panel: Part 3 - Addition of fan made accessories options & code improvement

Chr1Z93:
The option panel now includes options for spawning fan-made accessories, for now the tools I added are:
- Clean Up Helper
- Token Arranger
- Hand Helper
- Chaos Bag Manager

UI has been moved to the bottom right, is now animated and does not contain placeholder elements anymore.

Internal: Options are no longer referenced by numbers and now use a string as index

While working on the application of the options, a few other objects required small changes:
- Hand Helper: added function for external updating of color
- Token Arranger (and related: Bless/ Curse Manager): removed the option to change the tag that is used for detecting the laid out tokens

Snapshot of UI:
image

### DIFF
--- a/src/core/Global.ttslua
+++ b/src/core/Global.ttslua
@@ -11,6 +11,9 @@ optionPanel = {}
 -- GUID of data helper
 DATA_HELPER_GUID = "708279"
 
+-- GUID of fan-made accessories bag (also just called "barrel")
+BARREL_GUID = "aa8b38"
+
 -- GUIDs that will not be interactable (e.g. parts of the table)
 local NOT_INTERACTABLE = {
   "6161b4",
@@ -878,23 +881,20 @@ end
 function spawnHelperObject(name, position, rotation, color)
   if rotation == nil then rotation = {0, 270, 0} end
 
-  local barrel = getObjectFromGUID("aa8b38")
-  local barrelCopy = barrel.clone({position = barrel.getPosition() + Vector(0, 3, 0)})
-
-  for _, obj in ipairs(barrel.getObjects()) do
-    if obj.name == name then
-      barrelCopy.takeObject({
-        index = obj.index,
+  for _, obj in ipairs(getObjectFromGUID(BARREL_GUID).getData().ContainedObjects) do
+    if obj["Nickname"] == name then
+      spawnObjectData({
+        data = obj,
         position = position,
         rotation = rotation,
         callback_function = function(object)
           if name == "Hand Helper" then
             Wait.time(function() object.call("externalColorChange", color) end, 1)
+          elseif name == "Token Arranger" then
+            Wait.time(function() object.call("layout") end, 1)
           end
-        end,
-        smooth = false
+        end
       })
-      barrelCopy.destruct()
       return
     end
   end
--- a/src/core/Global.ttslua
+++ b/src/core/Global.ttslua
@@ -634,52 +634,6 @@ function onClick_load()
   UI.hide('load_button')
 end
 
-function onClick_defaultSettings()
-  print("Dummy: Load default settings")
-end
-
-function onClick_toggleOption(_, id)
-  local state = self.UI.getAttribute("toggle" .. id, "isOn")
-
-  -- flip state (and handle stupid "False" value)
-  if state == "False" then
-    state = true
-  else
-    state = false
-  end
-
-  self.UI.setAttribute("toggle" .. id, "isOn", state)
-
-  id = tonumber(id)
-  optionPanel[id] = state
-  applyOptionPanelChange(id, state)
-end
-
--- sets the option panel to the correct state (corresponding to 'optionPanel')
-function updateOptionPanelState()
-  for id, enabled in pairs(optionPanel) do
-    if enabled then self.UI.setAttribute("toggle" .. id, "isOn", true) end
-  end
-end
-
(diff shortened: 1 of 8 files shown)


src/core/Global.ttslua
```diff
@@ -825,3 +779,146 @@ function urldecode(str)
...
+  elseif id == "useClueClickers" then
+    playmatAPI.clickableClues(state, "All")
+
+    -- update master clue counter
+    getObjectFromGUID("4a3aa4").setVar("useClickableCounters", state)
+  
+  -- option: Show token arranger
+  elseif id == "showTokenArranger" then
+    -- delete previously pulled out tokens
+    for _, token in ipairs(getObjectsWithTag("to_be_deleted")) do token.destruct() end
```
Buhallin, line 839: Don't know the timing on this, but seems like this should happen in onLoad or onSave of the arranger?

src/core/Global.ttslua
```diff
@@ -825,3 +779,146 @@ function urldecode(str)
...
+
+    spawnOrRemoveHelper(state, "Token Arranger", {-42.3, 1.4, -46.5})
+
+  -- option: Show clean up helper
+  elseif id == "showCleanUpHelper" then
+    spawnOrRemoveHelper(state, "Clean Up Helper", {-68, 1.6, 35.5})
+  
+  -- option: Show hand helper for each player
+  elseif id == "showHandHelper" then
+    spawnOrRemoveHelper(state, "Hand Helper", {-50.84, 1.6, 7.02}, {0, 270, 0}, "White")
```
Buhallin, line 849: There's a lot of specific positioning here.  Can we find existent objects and derive the position relative to that?  Method could take a GUID and a relative position (like "above") and calculate a meaningful position.  Hooking to other objects would mean not needing to update if we rearranged things (though deletion would break it, so... up to you)

src/core/Global.ttslua
```diff
@@ -825,3 +779,146 @@ function urldecode(str)
...
+end
+
+-- copies the specified tool (by name) from the barrel
+---@param name String Name of the object that should be copied
+---@param position Position Desired position of the object
+function spawnHelperObject(name, position, rotation, color)
+  if rotation == nil then rotation = {0, 270, 0} end
+
+  local barrel = getObjectFromGUID("aa8b38")
+  local barrelCopy = barrel.clone({position = barrel.getPosition() + Vector(0, 3, 0)})
```
Buhallin, line 883: Use getData().ContainedObjects and spawnObjectData to avoid having to copy the barrel and do a take

src/core/Global.ttslua
```diff
@@ -825,3 +779,146 @@ function urldecode(str)
...
+      return
+    end
+  end
+end
+
+-- removes the specified tool (by name) from the provided position
+---@param name String Name of the object that should be removed
+---@param position Position Position of the object
+function removeHelperObject(name, position)
+  local search = Physics.cast({
```
Buhallin, line 908: You only look for the helper at the original spawn position, so this will fail if the user moves it, right?  That actually seems like a pretty common occurrence.

Could fix this by storing the GUID(s) of spawned objects as the value of the table (rather than just a boolean).  nil = off, string/string array = on

Buhallin (changes requested):

src/core/Global.ttslua
Chr1Z93: Good point - moving this to `onDestroy` of the TA

src/core/Global.ttslua
Chr1Z93: Good idea! Will change this once we have a more "final" table layout.

src/core/Global.ttslua
Chr1Z93: That would indeed be a bit cleaner, but I can't get that to work. Data structure that is returned by `getData()` is not clear to me and of course not documentated...

src/core/Global.ttslua
Chr1Z93: Yes, that's how I implemented it at the moment.

Storing the GUID would be possible, but only removing from the default position feels a bit safer to me.
Otherwise the custom set position would be gone.

src/core/Global.ttslua
Chr1Z93: Yeah, that is a better solution! Changed it

src/core/Global.ttslua
Buhallin: If that happens though, then the option panel doesn't do what it says.  If a user disables an option, it would remain in play and the operation not only fails but the presentation is inconsistent.  And if they re-enable it, now they've got multiples :)

I don't think "Turning this on places it at X and then you can move it" is the worst thing.  IMHO the user signal to remove the object is much stronger than repositioning it, and if it is done in error moving it back where they want it is easy for them.

If keeping the position is important, can always store the position before destroying, and spawn the new one there if they re-enable it.  Though with our inability to persist anything across new versions/saves, the object appearing at its default spot is going to be the behavior most of the time anyway, so there's probably a consistency win.

Chr1Z93: Storing the GUID of the spawned object is the best solution imo.
Since that needs a change to the save function, I'd like to mvoe that into the next part.

Buhallin (approved):

PR #106, opened 2022-12-17
Create a unified panel for groups of player cards.

Buhallin:
The will replace the investigator boxes, class card spread tokens, and weakness container.  It will (in the future) replace the bonded and upgrade sheet bags.

This is a code-only submit while we wait on final images for the new panel.

### DIFF
--- a/src/arkhamdb/ArkhamDb.ttslua
+++ b/src/arkhamdb/ArkhamDb.ttslua
@@ -67,13 +67,13 @@ do
 
     local deck = Request.start(deckUri, function(status)
       if string.find(status.text, "<!DOCTYPE html>") then
-        printToAll("Private deck ID " .. deckId .. " is not shared", playerColor)
+        internal.maybePrint("Private deck ID " .. deckId .. " is not shared", playerColor)
         return false, table.concat({ "Private deck ", deckId, " is not shared" })
       end
       local json = JSON.decode(status.text)
 
       if not json then
-        printToAll("Deck ID " .. deckId .. " not found", playerColor)
+        internal.maybePrint("Deck ID " .. deckId .. " not found", playerColor)
         return false, "Deck not found!"
       end
 
@@ -101,9 +101,9 @@ do
           if (adbCardInfo.xp ~= nil and adbCardInfo.xp > 0) then
             cardName = cardName .. " (" .. adbCardInfo.xp .. ")"
           end
-          printToAll("Card not found: " .. cardName .. ", ArkhamDB ID " .. cardId, playerColor)
+          internal.maybePrint("Card not found: " .. cardName .. ", ArkhamDB ID " .. cardId, playerColor)
         else
-          printToAll("Card not found in ArkhamDB, ID " .. cardId, playerColor)
+          internal.maybePrint("Card not found in ArkhamDB, ID " .. cardId, playerColor)
         end
       end)
   end
@@ -132,7 +132,7 @@ do
       return
     end
 
-    printToAll(table.concat({ "Found decklist: ", deck.name }), playerColor)
+    internal.maybePrint(table.concat({ "Found decklist: ", deck.name }), playerColor)
 
     log(table.concat({ "-", deck.name, "-" }))
     for k, v in pairs(deck) do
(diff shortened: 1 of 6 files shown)


src/arkhamdb/ArkhamDb.ttslua
```diff
@@ -304,6 +304,12 @@ do
     end
   end
 
+  internal.maybePrint = function(message, playerColor)
```
Chr1Z93, line 307: Is there something else in `internal` or why is this not just called `maybePrint`?

src/playercards/PlayerCardPanel.ttslua
```diff
@@ -0,0 +1,509 @@
...
+
+local FACE_UP_ROTATION = { x = 0, y = 270, z = 0}
+local FACE_DOWN_ROTATION = { x = 0, y = 270, z = 180}
+
+-- Coordinates to begin laying out cards to match the reserved areas of the
+-- table.  Cards will lay out horizontally, then create additional rows
+local START_POSITIONS = {
+  skill = Vector(58.384, 1.36, 92.4),
+  event = Vector(53.229, 1.36, 92.4),
+  asset = Vector(40.960, 1.36, 92.4),
```
Chr1Z93, line 19: Should probably round the values to two decimals

src/playercards/PlayerCardPanel.ttslua
```diff
@@ -0,0 +1,509 @@
...
+  self.createButton({
+      label="Clear", click_function="deleteAll", function_owner=self,
+      position={0.5,0.2,0.9}, rotation={0,0,0}, height=BUTTON_WIDTH, width=BUTTON_HEIGHT,
+      font_size=64, color={0,0,0}, font_color={1,1,1}, scale={0.25, 0.25, 0.25}
+  })
+  self.createButton({
+      label="Weaknesses", click_function="spawnWeaknesses", function_owner=self,
+      position={-0.5,0.2,0.9}, rotation={0,0,0}, height=BUTTON_WIDTH, width=BUTTON_HEIGHT,
+      font_size=64, color={0,0,0}, font_color={1,1,1}, scale={0.25, 0.25, 0.25}
+  })
```
Chr1Z93, line 213: The button creation begs for usage of a variable with common parameters and/or a loop to create these.
It would also be better to have a click_function with parameters instead of a separate function for each button (can be achieved by `setVar`)

src/playercards/PlayerCardSpawner.ttslua
```diff
@@ -3,6 +3,9 @@
 -- Note that the table rotation is weird, and the X axis is vertical while the
 -- Z axis is horizontal
 local SPREAD_Z_SHIFT = -2.3
+local SPREAD_X_SHIFT = -3.66
+
+
```
Chr1Z93, line 8: Nit: remove empty lines

Chr1Z93 (changes requested):

src/arkhamdb/ArkhamDb.ttslua
Buhallin: Internal is used to keep the method from leaking out of the module.  As we get more and more modules the potential for name collisions gets higher.

I'm not sure if I just haven't puzzled out the Lua or TTS handles it badly, but making a method like this local (which is basically the pattern I'm looking for) doesn't work.  I gave up trying to puzzle it out, so this may not be the best way to do it but it works.

src/playercards/PlayerCardPanel.ttslua
Buhallin: I've been going to three (unless the value I'm going for falls very close to a different one) because that's the standard we use for rounding with ModManager.  In this case, the values were taken directly from the card spreads, which were taken from the previous memory bags, to align with the decals.  Less an issue once we remove them, but broadly speaking I think as long as we're matching to that 3-place rounding, this should be a no-comment.

src/playercards/PlayerCardPanel.ttslua
Buhallin: Absolutely.  But the button creation is all going to need to be rewritten once the new UI panel is added, so I didn't think it was worth it yet since it was all going to change anyway.

I do already loop the function creation for things that I knew the values for.  Release names I was going to let CZ define while adding the data rather than enforcing it on him ahead of time.

src/playercards/PlayerCardSpawner.ttslua
Buhallin: Will fix tonight.

Chr1Z93 (approved):

PR #107, opened 2022-12-19
SCED Intro Tour, Part 3

Buhallin:
Adds additional capabilities for tour cards:
- Option to specify exact position for the camera, rather than an object.  This will enable looking at items which are not their own objects, such as the draw chaos token button.
- Option to specify height for object look.  This acts as a camera zoom; we cannot control the zoom at all, much less when attached.  By controlling the height when we're looking at an object we simulate zoom.
- Right/Left narrator image.  This will flip the card so the narrator's image is on the right or left, adding some variety to the presentation.
- Cardinal points for card position.  In addition to just adding variety, this will allow the card to be placed closer to where you want it, including other Global XML elements.


src/core/tour/TourManager.ttslua
```diff
@@ -19,6 +20,19 @@ do
...
+  local SCREEN_POSITIONS = {
+    center = "0 0 0",
+    north = "0 300 0",
+    east = "600 0 0",
+    west = "-600 0 0",
+    south = "0 -300 0",
+    northwest = "-600 300 0",
+    northeast = "600 300 0",
+    southwest = "-600 -300 0",
+    southeast = "600 -300 0"
```
Chr1Z93, line 33: Are these values universally applicable or dependant on screen resolution?

src/core/tour/TourManager.ttslua
```diff
@@ -84,10 +98,21 @@ do
...
+    elseif TOUR_SCRIPT[cardIndex].showPos ~= nil then
+      lookPos = TOUR_SCRIPT[cardIndex].showPos
+    end
+    if lookPos ~= nil then
       Wait.time(function()
-        local lookAtObj = getObjectFromGUID(TOUR_SCRIPT[cardIndex].showObj)
-        hook.setPositionSmooth(lookAtObj.getPosition(), false, false)
+        hook.setPositionSmooth(lookPos, false, false)
       end, delay)
       delay = delay + 0.5
```
Chr1Z93, line 117: Does increasing the delay actually do anything since it gets set to 0.5 before?

Chr1Z93 (changes requested):

src/core/tour/TourManager.ttslua
Buhallin: I'm honestly not sure.  Will have to test this more to see.

src/core/tour/TourManager.ttslua
Buhallin: Yes - this is the delay until the next card is shown (used below) and needs to be different based on whether the camera is centering, or is moving to a new spot (each of which takes half a second)

src/core/tour/TourManager.ttslua
Buhallin: There's still going to be a few more passes on this, including testing, so would prefer not to block this one at the moment.

Chr1Z93 (approved):

PR #109, opened 2022-12-19
Playermat: Allow non-numeric IDs for upkeep // Addition of promo and revised art for investigators as states

Chr1Z93:
TTS Zoop will add custom IDs to investigators and minicards to allow the upkeep button to flip them.
To avoid conflicting with arkhamDB, the IDs will start with a "Z", e.g. "Z1234".

This implements this feature into the playmat.

Additional feature:
- promo art added as additional state to existing cards (for the investigator card and the mini-card)
- investigator card have the same id
- mini-cards all have the same id
- following investigators are updated with states:
image

- revised art has also been added as state for the core investigators
image


src/playermat/Playmat.ttslua
```diff
@@ -257,10 +257,10 @@ function doUpkeep(_, color, alt_click)
     end
   end
 
-  -- flip investigator mini-card if found
-  -- flip summoned servitor mini-cards (To-Do: don't flip all of them)
+  -- flip investigator mini-card and summoned servitor mini-card
+  -- (all characters allowed to account for custom IDs - e.g. 'Z0000' for TTS Zoop generated IDs)
```
Buhallin, line 261: Can we fix the -promo issue here as well, as long as we're changing this?

Chr1Z93: We could account for it here, but I thought we wanted to do that by having them as states of the same card.
(With both having the same ID and also states for the mini card)

Buhallin:
We should add Revised Core as states as well

Buhallin (approved):

PR #113, opened 2022-12-19
Move official content to require for downloadboxes

Chr1Z93:
- more consistent usage of the file `src/DownloadBox` for all official content
- look of download button is not changed, script will decide based on gmnotes, which position / size the button has


src/core/DownloadBox.ttslua
```diff
@@ -1,17 +1,41 @@
 function onLoad()
-  self.createButton({
+  notes = self.getGMNotes()
```
Buhallin, line 2: This will be a lot easier to test if it's pulled as part of the click method, rather than just on load.  What's the reason for the move?

Buhallin (changes requested):

src/core/DownloadBox.ttslua
Chr1Z93: Notes get grabbed up here because there are different values for e.g. position and font size which will be detected by the notes too

Buhallin: Ah, I see it now.  Fair, but no reason we can't re-grab it in the click to handle both?

Chr1Z93: Indeed, changed it to re-grab on click

Buhallin (approved):

PR #115, opened 2022-12-19
Custom Playmat Images: Account for long label

Chr1Z93:
- reduced the longest label by shortening `(Fan-Made)` to `(FM)`
- reduced `font_size` from 100 (default value) to 90


Buhallin (approved):
I'm not sure the meaning of FM is going to be clear, and seems quite possible we'll get a straight-up name long enough to break the size limit in the future.

Chr1Z93:
> I'm not sure the meaning of FM is going to be clear, and seems quite possible we'll get a straight-up name long enough to break the size limit in the future.

Yeah, this will get reworked into a proper XML UI in the future (with a real display of images instead of these placeholder objects).
Until then, this bandaid fix should be fine.

PR #117, opened 2022-12-20
Option Panel: Part 4 - Rebuilding UI for scrollview // Token Arranger: Bugfix

Chr1Z93:
- option panel UI had to be rebuilt to use a scrollview
- added toggle switch images (not yet used in UI)
- header images from `img` folder are used
- navigation overlay added as option (will just spawn the bag with instructions at the top right of the play area)
- Token Arranger: Since the option allows to quickly place it and then remove it, I had to add some error handling for removal during the placement
- Snapshot of UI:
image

### DIFF
--- a/modsettings/CustomUIAssets.json
+++ b/modsettings/CustomUIAssets.json
@@ -75,13 +75,23 @@
     "URL": "http://cloud-3.steamusercontent.com/ugc/2027213118466516005/113C19D37CFFA9E554394FD5B11B32967F846A62/"
   },
   {
-    "Name": "option_image1",
+    "Name": "option_header1",
     "Type": 0,
-    "URL": "http://cloud-3.steamusercontent.com/ugc/2027213118468280563/911BA319CD7502258D570B23BD51B3D8DA8B2AC3/"
+    "URL": "http://cloud-3.steamusercontent.com/ugc/2024962321894443102/5727544F24E8C4546031F4DDE8BD4B45D93AC3FC/"
   },
   {
-    "Name": "option_image2",
+    "Name": "option_header2",
     "Type": 0,
-    "URL": "http://cloud-3.steamusercontent.com/ugc/2027213118470839572/FB133C41A6D8915A44C138BCF947ECFE3D111813/"
+    "URL": "http://cloud-3.steamusercontent.com/ugc/2024962321894443022/3262DF2ABAE8D9A2BF64ECB74C1BB34B33814554/"
+  },
+  {
+    "Name": "option_on",
+    "Type": 0,
+    "URL": "http://cloud-3.steamusercontent.com/ugc/2024962321889555728/22ABD35CBB49A001F3A5318E4AFCFB22D24FEA39/"
+  },
+  {
+    "Name": "option_off",
+    "Type": 0,
+    "URL": "http://cloud-3.steamusercontent.com/ugc/2024962321889555661/6643E5CC9160FF4624672C255D0DF7B313DA00A5/"
   }
 ]
--- a/objects/Fan-MadeAccessories.aa8b38/TokenArranger.022907.ttslua
+++ b/objects/Fan-MadeAccessories.aa8b38/TokenArranger.022907.ttslua
@@ -193,37 +193,22 @@ function layout(_, _, isRightClick)
     -- stop here if right-clicked
     if isRightClick then return end
 
-    local chaos_bag = getChaosBag()
-    local chaos_bag_objects = chaos_bag.getObjects()
-
-    -- take each token out and clone it
-    for _, data in ipairs(chaos_bag_objects) do
-        chaos_bag.takeObject {
-            guid = data.guid,
-            smooth = false,
-            callback_function = function(tok)
-                chaos_bag.putObject(tok.clone())
-                tok.addTag("to_be_deleted")
-            end
-        }
-    end
-
-    -- wait until all tokens have finished spawning
-    Wait.condition(function() placeTokens() end, function() return #chaos_bag_objects == #getObjectsWithTag("to_be_deleted") end)
-end
-
--- position tokens sorted by value
-function placeTokens()
+    local chaosBag = getChaosBag()
     local data = {}
 
-    -- create table with tokens
-    for i, token in ipairs(getObjectsWithTag("to_be_deleted")) do
-        local name = token.getName() or ""
-        local value = tonumber(name)
-        local precedence = TOKEN_PRECEDENCE[name]
+    -- clone tokens from chaos bag (default position above trash can)
+    for i, obj in ipairs(chaosBag.getData().ContainedObjects) do
+        obj["Tags"] = { "to_be_deleted" }
+        local spawnedObj = spawnObjectData({
+            data = obj,
+            position = { 0.49, 3, 0 }
+        })
+
+        local value = tonumber(obj["Nickname"])
+        local precedence = TOKEN_PRECEDENCE[obj["Nickname"]]
 
         data[i] = {
-            token = token,
+            token = spawnedObj,
             value = value or precedence[1]
         }
 
@@ -237,11 +222,17 @@ function placeTokens()
     -- sort table by value (symbols last if same value)
     table.sort(data, token_value_comparator)
 
+    -- error handling for removal of token arranger
+    if self == nil then
+        for _, token in ipairs(getObjectsWithTag("to_be_deleted")) do token.destruct() end
+        return
+    end
+    
     -- laying out the tokens
     local pos           = self.getPosition() + Vector(3.55, -0.05, -3.95)
     local location      = { x = pos.x, y = pos.y, z = pos.z }
     local current_value = data[1].value
-
+    
     for _, item in ipairs(data) do
         if item.value ~= current_value then
             location.x = location.x - 1.75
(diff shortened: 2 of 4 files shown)


Buhallin:
Won't be able to look at it until later tonight, but that looks like a horizontal scrollbar at the bottom?  Should definitely avoid that... horizontal scrolling is the devil's tool, and not the cool TCU-Tarot devil! 👿

Chr1Z93:
> 



> Won't be able to look at it until later tonight, but that looks like a horizontal scrollbar at the bottom? Should definitely avoid that... horizontal scrolling is the devil's tool, and not the cool TCU-Tarot devil! 👿

Unfortunately that is unavoidable thanks to TTS implementation of XML scrollviews - it is not functional though

modsettings/CustomUIAssets.json
```diff
@@ -75,13 +75,23 @@
     "URL": "http://cloud-3.steamusercontent.com/ugc/2027213118466516005/113C19D37CFFA9E554394FD5B11B32967F846A62/"
   },
   {
-    "Name": "option_image1",
+    "Name": "option_header1",
```
Buhallin, line 78: Better to name these what they are, and then they just get used.  Maybe like "header_core" and "header_acolyte"?

objects/Fan-MadeAccessories.aa8b38/TokenArranger.022907.ttslua
```diff
@@ -237,11 +222,17 @@ function placeTokens()
     -- sort table by value (symbols last if same value)
     table.sort(data, token_value_comparator)
 
+    -- error handling for removal of token arranger
```
Buhallin, line 225: Seems like this would be better in onDestroy()?  (onDestruct, whatever that method is)

src/core/Global.ttslua
```diff
@@ -857,6 +857,10 @@ function applyOptionPanelChange(id, state)
   -- option: Show chaos bag manager
   elseif id == "showChaosBagManager" then
     spawnOrRemoveHelper(state, "Chaos Bag Manager", {-67.8, 1.4, -49.5})
+
+  -- option: Show navigation overlay
+  elseif id == "showNavigationOverlay" then
+    spawnOrRemoveHelper(state, "jaqenZann's Navigation Overlay", {-11.7, 1.4, -15})
```
Buhallin, line 863: We were going to remove the author names on helpers, right?

xml/OptionPanel.xml
```diff
@@ -1,108 +1,182 @@
...
-  <Text color="white" alignment="MiddleLeft"/>
+  <Text color="#FFFFFF" alignment="MiddleLeft"/>
   <Toggle isOn="False" rectAlignment="MiddleRight"/>
+  <Cell dontUseTableCellBackground="true" outlineSize="0 1" outline="grey"/>
 
-  <VerticalLayout class="window" visibility="Admin" active="false" color="black" width="500" height="700" outlineSize="1 1" outline="grey" showAnimation="SlideIn_Right" hideAnimation="SlideOut_Right" animationDuration="0.1"/>
-  <VerticalLayout class="group" visibility="Admin" outlineSize="1 1" outline="grey" color="#222222" padding="10" spacing="5"/>
-  <VerticalLayout class="text-column" visibility="Admin" padding="5 20 0 0"/>
+  <!-- main window -->
+  <TableLayout class="window" width="500" height="700" active="false" color="#000000" outlineSize="2 2" outline="grey" showAnimation="SlideIn_Right" hideAnimation="SlideOut_Right" animationDuration="0.1"/>
```
Buhallin, line 8: Don't have to do it in this PR, but please consider a format pass here.  It's common with XML to split attributes to different lines, which greatly enhances readability.  One attribute is probably fine to do inline, but 2 or more should be on separate lines.

xml/OptionPanel.xml
```diff
@@ -1,108 +1,182 @@
...
-          <Text class="option-header">Show "Hand Helper"</Text>
-          <Text class="description">Never count your hand cards again! This tool does that for you and can even take "Dream-Enhancing Serum" into account. Also includes a button for randomly discard a card.</Text>
-        </VerticalLayout>
-        <Toggle id="showHandHelper" onValueChanged="onClick_toggleOption(showHandHelper)"/>
-      </HorizontalLayout>
+          <!-- Option: show token arranger -->
+          <Row class="option-text">
+            <Cell class="option-text">
+              <VerticalLayout class="text-column">
+                <Text class="option-header">Show "Token Arranger"</Text>
```
Buhallin, line 108: I think we can get rid of the repetitive "Show" throughout these, and the quotes around the item.  Just Token Arranger (et al) should do the job, the showing/enabling is implied with the checkbox.

xml/OptionPanel.xml
```diff
@@ -1,108 +1,182 @@
...
+              <Toggle id="showCleanUpHelper" onValueChanged="onClick_toggleOption(showCleanUpHelper)"/>
+            </Cell>
+          </Row>
+
+          <!-- Option: show hand helper -->
+          <Row class="option-text">
+            <Cell class="option-text">
+              <VerticalLayout class="text-column">
+                <Text class="option-header">Show "Hand Helper"</Text>
+                <Text class="description">Never count your hand cards again! This tool does that for you and can even take "Dream-Enhancing Serum" into account. Also includes a button for randomly discard a card.</Text>
```
Buhallin, line 135: Grammar nit: "can even take"... sounds like it's possible but doesn't necessarily.  "will take" or "will handle Dream-Enhancing Serum" (which is a bit shorter)

xml/OptionPanel.xml
```diff
@@ -1,108 +1,182 @@
...
+              <Toggle id="showChaosBagManager" onValueChanged="onClick_toggleOption(showChaosBagManager)"/>
+            </Cell>
+          </Row>
+
+          <!-- Option: show navigation overlay -->
+          <Row class="option-text">
+            <Cell class="option-text">
+              <VerticalLayout class="text-column">
+                <Text class="option-header">Show "Navigation Overlay"</Text>
+                <Text class="description">This enables an overlay for quickly navigating to various points on the table.</Text>
```
Buhallin, line 161: Maybe the description here should explicitly specify that it's moving the camera around?

xml/OptionPanel.xml
```diff
@@ -1,108 +1,182 @@
...
-      <HorizontalLayout class="group-content">
-        <VerticalLayout class="text-column">
-          <Text class="option-header">Show "Chaos Bag Manager"</Text>
-          <Text class="description">Panel for easy addition or removal of chaos tokens to the bag - very useful for EotE because of Frost tokens!</Text>
-        </VerticalLayout>
-        <Toggle id="showChaosBagManager" onValueChanged="onClick_toggleOption(showChaosBagManager)"/>
+  <!-- Buttons: Defaults and Close -->
+  <Row preferredHeight="50">
+    <Cell>
+      <HorizontalLayout minHeight="50" flexibleHeight="0" padding="5" spacing="5">
```
Buhallin, line 176: Minor, but I think the buttons don't need to be this large.  Lots of dead space on them... Maybe shrink them and leave empty space in between?

objects/Fan-MadeAccessories.aa8b38/TokenArranger.022907.ttslua
Chr1Z93: That method is getting used. This additional handling is required for the case that the token arranger is getting removed between spawning and positioning the additional tokens.

src/core/Global.ttslua
Chr1Z93: Yep, that is the plan but not part of this PR.

xml/OptionPanel.xml
Chr1Z93: Sure, can do so for the next update.

xml/OptionPanel.xml
Chr1Z93: Yeah, we could do that. I was cautious of unclear wording to make sure that everyone understands that this spawns an additional object. It's not unthinkable to have a tool that does not have an interactable object.

xml/OptionPanel.xml
Chr1Z93: Reason is that DES still requires the user to tick a box (though this could actually be handled by script - not sure if that would be worth it though)

xml/OptionPanel.xml
Chr1Z93: Does "navigating to points on the table" not imply camera movement?

xml/OptionPanel.xml
Chr1Z93: Agreed, gonna change that in the next update

modsettings/CustomUIAssets.json
Chr1Z93: Changed in part 5 to avoid conflict

xml/OptionPanel.xml
Buhallin: That's actually one of the reason I think we should get rid of "show".  As far as the options go you're just enabling or disabling something.  What form that takes shouldn't be relevant, though I can see both sides.  Going to resolve this and leave it up to you.

xml/OptionPanel.xml
Buhallin: Maybe?  "Navigate" feels like an overloaded word, and a bit ambiguous about what is being navigated.  It also doesn't match the core TTS terminology of camera save states.

xml/OptionPanel.xml
Buhallin: We should change that to be automatic detection :)  But fine for now.

xml/OptionPanel.xml
Buhallin: Leaving this one up to you as well, and resolving

Buhallin (approved):

PR #118, opened 2022-12-20
Option Panel: Part 5 - Storing GUIDs for spawned helper objects to allow later removal at any position

Chr1Z93:
- save GUIDs of spawned helper object to allow later removal even after moving them
- removed horizontal scrollbar
- removed "show" for fan-made accessories
- made buttons at the bottom smaller
- fixed XML formatting to use new lines

(This is built on top of Part 4).

Snapshot of UI:

image

### DIFF
--- a/src/core/Global.ttslua
+++ b/src/core/Global.ttslua
@@ -857,6 +857,10 @@ function applyOptionPanelChange(id, state)
   -- option: Show chaos bag manager
   elseif id == "showChaosBagManager" then
     spawnOrRemoveHelper(state, "Chaos Bag Manager", {-67.8, 1.4, -49.5})
+
+  -- option: Show navigation overlay
+  elseif id == "showNavigationOverlay" then
+    spawnOrRemoveHelper(state, "jaqenZann's Navigation Overlay", {-11.7, 1.4, -15})
   end
 end
 
--- a/src/core/Global.ttslua
+++ b/src/core/Global.ttslua
@@ -804,14 +804,15 @@ function onClick_toggleOption(_, id)
   end
 
   self.UI.setAttribute(id, "isOn", state)
-  optionPanel[id] = state
   applyOptionPanelChange(id, state)
 end
 
 -- sets the option panel to the correct state (corresponding to 'optionPanel')
 function updateOptionPanelState()
   for id, enabled in pairs(optionPanel) do
-    if enabled then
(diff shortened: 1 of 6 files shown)


src/core/Global.ttslua
```diff
@@ -804,14 +804,15 @@ function onClick_toggleOption(_, id)
...
-  optionPanel[id] = state
   applyOptionPanelChange(id, state)
 end
 
 -- sets the option panel to the correct state (corresponding to 'optionPanel')
 function updateOptionPanelState()
   for id, enabled in pairs(optionPanel) do
-    if enabled then
+    if (type(enabled) == "boolean" and enabled) or
+    (type(enabled) == "string" and enabled) or
```
Buhallin, line 814: Formatting nit: indent the continuation of the previous lines (2x standard indent is common)

src/core/Global.ttslua
```diff
@@ -841,56 +845,55 @@ function applyOptionPanelChange(id, state)
...
-  if rotation == nil then rotation = {0, 270, 0} end
-
   for _, obj in ipairs(getObjectFromGUID(BARREL_GUID).getData().ContainedObjects) do
     if obj["Nickname"] == name then
-      spawnObjectData({
+      return spawnObjectData({
         data = obj,
         position = position,
-        rotation = rotation,
+        rotation = rotation or {0, 270, 0},
```
Buhallin, line 896: The object's data should have a rotation to it, which we should use rather than hardcoding the 270.  If you just leave this off it will use the rotation defined in obj

Chr1Z93: This is used for overriding the position for e.g. spawning a correctly rotated hand helper for each player

src/core/Global.ttslua
Chr1Z93: fixed

src/core/Global.ttslua
Buhallin: Right - got that :)  Wasn't worried about using the passed rotation.  But if the rotation isn't provided, you're overriding it with 0,270,0 instead of the object's actual rotation when it went into the bag.  270 may not be the right "up" for some objects.

Not going to block on this, but please consider a cleanup.

Buhallin (approved):

src/core/Global.ttslua
Chr1Z93: Alright, changing this in part 6

PR #120, opened 2022-12-21
Moving token bags a bit up to use free space (plus automatic formatting by MM)

Chr1Z93:
(This includes some automatic JSON reformatting by the TTS Mod Manager by building&decomposing)

Since the clue counter swapper is no longer there, this moves the token bags a bit up to use the free space:
image


Buhallin:
So this is definitely personal, but I'd prefer not to do this.  That empty space between the playmats has utility for the white and orange players to share something - we use it for copies of the scenario token card.  I don't think there's a strong reason to compress the space, but I'm open to a discussion on it.

Chr1Z93:
> So this is definitely personal, but I'd prefer not to do this. That empty space between the playmats has utility for the white and orange players to share something - we use it for copies of the scenario token card. I don't think there's a strong reason to compress the space, but I'm open to a discussion on it.

I also like to put copies of the scenario card near each players chaos token button, but I don't think the current version has enough space for what you are suggesting:

image

I am very open to optimizing the table space (especially needed for red / green!), but this PR does not change the functionality of the bags (since there is not really space left anyway).

Buhallin:
> > So this is definitely personal, but I'd prefer not to do this. That empty space between the playmats has utility for the white and orange players to share something - we use it for copies of the scenario token card. I don't think there's a strong reason to compress the space, but I'm open to a discussion on it.
> 
> I also like to put copies of the scenario card near each players chaos token button, but I don't think the current version has enough space for what you are suggesting:
> 
> image
> 
> I am very open to optimizing the table space (especially needed for red / green!), but this PR does not change the functionality of the bags (since there is not really space left anyway).

Hrm, I don't think I'd used the new version yet then (we just finished a campaign off the older save) so I hadn't realized that.  I think I tried to keep them from moving up (unsuccessfully) in the first place :)  I would like to consider ways to reduce the space from those token pools, but this is fine for now.

Buhallin (approved):
Let's try and make sure we match to ModManager's formatting so we don't end up with these large unrelated cleanups in the future.

PR #122, opened 2022-12-22
Card: Short Supply - Moved function from "Hand Helper" directly to card // PlaymatAPI: New functions for playermat getting

Chr1Z93:
Short Supply:
- added scripting for "Short Supply" instead of doing so `onHover` via Hand Helper
- now only displays one option: "Discard 10 cards"

Hand Helper:
- removed this function from Hand Helper and did some minor reformatting
- bugfix for Hand Helper: raised buttons slightly to not flicker when viewed from directly above
- added tag to be ingnored by clean up helper

PlaymatAPI:
- added functions to playermatapi to return requested playmat by color or position


objects/AllPlayerCards.15bb07/ShortSupply.e5f541.ttslua
```diff
@@ -0,0 +1,57 @@
+local MAT_GUIDS = { "8b081b", "bd0ff4", "383d8b", "0840d5" }
```
Buhallin, line 1: Should update this to use the Playermat API

Buhallin (changes requested):

objects/AllPlayerCards.15bb07/ShortSupply.e5f541.ttslua
Buhallin: I was unclear, sorry :)  I didn't mean just this line, I meant most of the operations here.  This gets the mat ref through the API, but after that it's still doing the operations via call() and getVar on the mat itself.

getMatByPosition() should be a function in the API, as it's likely to see general use (e.g. customized cards will need it).  The actual operation here (discarding X from the top of the deck) could be generalized through the API as well, maybe something like discardFromDeck(count, completePartial) - second parameter there specifying whether the operation should continue if the deck is < count.

A good rule of thumb is that anywhere we see call() or setVar() or getVar() etc we should seriously look at moving those operations to the Playermat.

Buhallin (changes requested):

objects/AllPlayerCards.15bb07/ShortSupply.e5f541.ttslua
Chr1Z93: That makes sense. I'll update this PR when I get to it.

objects/AllPlayerCards.15bb07/ShortSupply.e5f541.ttslua
```diff
@@ -0,0 +1,36 @@
...
+function shortSupply(color)
+  local matColor = playmatAPI.getMatColorbyPosition(self.getPosition())
+
+  -- get draw deck and discard position
+  local drawDeck = playmatAPI.getDrawDeck(matColor)
+  local discardPos = playmatAPI.getDiscardPosition(matColor)
+
+  -- error handling
+  if discardPos == nil then
+    broadcastToAll("Couldn't retrieve discard position from playermat!", "Red")
```
Buhallin, line 17: Should probably limit the error to the calling player as well

src/playermat/PlaymatApi.ttslua
```diff
@@ -3,26 +3,70 @@ do
...
     Orange = "3f22e5",
-    Green = "891403",
-    Red = "4111de"
+    Green  = "891403",
+    Red    = "4111de"
   }
 
+  -- Returns the by color requested playermat as object
+  ---@param matColor String Color of the playermat to return
+  PlaymatApi.getMatbyColor = function(matColor)
```
Buhallin, line 28: Is anything using this?  A major point of the API is that it's an interface to get to other objects, returning the actual one seems counter to that.

src/playermat/PlaymatApi.ttslua
```diff
@@ -3,26 +3,70 @@ do
...
+        return "Green"
+      else
+        return "Red"
+      end
+    end
+  end
+
+  -- Returns the draw deck of the requested playmat
+  ---@param matColor String Color of the playermat
+  PlaymatApi.getDrawDeck = function(matColor)
```
Buhallin, line 57: Not worth blocking on here given our time constraint, but I'd remove this and add a "discard(count)" method to playmat, that's exposed from here.  That would make the playmat own its own components and handle operations on them, rather than passing them out for some unknown code to manipulate.

Buhallin (changes requested):

src/playermat/PlaymatApi.ttslua
Chr1Z93: Not yet - I added it for completeness.

objects/AllPlayerCards.15bb07/ShortSupply.e5f541.ttslua
Chr1Z93: Will do

src/playermat/PlaymatApi.ttslua
Chr1Z93: Yeah, that will be useful once we have more scripting regarding discarding - let's defer this for now.

src/playermat/PlaymatApi.ttslua
Buhallin: I think we should remove this.  Not only isn't it being used I think it's something we don't want to be used.  So if we put it in now it's either dead code until we do remove it, or gets used and is something else we have to clean up.

Chr1Z93: Done

Buhallin (approved):

PR #123, opened 2022-12-22
Option Panel: Part 6 - Addition of "load defaults" functionality

Chr1Z93:
- "Load Defaults"-button will reset everything to off (except for SnapTags) and reverse any changes that were made by enabling options
- only override rotation for spawned objects if there is a value for rotation provided (and keep objects rotation otherwise)
- because there is now an additional variable for playermats (showDrawButton) the MM splits the luascriptstate to separate files

### DIFF
--- a/src/core/Global.ttslua
+++ b/src/core/Global.ttslua
@@ -885,14 +885,22 @@ end
 ---@param name String Name of the object that should be copied
 ---@param position Position Desired position of the object
 function spawnHelperObject(name, position, rotation, color)
+  local barrel = getObjectFromGUID(BARREL_GUID)
+
+  -- error handling for missing barrel
+  if not barrel then
+    broadcastToAll("'Barrel' with fan-made accessories could not be found!", "Red")
+    return
+  end
+
   local spawnTable = {
     position = position,
     callback_function = function(object)
-          if name == "Hand Helper" then
-            Wait.time(function() object.call("externalColorChange", color) end, 1)
-          elseif name == "Token Arranger" then
-            Wait.time(function() object.call("layout") end, 1)
-          end
+      if name == "Hand Helper" then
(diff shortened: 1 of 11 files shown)


src/core/Global.ttslua
```diff
@@ -888,26 +885,32 @@ end
...
-      })
+    end
+  }
+
+  -- only overrride rotation if there is one provided (object's rotation used instead)
+  if rotation then
+    spawnTable.rotation = rotation
+  end
+
+  for _, obj in ipairs(getObjectFromGUID(BARREL_GUID).getData().ContainedObjects) do
```
Buhallin, line 904: Not sure if this is new or I missed it before, but this will throw an error if the barrel happens to be deleted.  Would be better to check for a nil and send a broadcast, then ensure that the option state is accurate (i.e. off).  Can consider for a future enhancement, assuming we move the barrel somewhere hidden it's low-risk.

Buhallin (approved):

src/core/Global.ttslua
```diff
@@ -888,26 +885,32 @@ end
...
-      return spawnObjectData({
-        data = obj,
-        position = position,
-        rotation = rotation or {0, 270, 0},
-        callback_function = function(object)
+  local spawnTable = {
+    position = position,
+    callback_function = function(object)
           if name == "Hand Helper" then
             Wait.time(function() object.call("externalColorChange", color) end, 1)
```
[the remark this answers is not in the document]
Chr1Z93: Was mostly for safety, will reduce it

src/core/Global.ttslua
Chr1Z93: Error handling added:

image

Unfortunately I can't get rid of the TTS error:
image

From my understanding the `return` should stop the rest of the function to execute, but apparently it's still trying to do some of it?

PR #125, opened 2022-12-22
Deck Importer: Removing unused blocks, directly requiring configuration instead of separate object

Chr1Z93:
There has been quite some legacy code in the mod related to the deck importer:
- UI builder object (unused)
- command manager object (unused)
- configuration object (used, but a lot of unused fields in used table)

The unused objects have been removed.
Configuration has been moved to be directly included via `require` instead of grabbing it from the object.
Configuration has been greatly reduced in size to only include used fields (legacy fields removed).


Buhallin:
The configuration object is still used by Whimsical (and possibly others) to target a custom ArkhamDB instance.  This is a pretty low-impact bit of tech debt and he's done enough for the mod that I don't think we should break that functionality for him.

Chr1Z93:
> The configuration object is still used by Whimsical (and possibly others) to target a custom ArkhamDB instance.  This is a pretty low-impact bit of tech debt and he's done enough for the mod that I don't think we should break that functionality for him.

Agreed. Will update this PR to only contain the configuration clean-up while keeping the configuration as separate object.

PR #126, opened 2022-12-22
Removed random notecard

Chr1Z93:
There was a notecard behind the tentacle stand with this description: 
`lua setNotes(getObjectFromGUID('the objects guid').getJSON())`

Looks like an old mental note of someone?


Buhallin:
Possibly something CZ used as a helper while adding metadata?

Chr1Z93:
> Possibly something CZ used as a helper while adding metadata?

It's indeed a mental note by CZ, so let's just keep it there.

PR #131, opened 2022-12-29
SCED Intro Tour, final

Buhallin:
This adds the large tour starter panel to the middle of the board.


Chr1Z93 (approved):
Minor nitpick commented - not going to block it because of that

PR #141, opened 2023-01-01
added a scenario name splash on placing down scenarios and setting to toggle it

Pokachi:
Whenever a scenario is placed down, if the Scenario Card's name is populated under description, it will display the title on screen for 2 seconds with a 2 seconds fade in and 2 seconds fade out animation.

Also added a setting in the options panel to enable/disable this feature.

### DIFF
--- a/src/core/Global.ttslua
+++ b/src/core/Global.ttslua
@@ -818,7 +818,7 @@ function applyOptionPanelChange(id, state)
     -- update master clue counter
     getObjectFromGUID("4a3aa4").setVar("useClickableCounters", state)
 
-  -- option: Clickable clue counters
+  -- option: Show Title on placing scenarios
   elseif id == "showTitleSplash" then
     optionPanel[id] = state
 
--- a/src/core/Global.ttslua
+++ b/src/core/Global.ttslua
@@ -818,6 +818,10 @@ function applyOptionPanelChange(id, state)
     -- update master clue counter
     getObjectFromGUID("4a3aa4").setVar("useClickableCounters", state)
 
+  -- option: Clickable clue counters
+  elseif id == "showTitleSplash" then
+    optionPanel[id] = state
+
   -- option: Show token arranger
   elseif id == "showTokenArranger" then
     -- delete previously pulled out tokens
@@ -931,7 +935,7 @@ function onClick_defaultSettings()
   for id, _ in pairs(optionPanel) do
     local state = false
     -- override for settings that are enabled by default
-    if id == "useSnapTags" then
+    if id == "useSnapTags" or id == "showTitleSplash" then
       state = true
     end
     applyOptionPanelChange(id, state)
@@ -942,6 +946,7 @@ function onClick_defaultSettings()
     useSnapTags = true,
     showDrawButton = false,
     useClueClickers = false,
+    showTitleSplash = true,
     showTokenArranger = false,
     showCleanUpHelper = false,
     showHandHelper = {},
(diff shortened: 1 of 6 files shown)


src/core/Global.ttslua
```diff
@@ -818,6 +818,10 @@ function applyOptionPanelChange(id, state)
     -- update master clue counter
     getObjectFromGUID("4a3aa4").setVar("useClickableCounters", state)
 
+  -- option: Clickable clue counters
```
Buhallin, line 821: Nit: Update comment for the new option

src/core/PlayArea.ttslua
```diff
@@ -130,4 +130,5 @@ end
 
 function onScenarioChanged(scenarioName)
   currentScenario = scenarioName
+  Global.call('titleSplash', {scenarioName = scenarioName})
```
Buhallin, line 133: Let's call this in MythosArea as part of the direct event, rather than deferring to here

xml/Global.xml
```diff
@@ -104,4 +104,17 @@
...
+<Text id='title_splash'
+  position='0 0 0'
+  fontSize="150"
+  font="font_teutonic-arkham"
+  outline="black"
+  outlineSize="3 -3"
+  showAnimation='FadeIn'
+  hideAnimation='FadeOut'
+  active='false'
+  animationDuration='2'>
```
Buhallin, line 117: Is this big enough to handle even very long names?  Probably want to set this to wrap just in case

Pokachi: It wraps by default. I tested a long name. It even wraps by word so you don't have words that are cut in half

Buhallin: Ah, the TTS docs say the default is Overflow, I'd never tried it without setting it to wrap.  Thanks for adding it anyway, will be good for clarity.

Buhallin (approved):

src/core/Global.ttslua
```diff
@@ -952,3 +957,26 @@ function onClick_defaultSettings()
   -- update UI
   updateOptionPanelState()
 end
+
+-- splash scenario title on setup
+titleSplashId = nil
+function titleSplash(params)
+  if self.UI.getAttribute('showTitleSplash', "isOn") == "True" then
```
Chr1Z93, line 964: This should use `optionPanel[showTitleSplash]` as condition

src/core/Global.ttslua
```diff
@@ -952,3 +957,26 @@ function onClick_defaultSettings()
   -- update UI
   updateOptionPanelState()
 end
+
+-- splash scenario title on setup
+titleSplashId = nil
+function titleSplash(params)
```
Chr1Z93, line 963: Will there be any parameters besides `currentScenario`? If not, I'd change the argument to not be a table.

src/core/Global.ttslua
```diff
@@ -952,3 +957,26 @@ function onClick_defaultSettings()
   -- update UI
   updateOptionPanelState()
 end
+
+-- splash scenario title on setup
+titleSplashId = nil
```
Chr1Z93, line 962: Would add `local` to the variable declaration and move it to the top of the script.
How about including `Loop` in the name so it's not going to be confused with the XML id?

src/core/Global.ttslua
```diff
@@ -952,3 +957,26 @@ function onClick_defaultSettings()
...
+    -- display scenario name and set a 4 seconds (2 seconds animation and 2 seconds on screen)
+    -- wait timer to hide the scenario name
+    UI.setValue('title_splash', params.scenarioName)
+    UI.show('title_splash')
+    titleSplashId = Wait.time(function()
+      UI.hide('title_splash')
+      titleSplashId = nil
+    end, 4)
+  end
+end
```
Chr1Z93, line 982: Empty line at the end will get added by decomposing, so should not be removed

src/core/MythosArea.ttslua
```diff
@@ -55,6 +55,7 @@ function resetTokensIfInDeckZone(container, object)
 end
 
 function fireScenarioChangedEvent()
+  Global.call('titleSplash', {scenarioName = currentScenario})
```
Chr1Z93, line 58: See above, would pass as string instead of table

xml/Global.xml
```diff
@@ -104,4 +104,18 @@
   </Panel>
 </VerticalLayout>
 
+<!-- Title Splash when starting a scenario -->
+<Text id="title_splash"
+  position="0 0 0"
```
Chr1Z93, line 109: I don't think this is a valid attribute (and should be omitted if this is the default value anyway)

Chr1Z93 (changes requested):

xml/Global.xml
Pokachi: this is valid, but it's also the default https://api.tabletopsimulator.com/ui/attributes/#positionsize-attributes-advanced

Chr1Z93 (approved):

PR #143, opened 2023-01-01
Playmat Image Swapper: Fix variable name and declaration

Chr1Z93:
- fixes the variable name and declaration
- reloads the location data
- adds a cool down of 0.1s to the collision event


src/core/PlayArea.ttslua
```diff
@@ -54,26 +56,31 @@ end
...
   end
 end
 
 function onCollisionEnter(collision_info)
-  if not COLLISION_ENABLED then return end
+  local obj = collision_info.collision_object
+
+  -- only trigger the following once every 0.1s for each object
+  -- (otherwise TTS fires this event 20 times and that causes 20 JSON decodes)
+  if not COLLISION_ENABLED or lastCollisionObject == obj then return end
```
Buhallin, line 70: The multiple call only seems to happen for tokens.  Let's limit this to cards only (which is all we really care about) rather than the time gate

Buhallin (changes requested):

src/core/PlayArea.ttslua
Chr1Z93: Changed

Buhallin (approved):

PR #146, opened 2023-01-01
Playmat - Upkeep: Dont unexhaust investigators

Chr1Z93:
This stops investigators with missing metadata to get rotated on upkeep

### DIFF
--- a/src/playermat/Playmat.ttslua
+++ b/src/playermat/Playmat.ttslua
@@ -237,9 +237,9 @@ function doUpkeep(_, color, alt_click)
     local obj = v.hit_object
     if obj.getDescription() == "Action Token" and obj.is_face_down then
       obj.flip()
-    elseif obj.tag == "Card" and not obj.is_face_down then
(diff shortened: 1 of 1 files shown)


src/playermat/Playmat.ttslua
```diff
@@ -536,7 +536,7 @@ end
 
 function maybeUpdateActiveInvestigator(card)
   local notes = JSON.decode(card.getGMNotes())
-  if notes ~= nil and notes.type == "Investigator" and notes.id ~= activeInvestigatorId then
+  if notes ~= nil and notes.type == "Investigator" and notes.id ~= nil and notes.id ~= activeInvestigatorId then
```
Buhallin, line 539: If you drop an investigator with data then an investigator that has data but no ID, this will leave the old stats, right?

Chr1Z93: Yep, exactly

Chr1Z93: Added handling to load default values for investigators without metadata

Buhallin (approved):

PR #147, opened 2023-01-01
Token Manager: Bugfix for spawning of multiple tokens and default rotation of damage and horror tokens

Chr1Z93:
- fixes wrong name of variable in `TokenManager.spawnCounterToken`
- implements usage of the `shiftDown` paramater counter tokens
- removes unused code from that function
- default rotation (X and Z) of damage and horror tokens fixed (so that they come out flat)


src/core/token/TokenManager.ttslua
```diff
@@ -161,52 +161,39 @@ do
...
-        offsets[i] = baseOffset
-        offsets[i][3] = offsets[i][3] + shiftDown
-      end
-    end
-    local pos = card.positionToWorld(offsets[1])
-    pos.y = card.getPosition().y + 0.15
-    TokenManager.spawnToken(pos, tokenType, card.getRotation(), function(spawned)
+    if tokenValue < 1 or tokenValue > 50 then return end
+
+    local pos = card.positionToWorld(card.positionToLocal(card.getPosition()) + Vector(0, 0.2, shiftDown))
```
Buhallin, line 168: Can remove the interior card.positionToLocal(card.getPosition()) here - that's just going to give the origin of the card.  positionToWorld(Vector(0, 0.2, shiftDown)) is sufficient

Buhallin, line 168: Why make this different than the single-item spawn list?

Chr1Z93: Different in what sense?

Buhallin: Why the explicit position rather than referencing the position table for 1 token?

Buhallin (approved):

src/core/token/TokenManager.ttslua
Chr1Z93: Hm that didn't have any deep reasoning, it just felt more natural because there is no variation. Could change this if you want to futureproof this for changing of the offsets

Chr1Z93: updated

PR #148, opened 2023-01-02
Option: Clickable resource tokens

Chr1Z93:
This adds a new option to the option panel: Clickable resource tokens.

Preview:
image


src/core/token/TokenManager.ttslua
```diff
@@ -149,8 +149,12 @@ do
   --     spawned state object rather than spawning multiple tokens
   ---@param shiftDown An offset for the z-value of this group of tokens
   TokenManager.spawnTokenGroup = function(card, tokenType, tokenCount, shiftDown)
+    local optionPanel = Global.getTable("optionPanel")
```
Buhallin, line 152: Comment for future, but we really need to convert this to an API and have a method to get each option value.  Other objects needing to know the keys that are being used (and hardcoding them) is going to be fragile.

src/playermat/Playmat.ttslua
```diff
@@ -415,6 +415,10 @@ function replenishTokens(card, count, replenish)
     if obj.getCustomObject().image == "http://cloud-3.steamusercontent.com/ugc/1758068501357192910/11DDDC7EF621320962FDCF3AE3211D5EDC3D1573/" then
       foundTokens = foundTokens + math.abs(obj.getQuantity())
       obj.destruct()
+    elseif obj.getName() == "Resource Counter" then
+      foundTokens = obj.getVar("value")
+      obj.destruct()
```
Buhallin, line 420: Why destroy the object here?  Should be able to just set the value on the current token, rather than destroying and creating a new one (which then has the value set on it)

Chr1Z93: This was mostly done to minimize the code changes and have the same behaviour for "normal" and clickable resource tokens - replenish removes old tokens and then respawns the correct amount

src/core/token/TokenManager.ttslua
Chr1Z93: True, that would make options that affect other objects / script easier to implement too

objects/TokenSource.124381/ResourceCounter.498ec0.ttslua
```diff
@@ -0,0 +1,43 @@
+value = 0
```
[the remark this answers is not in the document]
Chr1Z93: Good point, gonna rework this

src/playermat/Playmat.ttslua
Chr1Z93: Updated

Buhallin (approved):

PR #151, opened 2023-01-02
Option: Location snap tags for play area

Chr1Z93:
This PR includes the following changes:
- moving snap points from the table to the play area
- function to rebuild these snap points (used by me to create them and could be useful in the future)
- new function for the playarea to toggle snap tags for its snap points
- new option in the option panel to enable snap tags for locations
- added the tag "Location" to the connection markers so that they always snap

Snap point layout was kept (if uneven row + uneven column then rotation snap), but I added points "inbetween" rows:
image


src/core/Global.ttslua
```diff
@@ -819,6 +820,11 @@ function applyOptionPanelChange(id, state)
     -- update master clue counter
     getObjectFromGUID("4a3aa4").setVar("useClickableCounters", state)
 
+  -- option: Draw 1 button
```
Buhallin, line 823: Update comment

src/core/Global.ttslua
```diff
@@ -945,6 +951,7 @@ function onClick_defaultSettings()
   -- clean reset of variable
   optionPanel = {
     useSnapTags = true,
+    PlayAreaSnapTags = false,
```
Buhallin, line 954: This should default true, right?

xml/OptionPanel.xml
```diff
@@ -142,6 +142,20 @@
...
+          <!-- Option: play area snap tags -->
+          <Row class="option-text">
+            <Cell class="option-text">
+              <VerticalLayout class="text-column">
+                <Text class="option-header">Enable snap tags for play area</Text>
+                <Text class="description">Only cards with the tag "Location" will snap (official cards are supported by default). Disable this if you are having issues with custom content.</Text>
+              </VerticalLayout>
+            </Cell>
+            <Cell class="option-button">
+              <Toggle id="PlayAreaSnapTags"
```
Buhallin, line 154: Lower snake case for consistency with other IDs

src/core/Global.ttslua
Chr1Z93: Was hesitant to do that because custom campaigns are way more prominent than custom player cards - so it would happen quite often that people would need to disable this

src/core/Global.ttslua
Chr1Z93: good catch

src/core/Global.ttslua
Buhallin: You're checking in the snap points with the location set though, which means the default is out of sync with what the mod loads with.

Generally we've been optimizing the default state of the mod for official content, with options to change that for custom.  I'm not completely against this being different, but I do think we should be consistent where possible.

Chr1Z93: Alright, changed it

Buhallin (approved):

PR #152, opened 2023-01-02
Token Remover: Exclude chaos tokens + code cleanup

Chr1Z93:
The token remover now excludes chaos tokens from it's deletion process. Previously this could cause some unwanted interactions if the remover was placed near playermats.

Also removes the name from it and makes the scripting zone less annoying when activating the gizmo tool by reducing it's height.


src/util/TokenRemover.ttslua
```diff
@@ -1,72 +1,80 @@
...
+  end
 
----@param entering TTSObject
----@param object TTSObject
-function onObjectEnterScriptingZone(entering , object)
-    if zone~=entering then return end
-    if object==self then return end
-    if object.type=="Deck" or object.type=="Card" then return end
+  local name = object.getName()
+  for _, entry in ipairs(CHAOS_TOKEN_NAMES) do
```
Buhallin, line 71: Can make the token name list a map using the names as keys, will remove the loop here and give an O(1) check time

Buhallin (changes requested):

src/util/TokenRemover.ttslua
Chr1Z93: Updated

Buhallin (approved):

PR #158, opened 2023-01-05
New utility file: TokenChecker

Chr1Z93:
New file added to `src\core\token`: TokenChecker
Just contains one function for now: `isChaosToken(obj)`
- returns true if the object is a chaos token and false if not
- implemented that for the token remover and playermat discard button (to return chaos tokens on cards like unrelenting)

Additionally:
- bless / curse token bag: added missing name to tokesn, changed contained object to be 10x the same (no need for multiple files)


src/util/isChaosToken.ttslua
```diff
@@ -0,0 +1,31 @@
+do
```
Buhallin, line 1: I don't think we're going to want a pattern with one function per file.  That's going to explode fast.  Should probably make this something like ChaosTokenUtils and attach the function to that

objects/AllPlayerCards.15bb07/Unrelenting1.45386d.ttslua
```diff
@@ -1,5 +1,3 @@
-ASSEMBLY_POSITION = {69.08, 4, 36.45}
```
Buhallin, line 1: Should there be more changes in this file?

Buhallin (changes requested):

src/util/isChaosToken.ttslua
Buhallin: Actually let's just make this TokenUtils or TokenChecker - we can reasonably expand it to isClue(), isDoom(), etc

objects/AllPlayerCards.15bb07/Unrelenting1.45386d.ttslua
Chr1Z93: Nope, just a removal of a dead table (noticed this while looking at all the cards that can have chaos tokens on them)

src/util/isChaosToken.ttslua
Chr1Z93: Done

Buhallin (approved):

PR #179, opened 2023-01-10
Playarea Image Swapper: Bugfix for custom data helpers

Chr1Z93:
changes the scope of the variable so that it can be accessed from outside


Chr1Z93:
I confused the play area with the playermats... Closing this

PR #183, opened 2023-01-11
Enhancements and tweaks for the scenario title splash

Buhallin:
- Add a background gradient
- Add a wrapping panel to hold both text and gradient
- Add font resizing in case of very long scenario names.  None currently overflow, but this should ensure they stay on one line if we get a longer one

### DIFF
--- a/src/core/Global.ttslua
+++ b/src/core/Global.ttslua
@@ -875,7 +875,7 @@ function applyOptionPanelChange(id, state)
   -- option: Show CYOA campaign guides
   elseif id == "showCYOA" then
     optionPanel[id] = spawnOrRemoveHelper(state, "CYOA Campaign Guides", {65, 1.6, -11})
-    
+
   -- option: Show custom playmat images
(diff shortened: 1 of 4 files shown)


src/core/Global.ttslua
```diff
@@ -1014,14 +1014,17 @@ function titleSplash(scenarioName)
...
 
     -- display scenario name and set a 4 seconds (2 seconds animation and 2 seconds on screen)
     -- wait timer to hide the scenario name
-    UI.setValue('title_splash', scenarioName)
+    UI.setValue('title_splash_text', scenarioName)
     UI.show('title_splash')
+    -- UI.show('title_gradient')
     hideTitleSplashWaitFunctionId = Wait.time(function()
       UI.hide('title_splash')
+      -- UI.hide('title_gradient')
```
Chr1Z93, line 1027: What's the reasoning for the commented out lines with `title_gradient`? Should they be here?

Chr1Z93 (approved):

PR #184, opened 2023-01-11
Optimize location connection drawing during drag operations

Buhallin:
Previously the entire location map was being rebuilt and redrawn in onUpdate() during a drag operation, including the connectors which weren't changing.  This was causing notable lag on some systems.

This splits the draw operations into two separate pieces - locations on the board which are unchanging, and locations which are being dragged.  This allows only connections which are actually moving to be recalculated.

Since TTS only allows a single list of vector lines per object the dragged lines are now set on the cards themselves rather than the PlayArea.  This may possibly improve performance in itself, by keeping each list of vectors smaller.  Supporting this required some changes in how pre-built connections are handled.

### DIFF
--- a/src/core/PlayArea.ttslua
+++ b/src/core/PlayArea.ttslua
@@ -103,7 +103,12 @@ function onCollisionEnter(collisionInfo)
   if shouldSpawnTokens(card) then
     tokenManager.spawnForCard(card)
   end
-  draggingGuids[card.getGUID()] = nil
+  -- If this card was being dragged, clear the dragging connections.  A multi-drag/drop may send
+  -- the dropped card immediately into a deck, so this has to be done here
+  if draggingGuids[card.getGUID()] ~= nil then
+    card.setVectorLines(nil)
+    draggingGuids[card.getGUID()] = nil
+  end
   maybeTrackLocation(card)
 end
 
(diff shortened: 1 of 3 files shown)


src/core/PlayArea.ttslua
```diff
@@ -3,14 +3,17 @@
 ---------------------------------------------------------
 
 -- set true to enable debug logging
-local DEBUG = false
+local DEBUG = true
```
Chr1Z93, line 6: Set to false

src/core/PlayArea.ttslua
```diff
@@ -3,14 +3,17 @@
...
 
 -- Location connection directional options
 local BIDIRECTIONAL = 0
 local ONE_WAY = 1
+local INCOMING_ONE_WAY = 2
 
 -- Connector draw parameters
 local CONNECTION_THICKNESS = 0.015
+local DRAGGING_CONNECTION_THICKNESS = 0.15
```
Chr1Z93, line 15: Why do these have 10x the thickness?

src/core/PlayArea.ttslua
Buhallin: I am far too used to runtime debug levels.  Takes me 10 minutes to figure out why my logs aren't showing anything, and I forget to turn it off every time

src/core/PlayArea.ttslua
Buhallin: Vector lines are in the context of the object they're set on, including scale.  Playmat is 10X so has to be 1/10th the width.

Chr1Z93: Ah that makes sense

Chr1Z93 (approved):

PR #189, opened 2023-01-14
Standalone Scenario Box: increasing size by 30%

Chr1Z93:
Size increase because these boxes are pretty small, which makes the "Recall" and "Place" button unneccessarily hard to click.
Position udpated for new table layout.

image


Buhallin:
I think it would be good to increase the spacing here a bit too, doesn't seem any reason for them to be that close together

Buhallin (approved):

PR #191, opened 2023-01-15
Playmat: Increasing size of discard buttons

Chr1Z93:
Increases the size of the discard buttons to make them easier to click.

Before:
image

New:
image


BootleggerFinn (approved):
Looks good!

PR #199, opened 2023-01-24
TSK: Adding story assets to playercard bag

Chr1Z93:
- added the 6 story assets to the playercards bag
- bunch of changed files by automatic changes because of decomposing


Buhallin:
Need to add the cycle tag for these since they're player cards

Chr1Z93:
> Need to add the cycle tag for these since they're player cards

Updated

Buhallin (approved):

PR #207, opened 2023-02-02
"Hidden Zone" tool cleanup

Chr1Z93:
- moves the two unused objects "UI Builder" and "Command Manager" into the legacy asset container (the hat in the bottom right corner)
- updates for conflicting code of tools with scriptingbutton triggers (button 3 is used for multiple)
- added feedback to deckcutter when conditions are not met
- renamed "Token Spawner" to "Token Spawn Tool" (because it was confusable with the "Token Manager" and "Token Spawn Tracker"


Buhallin:
UI Manager and Command Manager can just be deleted.  Since we've got history thanks to the repo, I don't think the archive is necessary any more

Chr1Z93:
> UI Manager and Command Manager can just be deleted. Since we've got history thanks to the repo, I don't think the archive is necessary any more

Good point, updated it. Also removed the old version of the Token Arranger and Attachment Helper

Tikatoy (approved):

PR #208, opened 2023-02-02
Barkham Horror: Changing into download box

Chr1Z93:
Changes the Barkham Horror box into a download box instead of having the scenario preloaded


Tikatoy (commented):
Tested loading from UI and dragging from cat object. Both work!

Should this also be added to the scenarios placed by Official Scenarios -> The Side Missions? It probably wasn't in there originally so as to not duplicate it, but now that it's on load on demand, seems like it could now be added to Side Missions.

Chr1Z93:
> Tested loading from UI and dragging from cat object. Both work!
> 
> Should this also be added to the scenarios placed by Official Scenarios -> The Side Missions? It probably wasn't in there originally so as to not duplicate it, but now that it's on load on demand, seems like it could now be added to Side Missions.

Good idea. Updated the PR
image

Tikatoy (approved):

PR #210, opened 2023-02-03
Family Inheritance: Context menu functions

Chr1Z93:
This adds 3 context menu functions to "Family Inheritance":
- add 4: adds four tokens to it, if the new count is above 12, a clickable counter will be spawned
- take all: moves the resources to the resource pool of the player's mat
- discard all: removes all resources from it

Additionally, there now is "gainResource" function for the playermats which can be accessed through their API.

image


Tikatoy (approved):
Nice!

PR #211, opened 2023-02-03
Victory Display: Visual update and automation

Chr1Z93:
- updated image
- bigger size (10 cards)
- automated VP counting (handles decks of cards too)
- includes a performance optimization for location connection drawing

image


Tikatoy (commented):
Works well!

Regarding the highlight cards without metadata toggle, it's not clear which state it's in; perhaps just change the tooltip so that when it's disabled it says "enable to..." and conversely when it's enabled show "disable to..."?

Minor issue, I'm good with approving if you feel it's unnecessary.

Chr1Z93:
> Works well!
> 
> Regarding the highlight cards without metadata toggle, it's not clear which state it's in; perhaps just change the tooltip so that when it's disabled it says "enable to..." and conversely when it's enabled show "disable to..."?
> 
> Minor issue, I'm good with approving if you feel it's unnecessary.

Good idea! Updated it

Tikatoy (approved):

PR #213, opened 2023-02-06
Moving player card scripts

Chr1Z93:
This moves the scripts on player cards into their own folder: `src/playercards/cards`.
(to make them easier accessible)


Tikatoy (approved):
Spot checked a customizable card and a card that seals tokens. Working as expected.

PR #215, opened 2023-02-11
Added broadcastToAll in DrawChaostoken

fastjoe23:
Hi,
we used the  mod a lot for gaming in the last months. And so i added a small tweak in our personal save games. But maybe you like this tweak to have in the official mod. 
I added a broadcastToAll Message in DrawChaosToken, when a ChaosToken is revealed. So you do not have to look at the playermat of your teammate, but you see a broadcastToAll message at the screen.
Of course if you don't like the change, free feel to dismiss it. 
Greetings
fastjoe23

### DIFF
--- a/src/core/Global.ttslua
+++ b/src/core/Global.ttslua
@@ -311,6 +311,7 @@ function drawChaostoken(params)
   local mat = params[1]
   local tokenOffset = params[2]
   local isRightClick = params[3]
+  local playerColor = MAT_GUID_TO_COLOUR[mat.getGUID()]
   chaosbag = findChaosBag()
 
   -- return token(s) on other playmat first
@@ -334,6 +335,7 @@ function drawChaostoken(params)
       rotation = mat.getRotation(),
 	    callback_function = function(obj) trackChaosToken(obj, mat.getGUID()) end
     })
+    broadcastToAll(Player[playerColor].steam_name ..  " draws " .. token.getName(), Player[playerColor].color)
     chaosTokens[#chaosTokens + 1] = token
     return
   else


Buhallin:
I don't really care for this as a general change.  It seems very spammy, especially as a broadcast, and I can see it being far more annoying than useful.  I'll let the other owners chime in on it though.

Chr1Z93:
I also think this would be a bit too much for most players. Could maybe do something similar as option, but I'd prefer having something less noticeable than a broadcast.

fastjoe23:
I admit there a definitely more messages than before. Especially when using an investigator like jacqueline. But i added these lines of code in our save games and none of my group ever complained about it. Even when Jacqui was in play. 

Another possibility would be to make this optional. If you can give me any hint, how to integrate this in any options/settings part of the mod, i would try to implememt it optionally. So people can switch it on and off.

Buhallin:
We look for something more than "Nobody in my group complains about it" :)  We haven't had many (any?) requests for something like this, it's not hard to see what someone had drawn if you're interested, and it would increase the number of broadcasts during a session by something like 500x.  I may be an excessive fan of minimalist design, but it just feels like a minimal benefit for those who want it and a very large drawback for those who would find it annoying.

Even as an option, I think this is a net negative.  Our options panel is already larger than I personally like, and I don't think this meets the usefulness bar for cluttering it further.

fastjoe23:
Ok, then just deny the request.

Already much more lines of comment then of code. :-)



Am 12.02.2023 um 11:29 schrieb Buhallin ***@***.***>:

﻿

We look for something more than "Nobody in my group complains about it" :) We haven't had many (any?) requests for something like this, it's not hard to see what someone had drawn if you're interested, and it would increase the number of broadcasts during a session by something like 500x. I may be an excessive fan of minimalist design, but it just feels like a minimal benefit for those who want it and a very large drawback for those who would find it annoying.

Even as an option, I think this is a net negative. Our options panel is already larger than I personally like, and I don't think this meets the usefulness bar for cluttering it further.

Buhallin:
Software is always more discussion than code :)  I did want to give the other reviewers time to chime in on this, and it sounds like we're in agreement.

Please feel free to make other submissions, and if there's something that's going to be more effort you can always drop a question in the contributors channel of the Discord for early feedback and thoughts.

PR #216, opened 2023-02-24
Search Assistant: Adding missing tag for movement exclusion via displacement tool

Chr1Z93:


Tikatoy (approved):
Tested search assistant with:
- clean up helper
- displacement tool
Works as expected!

PR #217, opened 2023-02-26
Attachment Helper: support for decks of cards, added backpack image and misc. updates

Chr1Z93:
- code restructured to allow easier addition of new images
- added support for different font colors (based on image)
- included function to automatically split decks of cards that enter it
- increased font size and moved it upwards to allow for more buttons
- added new image / option for backpack
- removed orphaned lua file in `object` folder

Preview of new backpack image:
image


Tikatoy (approved):
Tested dropping deck and loaded all of the images. Looks good!

PR #220, opened 2023-02-28
Hand Helper: code updates

Chr1Z93:
- added automatic color detection (including custom hand colors that don't match the playermat color!)
- added automatic DES detection
- new utility functions added to playmatApi
- updated random discard function
- updated description


Tikatoy (approved):
Tested and looks good!

PR #221, opened 2023-02-28
Underworld Market: stop including permanent cards

Chr1Z93:
Bugfix to stop permanent cards from being included in the market deck (e.g. Underworld Support)


src/arkhamdb/DeckImporterMain.ttslua
```diff
@@ -269,7 +269,8 @@ function handleUnderworldMarket(cardList, playerColor)
       card.zone = "SetAside3"
     elseif (card.metadata.traits ~= nil and string.find(card.metadata.traits, "Illicit", 1, true)
         and card.metadata.bonded_to == nil
```
Buhallin, line 271: Might want to fix this while you're at it.  We don't populate bonded_to any more.  I think this and the permanent check can be combined into a zone check to only pull cards from the Deck.

Buhallin (changes requested):

Buhallin (approved):

PR #222, opened 2023-03-01
Adding MythosAreaAPI

Chr1Z93:
This cleanes the latest integration of the Token Arranger a bit up by introducing a separate API for the Mythos Area.
--> contains no functional changes


Tikatoy (approved):
Tested, looks good!

PR #223, opened 2023-03-02
Moving spawnable tools into a separate (hidden) bag

Chr1Z93:
- this moves the tools that are spawnable via option panel into a new bag (hidden behind the tentacle stand)
- the option panel section for accessories has been alphabetically sorted
- added a description to fan-made accessories barrell


Tikatoy (approved):
Spot checked Token Arranger and Chaos Bag Manager, looks good!

PR #224, opened 2023-03-02
Deckimporter: Updated handling for parallel, revised and promo investigators

Chr1Z93:
Since the revised and promo art for investigators are no longer added as separate cards (but states instead) this updates the deckimporter to automatically load the correct state for these cards.

### DIFF
--- a/src/arkhamdb/DeckImporterMain.ttslua
+++ b/src/arkhamdb/DeckImporterMain.ttslua
@@ -99,7 +99,8 @@ end
 --     from a parent bonded card.
 ---@param customizations String ArkhamDB data for customizations on customizable cards
 ---@param playerColor String Color name of the player mat to place this deck on (e.g. "Red")
-function loadCards(slots, investigatorId, bondedList, customizations, playerColor)
+---@param loadAltInvestigator String Contains the name of alternative art for the investigator ("normal", "revised" or "promo")
+function loadCards(slots, investigatorId, bondedList, customizations, playerColor, loadAltInvestigator)
   function coinside()
     local allCardsBag = getObjectFromGUID(ALL_CARDS_GUID)
     local yPos = {}
@@ -143,6 +144,8 @@ function loadCards(slots, investigatorId, bondedList, customizations, playerColo
         callback = function(deck) deck.spread(spreadDistance) end
       elseif zone == "Deck" then
         callback = function(deck) deckSpawned(deck, playerColor) end
+      elseif zone == "Investigator" or zone == "Minicard" then
+        callback = function(card) loadAltArt(card, loadAltInvestigator) end
       end
       Spawner.spawnCards(
           zoneCards,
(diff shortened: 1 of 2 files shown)


Tikatoy:
Are the revised core art investigators  in the mod?

Or was this for promo investigator art? I tested with an old Norman Withers deck that has the promo art, and it did load the Norman Withers investigator card correctly, and I was able to change state.

Chr1Z93:
> Are the revised core art investigators in the mod?
> 
> Or was this for promo investigator art? I tested with an old Norman Withers deck that has the promo art, and it did load the Norman Withers investigator card correctly, and I was able to change state.

Yeah, the alternative arts (revised and promo) are added as states to the normal cards. When I did this for the last update, I didn't account for the deck importer. The bug at hand is that it still tries to load the revised investigator cards by ID. This PR replaces that ID changing for revised cards with a message that those can be found in the state. (If this didn't make sense, try loading a revised Daisy deck in 3.0.0 and then in the version from this mod.)

Tikatoy:
I created a Daisy deck selecting Revised Core for  front and back: https://arkhamdb.com/deck/view/2800646

In both 3.0.0 and the save I generated from the branch, it loaded the (stateless) parallel front/back Daisy. Do I need to be doing something different?

Chr1Z93:
> I created a Daisy deck selecting Revised Core for front and back: https://arkhamdb.com/deck/view/2800646
> 
> In both 3.0.0 and the save I generated from the branch, it loaded the (stateless) parallel front/back Daisy. Do I need to be doing something different?

Yep, that's the issue. It should not be loading the parallel ones if you select revised front/back.

Chr1Z93:
> I created a Daisy deck selecting Revised Core for front and back: https://arkhamdb.com/deck/view/2800646
> 
> In both 3.0.0 and the save I generated from the branch, it loaded the (stateless) parallel front/back Daisy. Do I need to be doing something different?

Uhh that's a good catch. I only tested revised front and didn't add revised back. Will need to update this PR

Chr1Z93:
Pr is updated and should now handle alt art correctly. I didn't have an old decklist with promo art available for testing, but assume that that will be analog to the revised art.

src/arkhamdb/DeckImporterMain.ttslua
```diff
@@ -407,6 +410,51 @@ function handleCustomizableUpgrades(cardList, customizations)
...
+---@param card Object Card which needs to be set the state for
+---@param loadAltInvestigator String Contains the name of alternative art for the investigator ("normal", "revised" or "promo")
+function loadAltArt(card, loadAltInvestigator)
+  if loadAltInvestigator == "normal" then return end
+
+  -- lookup correct stateId for investigator and alt art state
+  local baseId = string.gsub(JSON.decode(card.getGMNotes()).id, "-m", "")
+
+  local stateIdTable = {}
+  -- Roland Banks
```
Buhallin, line 423: I don't love the hardcoding here.

Could do this by adding something to the states - maybe Memo, maybe an extra metadata field - with the "revised" or "promo" tag, then inspect the card data to find the correct one.  Would be more durable for future changes, but it's probably not worth it given the limited number of alt arts,

Buhallin (approved):

PR #227, opened 2023-03-06
Summoned Servitor Dominance upgrade sheet buttons

Entrox-Licher:
Added buttons to indicate selection for the Summoned Servitor Dominance upgrade (on the corresponding upgrade sheet, not the SS card itself).  The buttons are based off of the ones for Living Ink (only rectangles instead of circles).  

**Be advised**, the ArkhamDB Importer code has not been updated yet, so another commit will be needed to tie up that loose end (working on it myself, but feel free to beat me to the punch)


src/playercards/customizable/SummonedServitorUpgradeSheet.ttslua
```diff
@@ -178,4 +206,91 @@ function createButtonsAndBoxes()
...
+    }
+    self.createButton(buttonData)
+    buttonData.click_function = "click_ally"
+    buttonPositions.x = -1 * SLOT_ICON_POSITIONS.ally.x
+    buttonData.position = buttonPositions
+    self.createButton(buttonData)
+  end
+
+function click_arcane()
+    selectedSlots.arcane = not selectedSlots.arcane
```
Buhallin, line 233: Since these are mutually exclusive, turning on Arcane should turn off Ally

src/playercards/customizable/SummonedServitorUpgradeSheet.ttslua
```diff
@@ -178,4 +206,91 @@ function createButtonsAndBoxes()
...
+        local circle = getCircleVector(SLOT_ICON_POSITIONS[slot], slot)
+        if circle ~= nil then
+          table.insert(circles, circle)
+        end
+      end
+    end
+    self.setVectorLines(circles)
+  end
+  
+  function getCircleVector(center, slot)
```
Buhallin, line 257: Clean up old circle code, and rename the method (getBoxVector?)

src/playercards/customizable/SummonedServitorUpgradeSheet.ttslua
```diff
@@ -31,27 +31,54 @@ existingBoxes = { 1, 1, 1, 1, 1, 2, 3, 5 }
 
 inputBoxes = {}
 
+-- Locations of the skill selectors
+local SLOT_ICON_POSITIONS = {
+    arcane = { x = 0.160, z = 0.65},
+    ally = { x = -0.073, z = 0.65}
+  }
+  
+  local selectedSlots = {
```
Buhallin, line 40: Since only one of these can be selected at a time (unlike Living Ink) I think a single string value would be cleaner

Buhallin (changes requested):

src/playercards/customizable/SummonedServitorUpgradeSheet.ttslua
```diff
@@ -31,27 +31,54 @@ existingBoxes = { 1, 1, 1, 1, 1, 2, 3, 5 }
...
+  end
 
 -- Startup procedure
 function onLoad(saved_data)
     if saved_data ~= "" and markDEBUG == "" then
         local loaded_data = JSON.decode(saved_data)
         markedBoxes = loaded_data[1]
         inputValues = loaded_data[2]
+        log(loaded_data[3])
+        for _, slot in ipairs(loaded_data[3]) do
```
Buhallin, line 68: This throws an error when loaded with an old save state.  Add a nil check to make it more resilient

Buhallin (changes requested):

src/playercards/customizable/SummonedServitorUpgradeSheet.ttslua
Chr1Z93: Additionally, this PR should update the script state accordingly

src/playercards/customizable/SummonedServitorUpgradeSheet.ttslua
```diff
@@ -178,4 +193,79 @@ function createButtonsAndBoxes()
...
+    }
+    self.createButton(buttonData)
+    buttonData.click_function = "click_ally"
+    buttonPositions.x = -1 * SLOT_ICON_POSITIONS.ally.x
+    buttonData.position = buttonPositions
+    self.createButton(buttonData)
+  end
+
+function click_arcane()
+    if selectedSlot ~= "arcane" then
```
Buhallin, line 220: Reverse this to avoid the negative logic

src/playercards/customizable/SummonedServitorUpgradeSheet.ttslua
Buhallin: Not sure what might be going on there, but as long as the load handles it if it's missing it shouldn't need to be updated.  Would be nice, but not critical.

Buhallin:
Quick thought on this - should we put a transparent gray box around the text to indicate they're clickable?  Can easily see people not realizing they can select it

src/playercards/customizable/SummonedServitorUpgradeSheet.ttslua
Chr1Z93: > For some reason whenever I try to stage the changes to the script state, it doesn't resolve? I get a warning about line endings and then git tells me there were actually no changes to the script state to begin with (i.e. nothing to stage anymore).

Just to be sure, are you changing this line? https://github.com/argonui/SCED/blob/main/objects/AllPlayerCards.15bb07/LivingLinkUpgradeSheet.19a05b.json#L37

Chr1Z93:
> Quick thought on this - should we put a transparent gray box around the text to indicate they're clickable? Can easily see people not realizing they can select it

Good idea!

Buhallin:
If you do add the box, I'd recommend doing it with vector lines rather than making the buttons visible - you can't control the shape of them, and they get very weird when they're small...  A single thick line drawn across should give the same effect

src/playercards/customizable/SummonedServitorUpgradeSheet.ttslua
Entrox-Licher: I got it now!  I assumed the mod manager would be updating these automatically, that was my mistake.  Went in and manually updated, should work correctly now.

Entrox-Licher:
> If you do add the box, I'd recommend doing it with vector lines rather than making the buttons visible - you can't control the shape of them, and they get very weird when they're small... A single thick line drawn across should give the same effect

Done!

src/playercards/customizable/SummonedServitorUpgradeSheet.ttslua
Entrox-Licher: To be clear, you prefer it like this?

```
    if selectedSlot == "arcane" then
        selectedSlot = ""
    else
        selectedSlot = "arcane"
    end
    updateSlotDisplay()
```

Buhallin: Yep, that looks better.  It's best to avoid negative logic unless it's absolutely necessary for clarity

https://csharpdeveloper.wordpress.com/2016/01/20/roberts-rules-of-coders-8-avoid-negative-conditionals-and-method-names/

src/playercards/customizable/SummonedServitorUpgradeSheet.ttslua
```diff
@@ -31,27 +31,41 @@ existingBoxes = { 1, 1, 1, 1, 1, 2, 3, 5 }
...
 function onLoad(saved_data)
     if saved_data ~= "" and markDEBUG == "" then
         local loaded_data = JSON.decode(saved_data)
         markedBoxes = loaded_data[1]
         inputValues = loaded_data[2]
+        if #loaded_data > 2 then
+            selectedSlot = loaded_data[3]
+        else
+            selectedSlot = ""
+        end
```
Chr1Z93, line 60: Since you now updated the save state, you could get rid of LOC 56 and LOC 58-60

Chr1Z93 (approved):
Looks good to me! Commented a minor change, but not going to block based on that.

src/playercards/customizable/SummonedServitorUpgradeSheet.ttslua
Entrox-Licher: Perfect, addressed!

src/playercards/customizable/SummonedServitorUpgradeSheet.ttslua
Entrox-Licher: Fixed this incidentally while working on the importer code!  There is some code that might be worth cleaning up in some of the other sheets like Living Ink that have dead variables like `inputValues`, but it's more a clarity thing than functional

src/playercards/customizable/SummonedServitorUpgradeSheet.ttslua
```diff
@@ -35,32 +35,30 @@ inputBoxes = {}
...
+local selectedSlot = slotIndices.none
 
 -- override 'marked boxes' for debugging ('all' or 'none')
 markDEBUG = ""
 
 -- save state when going into bags / decks
 function onDestroy() self.script_state = onSave() end
 
-function onSave() return JSON.encode({ markedBoxes, inputValues, selectedSlot }) end
+function onSave() return JSON.encode({ markedBoxes, {selectedSlot} }) end
```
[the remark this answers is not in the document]
Entrox-Licher: This shouldn't actually affect the checkboxes at all.  It was actually a mistake for me to make the save state to be 3 items long at all, since the way it works is the first item stores checkbox data, and the second item stores all non-checkbox data.  This is the inputValues variable for all of the textbox customizables (eg. Grizzled), but is not used in the other non-checkbox, non-textbox input card, Living Ink (which is largely what I modelled this one against).

Any references to inputValues are vestigial references to textboxes that are simply not present in Living Ink or Summoned Servitor

src/playercards/customizable/SummonedServitorUpgradeSheet.ttslua
```diff
@@ -35,32 +35,30 @@ inputBoxes = {}
 local SLOT_ICON_POSITIONS = {
     arcane = { x = 0.160, z = 0.65},
     ally = { x = -0.073, z = 0.65}
-  }
-  
-local selectedSlot = ""
+}
+
+-- These match with ArkhamDB's way of storing the data in the dropdown menu
+local slotIndices = {arcane = "1", ally = "0", none = ""}
```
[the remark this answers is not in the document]
Entrox-Licher: This is mainly for consistency & clarity (i,e, whenever the selectedSlot variable is changed, it is set to a value from this table).  It is used in the clickArcane and clickAlly button functions

Chr1Z93 (approved):

Buhallin (approved):

PR #231, opened 2023-03-09
Trash: adding context menu function to empty

Chr1Z93:
This function removes all objects from the trash can.
image


BootleggerFinn (approved):
Fantastic Addition!

PR #238, opened 2023-03-16
Colorblindfriendly action tokens

Chr1Z93:
- unified the "grey" color
- turned interior white parts into black

Preview of new backs:
image


Buhallin:
Why no icon on the back of the neutral(?) token?  One up and in from the bottom left

Buhallin (approved):

PR #240, opened 2023-03-17
Data Helper: uncomment deprecated tables for tokenmanager

Chr1Z93:


src/core/DataHelper.ttslua
```diff
@@ -550,28 +546,24 @@ function updateHiddenCards(args)
...
 -- deprecated, use metadata (GM Notes) instead
 -- Known locations and clues. We check this to determine if we should atttempt to spawn clues,
 -- first we look for <LOCATION_NAME>_<GUID> and if we find nothing we look for <LOCATION_NAME>
 -- format is [location_guid -> clueCount]
--- LOCATIONS_DATA = JSON.decode([[{
---   "XXXX": {"type": "fixed", "value": 2, "clueSide": "back"},
---   "xxx": {"type": "perPlayer", "value": 2, "clueSide": "back"}
--- }]])
+LOCATIONS_DATA = JSON.decode([[{
+  "XXXX": {"type": "fixed", "value": 2, "clueSide": "back"},
```
Buhallin, line 557: This is because the table is referenced and is nil if unassigned, right?  Let's remove these dummy values from the actual code, and beef up the documentation to include specific examples.  Same for below.

Chr1Z93: Updated

Buhallin (approved):

PR #241, opened 2023-03-19
Damage token capitalization

Chr1Z93:
Changes "Damage Tokens" to "Damage tokens" for consistency

image


config.json
```diff
@@ -84,12 +84,12 @@
...
     "Horrortokens.7bd2a0",
     "Doomtokens.16fcd6",
-    "DamageTokens.93f4a0",
+    "Damagetokens.93f4a0",
     "Resourcetokens.fd617a",
     "Cluetokens.31fa39",
     "Horrortokens.c3ecf4",
     "Doomtokens.47ffc3",
-    "DamageTokens.480bda",
+    "Damagetokens.480bda",
```
Buhallin, line 92: This changes the reference but not the file name, which could be an issue on a case-sensitive file system.  Should update the file names as well

Chr1Z93: Updated

Buhallin (approved):

PR #243, opened 2023-03-20
Token Arranger Percentages

Entrox-Licher:
Added percentages to the token arranger.

This will show the percentage of drawing a token from a given row (i.e. of any given modifier value).  There is also a separate option to also view cumulative percentages (with an included warning that these percentages will be unreliable with tokens that draw more tokens, such as bless/curse). Both options are individually toggleable via context menu items.

In order to ensure numbers were properly aligned, I forced a specific number of digits per number, by adding both trailing and leading 0s (padding with spaces sadly did not ensure the correct alignment).  It does look a bit funny sometimes, but I think it is preferable for it to be aligned (though I'm open to suggestions if others disagree).

Percentages
Screenshot 2023-03-21 105758

Cumulative Percentages
Screenshot 2023-03-21 110057

Context Menu
Screenshot 2023-03-21 110121


Buhallin:
Can you add a few screenshots showing the new look?  Easier to quickly understand what changes are being made

src/accessories/TokenArranger.ttslua
```diff
@@ -20,6 +20,8 @@ inputParameters.tab             = 2
 
 local latestLoad = "XXX"
 local updating = false
+local percentage = false
+local cumulative = false
```
Chr1Z93, line 24: Imo it would be better if this was combined into one variable, for example:
disabled (no percentages): `showPercentages = false`
enabled (specific values): `showPercentages = "specific"`
enabled (cumulative values): `showPercentages = "cumulative"`

src/accessories/TokenArranger.ttslua
```diff
@@ -157,6 +145,56 @@ function loadDefaultValues()
...
+    end
+
+    buttonParameters.click_function = attachIndex("tokenClick", i)
+    self.createButton(buttonParameters)
+
+    if loadInputs then
+      inputParameters.input_function  = attachIndex2("tokenInput", i)
+      inputParameters.value           = tokenPrecedence[TOKEN_NAMES[i]][1]
+      self.createInput(inputParameters)
+    end
```
Chr1Z93, line 183: What's the use of this? If I understand it correctly, this optionally removes the input fields.

src/accessories/TokenArranger.ttslua
```diff
@@ -244,17 +284,70 @@ function layout(_, _, isRightClick)
   local pos           = self.getPosition() + Vector(3.55, -0.05, -3.95)
   local location      = { x = pos.x, y = pos.y, z = pos.z }
   local current_value = data[1].value
+  local token_count = 0
+	local value_count = 1
+	local cumulative_percentage = 0
```
Chr1Z93, line 289: Here and following: Make sure the indentation is consistent and aligns with the existing code.

src/accessories/TokenArranger.ttslua
```diff
@@ -244,17 +284,70 @@ function layout(_, _, isRightClick)
...
+				cumulative_percentage = cumulative_percentage + math.floor((token_count / #data) * 10000) / 100
+				self.createButton({
+					label = string.format("%s", string.format("%05.2f", math.floor((token_count / #data) * 10000) / 100) .. "%"),
+					click_function = "print",
+					width = 0,
+					height = 0,
+					scale = {2, 2, 2},
+					font_color = {r=1,g=1,b=1},
+					position = (Vector(2.2, -0.05, -2.675) + Vector((0.875 * token_count) + 0.1, 0, 0.875 * value_count))
+				})
```
Chr1Z93, line 303: It would be nice if this was turned into a function instead, since this button creation is used 4 times in the script

Chr1Z93 (changes requested):

src/accessories/TokenArranger.ttslua
Entrox-Licher: Excellent idea!  Implemented

src/accessories/TokenArranger.ttslua
Entrox-Licher: This is so that the button creation can all be one function, but so the input fields are only created onLoad.  

This is because the way the script handles it is that all buttons are deleted whenever the copied tokens are deleted (since the percentage text is on buttons) and then the core buttons are immediately recreated.  This is to avoid having to deal with button indexing, which can get a bit hairy in my experience (so I just use the clearButtons() method).  No reason to do that to the input fields though, so they only need to be created once; however, since they depend on similar local variables, I saw fit to put them in the function as well.  I can try to carve this bit out and put it back in onLoad if you think that would help clarity though.

src/accessories/TokenArranger.ttslua
Entrox-Licher: Went through and did a check, I believe it should be consistent now

src/accessories/TokenArranger.ttslua
Entrox-Licher: Great suggestion, implemented

src/accessories/TokenArranger.ttslua
Chr1Z93: Ah, now I see what this is about. Since the buttons are cleared when updating the percentages, we only need to recreate the buttons and not the input field.
I think it would be a bit clearer if the function name indicated this (maybe `createButtonsAndInputs()`?). Could also split the input field creation into its own function (and only call that in onLoad() and not at the later point). Alternatively, you could also work with `editButton()` instead of removing all of them (though this is probably a bit cumbersome).

Entrox-Licher:
> Couple of UI suggestions, and they're only that :)
> 
> * Make the third menu option "Toggle Cumulative" only so it doesn't wrap.
> * Consider moving all the options to button for visibility (Percentage could be a multi-state toggle).  This will require an image update so likely a followon from Chr1Z.  Would also allow the percentage to be done as Off/Base/Cumulative/Both.
> * Move the percentages to the left of the tokens so they present in an aligned column.  This will require bumping the token start point to the right when they're active, but the jagged position is tricky to read.  This would probably require the cumulative to be beneath the basic chance.

First point, good idea haha.  If you look really closely I did a very hacky trick with underscores to even get it to align properly XD.  

Second point, totally neutral on, don't mind either way.  I originally had it as a button (though not multi-state) and Chr1Z suggested a context menu.  I don't really prefer it one way or the other, so I'm happy to whip up the code for the multi-state if that is the preference. 

For the third point I'll admit, I'm a bit more partial to the text after the tokens even after seeing it here.  That being said, I don't feel super strongly about it, so if you all think it helps legibility, I'm happy to commit the change.
Screenshot 2023-03-21 182633
Screenshot 2023-03-21 182751

src/accessories/TokenArranger.ttslua
Entrox-Licher: Yeah, editButton() I want to avoid because button indexing is a bit of a mess in my experience. I've decided to stick to renaming the function for added clarity.  Will update once commit is pushed

Chr1Z93:
I think the percentages on the left make it was easier to read. Would they still be legible if they were half size (so that then the specific and cumulative value can be stacked on top of each other)?
My recommendation for the context menu was based on the UI with the floating button you had made 😉
While I am not against updating the image to make more space for buttons, I don't think it's worth the effort at this point since there will be a combining with the chaos bag manager in the future which will need a new UI anyway.

Chr1Z93:
> Yeah, I think I agree on the effort to remake the whole image. If we're not going down that route do we want to stick to context menu then, or try the floating (multi-state) button? I do think the multi-state button might be more intuitive to use, but the context menu feels like the cleaner UI choice.
> 
> Here it is with the percentages stacked on top. The size still leaves it legible but it feels slightly cluttered here maybe?
> 
> Screenshot 2023-03-22 113800

I prefer the context menu over the floating button.
Stacked percentages look more concise, but definitely a bit "compressed". I'm fine with either version (stacked or two columns), feel free to choose what you prefer.

Chr1Z93:
> The logic here is: Elder Sign: Blue Auto-fail: Red Symbol (of any kind, even if mixed in with numbers): Purple Only Number: Yellow
> 
> What do you all think?

I think adding color is agreat touch!

src/accessories/TokenArranger.ttslua
Entrox-Licher: Done!

Chr1Z93 (approved):

PR #244, opened 2023-03-24
Add alt view angle for sideways decks

Chr1Z93:
This sets the alt view angle for sideways decks (namely the stacks of investigators spawned by the playercard panel) so that the preview is correctly rotated.

Preview of difference:
https://user-images.githubusercontent.com/97286811/227521058-95ca67a8-78c0-4732-af62-0f801cd668c9.mp4

### DIFF
--- a/src/playercards/PlayerCardSpawner.ttslua
+++ b/src/playercards/PlayerCardSpawner.ttslua
@@ -37,7 +37,7 @@ Spawner.spawnCards = function(cardList, pos, rot, sort, callback)
   -- Spawn each of the three types individually.  Each Y position shift accounts for the thickness
   -- of the spawned deck
   local position = { x = pos.x, y = pos.y, z = pos.z }
-  Spawner.spawn(investigatorCards, position, { rot.x, rot.y - 90, rot.z}, callback)
+  Spawner.spawn(investigatorCards, position, { rot.x, rot.y - 90, rot.z }, callback, true)
 
   position.y = position.y + (#investigatorCards + #standardCards) * 0.07
   Spawner.spawn(standardCards, position, rot, callback)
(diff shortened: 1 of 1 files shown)


src/playercards/PlayerCardSpawner.ttslua
```diff
@@ -82,7 +82,7 @@ end
 -- @param pos Position table where the cards should be spawned (global)
 -- @param rot Rotation table for the orientation of the spawned cards (global)
 -- @param callback Function, callback to be called after the card/deck spawns.
```
Buhallin, line 84: Add doc entry for the new parameter (if it stays)

src/playercards/PlayerCardSpawner.ttslua
```diff
@@ -82,7 +82,7 @@ end
...
-Spawner.spawn = function(cardList, pos, rot, callback)
+Spawner.spawn = function(cardList, pos, rot, callback, sidewaysDeck)
```
Buhallin, line 85: Is there a way we can inspect the cards in the pile rather than relying on a passed parameter?

src/playercards/PlayerCardSpawner.ttslua
```diff
@@ -105,8 +105,16 @@ Spawner.spawn = function(cardList, pos, rot, callback)
     scaleY = 1,
     scaleZ = cardList[1].data.Transform.scaleZ,
   }
+  local sidewaysDeck = true
   for _, spawnCard in ipairs(cardList) do
     Spawner.addCardToDeck(deck, spawnCard.data)
+    if not spawnCard.data.SidewaysCard then
```
Buhallin, line 111: If you want to get fancy here, the if can be replaced with one line:
sidewaysDeck = sidewaysDeck and spawnCard.data.SidewaysCard

(Hopefully got the Lua syntax right there)

Buhallin (approved):

PR #246, opened 2023-03-27
Token Arranger: code cleanup, variable renaming

Chr1Z93:
Does not change functionality.
- cleanup to bring the old layout code in line with the new code
- minor optimization
- addition of more documentation

### DIFF
--- a/src/accessories/TokenArranger.ttslua
+++ b/src/accessories/TokenArranger.ttslua
@@ -74,6 +74,8 @@ function onLoad(saveState)
       percentage = false
     else
       percentage = "basic"
+      broadcastToAll("Percentages are unreliable when using tokens that draw other tokens (bless or curse for example).",
+      "Yellow")
     end
     layout()
   end)
@@ -83,8 +85,9 @@ function onLoad(saveState)
       percentage = "basic"
     else
       percentage = "cumulative"
-      broadcastToAll("Cumulative percentages are unreliable when using tokens that draw other tokens (bless or curse for example).", Color.Yellow)
-    end 
+    end
+    broadcastToAll("Percentages are unreliable when using tokens that draw other tokens (bless or curse for example).",
+    "Yellow")
     layout()
   end)
 
@@ -209,16 +212,12 @@ end
 
 -- order function for data sorting
 function tokenValueComparator(left, right)
-  if left.value > right.value then
-    return true
-  elseif right.value > left.value then
-    return false
-  elseif left.order < right.order then
-    return true
-  elseif right.order < left.order then
-    return false
+  if (left.value ~= right.value) then
+    return left.value > right.value
+  elseif left.order ~= right.order then
+    return left.order < right.order
   else
-    return left.token.getGUID() > right.token.getGUID()
+    return false
   end
 end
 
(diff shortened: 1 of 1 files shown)


src/accessories/TokenArranger.ttslua
```diff
@@ -54,41 +58,45 @@ function onLoad(saveState)
...
-      percentage = "cumulative"
-      broadcastToAll("Cumulative percentages are unreliable when using tokens that draw other tokens (bless or curse for example)", Color.Yellow)
-    else
+
+  self.addContextMenuItem("Toggle cumulative", function() 
+    if percentage == "cumulative" then
       percentage = "basic"
+    else
+      percentage = "cumulative"
+      broadcastToAll("Cumulative percentages are unreliable when using tokens that draw other tokens (bless or curse for example).", Color.Yellow)
```
Buhallin, line 86: This is true for any of the percentages, not just cumulative

src/accessories/TokenArranger.ttslua
```diff
@@ -131,81 +142,73 @@ end
...
+      index = i - 1,
+      value = tokenPrecedence[TOKEN_NAMES[i]][1]
+    })
   end
 end
 
 -- order function for data sorting
-function token_value_comparator(left, right)
+function tokenValueComparator(left, right)
   if left.value > right.value then
```
Buhallin, line 212: Two separate readability suggestions here:
Keep the left/right consistent all the way down:
if left.value > right.value then...
elseif left.value < right.value

The gt/lt and order flip is easy to miss

Whole thing could be a little more compact (and avoid the above issue) as:
if (left.value ~= right.value) then
  return left.value > right.value
elseif left.order ~= right.order
  return left.order < right.order
else
  return left.token.getGUID() > right.token.getGUID()
end
Though I don't think the GUID comparison is necessary, since identical tokens shouldn't really matter what order they come out in, and if the left/right flips it shouldn't actually change anything?  But if there's something attached to them where a shift could actually impact something it's fine to leave

src/accessories/TokenArranger.ttslua
```diff
@@ -315,66 +320,68 @@ function layout(_, _, isRightClick)
...
-	local value_count = 1
-	local cumulative_percentage = 0
-  local current_token = false
+  local pos = self.getPosition() + Vector(3.55, -0.05, -3.95)
+  if percentage then pos.z = pos.z - 3.05 end
+
+  local location = { x = pos.x, y = pos.y, z = pos.z }
+  local currentValue = data[1].value
+  local tokenCount = 0
+	local valueCount = 1
```
Buhallin, line 332: Why the indent here?

src/accessories/TokenArranger.ttslua
```diff
@@ -315,66 +320,68 @@ function layout(_, _, isRightClick)
...
     end
+    
     item.token.setPosition(location)
     item.token.setRotation(self.getRotation())
     location.z = location.z - 1.75
-    token_count = token_count + 1
+    tokenCount = tokenCount + 1
+    tokenName  = item.token.getName()
+
+    -- set tokenName to false if it does not contain letters
```
Buhallin, line 360: Comment would be more helpful as a "why" it's doing this

src/accessories/TokenArranger.ttslua
Chr1Z93: Weird, that's not the case for me locally: 
image

BootleggerFinn (approved):

PR #248, opened 2023-04-03
Automatically remove tokens on discarded cards

Chr1Z93:
Demo video:

https://user-images.githubusercontent.com/97286811/229492390-6ec45e72-ad71-48fb-93b7-dc87559a5e51.mov

### DIFF
--- a/src/playermat/Playmat.ttslua
+++ b/src/playermat/Playmat.ttslua
@@ -567,6 +567,7 @@ function onCollisionEnter(collision_info)
 
   if isInDeckZone(object) then
     tokenManager.resetTokensSpawned(object)
+    removeTokensFromObject(object)
   elseif shouldSpawnTokens(object) then
     spawnTokensFor(object)
   end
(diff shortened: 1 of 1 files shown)


src/playermat/Playmat.ttslua
```diff
@@ -626,6 +637,20 @@ function isInDeckZone(checkObject)
   return false
 end
 
+-- removes tokens from the provided card/deck
+function removeTokensFromObject(object)
+  for _, v in ipairs(searchArea(object.getPosition(), { 3, 1, 4 })) do
+    local obj = v.hit_object
+    if obj.getGUID() ~= "4ee1f2" and
```
Buhallin, line 644: Add a comment for what this GUID is

SCED.wiki
Buhallin, line 1: Think this needs to be reverted?

Chr1Z93: Looks weird on GitHub, but I didn't commit anything related to the wiki. Maybe this is just the first PR after the wiki was added? Seems like there is nothing changed with this

Buhallin (approved):

PR #249, opened 2023-04-03
Adding soundcube

Chr1Z93:
This PR adds a soundcube, which contains the following sounds:
- a vacuum noise (used by the Clean Up Helper)
- a deep bell sound (triggered with the Title Splash Screen for new scenarios)

Open for discussion about different or additional sounds, so I'll make this a "Draft" for now!


src/accessories/CleanUpHelper.ttslua
```diff
@@ -185,6 +186,13 @@ end
 -- modular functions, called by other functions
 ---------------------------------------------------------
 
+function triggerSoundCube()
+  local soundCube = getObjectsWithTag("SoundCube")[1]
+  if soundCube then
+    soundCube.AssetBundle.playTriggerEffect(0)
```
Buhallin, line 192: Trying to remember (and avoid changes to) these indexes is going to be fragile.  Suggest:
- Add code and an Api to the SoundCube with "playVacuum", "tollTheBell", etc
- Reference the Api and use those methods

Buhallin:
I think we need to be careful about going overboard with sounds, as they're purely fluff and could be an issue for people who use voice chat.  That said, these are both once-per-scenario sounds and I think they're fine.

src/accessories/CleanUpHelper.ttslua
Chr1Z93: Yeah, that is a good idea (and I thought of it too after I created the PR 😅)

Chr1Z93:
> I think we need to be careful about going overboard with sounds, as they're purely fluff and could be an issue for people who use voice chat. That said, these are both once-per-scenario sounds and I think they're fine.

That's fair. I think we could have a "General" setting to disable all sounds.

Buhallin:
> That's fair. I think we could have a "General" setting to disable all sounds. EDIT: Added this option

I know you already did it, but I really don't think it's necessary at this point.  We have two sounds per scenario right now - even if someone dislikes it the level of annoyance is trivial and an option is unnecessary.  If we continue to add more sounds it might become necessary in the future, but I don't think it's worth adding an option at this point.

Chr1Z93:
Totally agree that we don't want an option for these two sounds. I added it because I expected more sounds to get requested quickly. Since that wasn't the case I'll remove the option 😅

BootleggerFinn (approved):
Hooray for the soundcube

BootleggerFinn:
I definitely would still like to request a certain sound at some point as an option, but that's not something that has to hold this up and there's not much need for options until then, I suppose.

Chr1Z93:
> I definitely would still like to request a certain sound at some point as an option, but that's not something that has to hold this up and there's not much need for options until then, I suppose.

Yeah, I'm open to adding more sounds at any point. This is the framework for that and will make future additions very easy.

PR #263, opened 2023-04-14
[WIP] update notification

Chr1Z93:
This implements a update notification if an outdated mod is loaded.
The notification is dynamically populated with release notes for the new version.
The window can be closed or marked as "don't show again".

Current preview:
image

To-Do:
- text-spacing
- only show highlights unless a "Details" button is pressed
- positioning on screen (upper left corner maybe?)


Chr1Z93:
Note: This will not show unless you are on an outdated version! You can comment out the highlighted line (in the Global script) to see the notification (for testing).
image

src/core/Global.ttslua
```diff
@@ -1035,3 +1039,92 @@ function titleSplash(scenarioName)
...
+  -- global variable to make it accessible for other functions
+  modMeta = JSON.decode(request.text)
+
+  -- stop here if on latest version or "don't show again" was clicked
+  if MOD_VERSION == modMeta["latestVersion"] then return end
+  
+  if dontNotify[modMeta["latestVersion"]] then return end
+
+  -- end here if no release notes are found
+  if modMeta["releaseNotes"][modMeta["latestVersion"]] == nil then
```
Buhallin, line 1068: The releaseNotes node seems unnecessary.  We shouldn't have more than a single version recorded there, so Highlights can be at the same level as latestVersion.  I think Details can be removed as well, since it's not used?

src/core/Global.ttslua
```diff
@@ -1035,3 +1039,92 @@ function titleSplash(scenarioName)
...
+
+    -- delay to avoid lagging during onLoad()
+    Wait.time(function() UI.show("FinnIcon") end, 1)
+  end
+end
+
+-- updates the XML update notification based on the mod metadata
+function updateNotificationLoading()
+  -- grab data
+  local highlights = modMeta["releaseNotes"][modMeta["latestVersion"]]["Highlights"]
```
Buhallin, line 1081: highlights should be camelCase

src/core/Global.ttslua
```diff
@@ -614,21 +622,17 @@ end
...
 
-local source_repo = 'https://raw.githubusercontent.com/chr1z93/loadable-objects/main'
-local library = nil
-local request_obj
-
 function onClick_refreshList()
-  local request = WebRequest.get(source_repo .. '/library.json', completed_list_update)
-  request_obj = request
+  local request = WebRequest.get(SOURCE_REPO .. '/library.json', completed_list_update)
+  requestObj = request
```
Buhallin, line 627: Please try and keep no-op refactorings in separate PRs

xml/OptionPanel.xml
```diff
@@ -20,7 +20,7 @@
     outline="grey"
     showAnimation="SlideIn_Right"
     hideAnimation="SlideOut_Right"
-    animationDuration="0.1" />
+    animationDuration="0.2" />
```
Buhallin, line 23: Unrelated change, is this intentional?

src/core/Global.ttslua
```diff
@@ -39,6 +39,12 @@ local mythosAreaApi = require("core/MythosAreaApi")
 local tokenArrangerApi = require("accessories/TokenArrangerApi")
 local blessCurseManagerApi = require("chaosbag/BlessCurseManagerApi")
 
+-- online functionality related variables
+local MOD_VERSION = "3.1.0"
+local SOURCE_REPO = 'https://raw.githubusercontent.com/chr1z93/loadable-objects/main'
+local library, requestObj, modMeta, notificationVisible
+local dontNotify = {}
```
Buhallin, line 46: Prefer positive logic here rather than negative (which is often confusing).  acknowledgedUpgradeVersions avoids the negative and also explains what is stored in the list

src/core/Global.ttslua
```diff
@@ -1035,3 +1039,92 @@ function titleSplash(scenarioName)
...
+
+-- updates the XML update notification based on the mod metadata
+function updateNotificationLoading()
+  -- grab data
+  local highlights = modMeta["releaseNotes"][modMeta["latestVersion"]]["Highlights"]
+  local ui = UI.getXmlTable()
+
+  -- update the header
+  local header = find_tag_with_id(ui, 'notificationHeader')
+  header.value = header.value .. modMeta["latestVersion"]
```
Buhallin, line 1086: Any advantage to compositing the header string like this?  It splits the text value (some here, some XML), and while I can't find a path to make it happen runs the risk of this triggering multiple times so you'd get something like New version available: 3.1.03.1.0.  Simpler and safer to just set the full string here.

src/core/Global.ttslua
```diff
@@ -1035,3 +1039,92 @@ function titleSplash(scenarioName)
...
+    -- delay to avoid lagging during onLoad()
+    Wait.time(function() UI.show("FinnIcon") end, 1)
+  end
+end
+
+-- updates the XML update notification based on the mod metadata
+function updateNotificationLoading()
+  -- grab data
+  local highlights = modMeta["releaseNotes"][modMeta["latestVersion"]]["Highlights"]
+  local ui = UI.getXmlTable()
```
Buhallin, line 1082: Reading and writing the entire XML table is very inefficient.  You can use UI.setAttribute() to update only the specific values you care about.  This will also remove the need for find_tag_with_id, which it very inefficient itself -  it's an O(n) lookup, which makes the entire operation O(n^2)

Buhallin (changes requested):

src/core/Global.ttslua
Chr1Z93: Will do 🫡

src/core/Global.ttslua
Chr1Z93: I intend this file to be used for the whole release notes in the future. E.g. by being able to select a version in a drop-down menu and then seeing the notes for that version (which "Details" is for)

xml/OptionPanel.xml
Chr1Z93: Against the principle, but yeah this was intended. Upon testing with the OptionPanel it felt a bit too quick on the animation.

src/core/Global.ttslua
Chr1Z93: Wow that's an awesome suggestion! I didn't know these awesome functions before

src/core/Global.ttslua
Buhallin: Fair on Details, but it still shouldn't need to be a table, right?  Unless we're going to have multiple versions (and I don't think there's a need to), that layer is redundant.  Top-level fields can be latestVersion (or even just version), highlights, and details.

Chr1Z93: Why wouldn't we have multiple entries? If we don't, you can only view the release notes of you are on the newest version.

Buhallin (approved):

PR #270, opened 2023-04-19
Custom Playmat Images: added TSK and Ghosts of Onigawa, minor layout update

Chr1Z93:
updated layouting to utilize the space to the right instead of starting below it:

image


Tikatoy (approved):
Looks good!

PR #271, opened 2023-04-20
Added Featured Content

BootleggerFinn:
Added featured content section containing featured fan campaigns.  also updated date on release notes. Note that this commit has many files added as part of the two fan campaigns, as they have been "pre-downloaded".


Chr1Z93:
This does not compile because the placeholder boxes were removed, but their files are still in there. Should probably keep them so that we don't need to readd them next release anyway.

Chr1Z93 (changes requested):
generateMod(<config>) : objects.ParseAllObjectStates() : expected order (177) and db.root (178) to have same length

Chr1Z93 (approved):

PR #272, opened 2023-04-20
updated Phase Tracker

Chr1Z93:
- added context menu option to enable broadcasting
- changed from using states to respawning itself and copies


Tikatoy (approved):
Not ideal to have a context menu on a "button" because of the small clickable region, but given that not many players use it (I think?) and that the setting persists through saves,  I think this is fine. The people who want to use it will just have to try harder! 😆

PR #274, opened 2023-04-20
Playermat: additional condition for discarding tokens

Chr1Z93:
This stops the auto-discard from removing tokens like the keys in TIC and all other custom tokens.


BootleggerFinn (approved):
Good catch!

PR #277, opened 2023-04-27
Tarot Cards: added context menu option to flip preview

Chr1Z93:
The context menu option flips the alt preview so that reversed cards can be read when previewing

### DIFF
--- a/src/playercards/Tarotcard.ttslua
+++ b/src/playercards/Tarotcard.ttslua
@@ -1,0 +1,15 @@
+-- context menu to manually fix rotation
+function onLoad()
+  self.addContextMenuItem("rotate alt preview", reverse)
+end
+
+-- rotates the alt_view_angle
+function reverse()
+  local angle = self.alt_view_angle
+  if angle.y == 0 then
+    angle.y = 180
+  else
+    angle.y = 0
+  end
+  self.alt_view_angle = angle
+end
--- a/src/playercards/Tarotcard.ttslua
+++ b/src/playercards/Tarotcard.ttslua
@@ -1,10 +1,7 @@
 -- context menu to manually fix rotation
 function onLoad()
-  self.addContextMenuItem("rotate preview", rotatePreview)
-  self.addContextMenuItem("rotate card+preview", function()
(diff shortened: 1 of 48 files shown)


src/playercards/Tarotcard.ttslua
```diff
@@ -0,0 +1,19 @@
+-- context menu to manually fix rotation
+function onLoad()
+  self.addContextMenuItem("rotate preview", rotatePreview)
+  self.addContextMenuItem("rotate card+preview", function()
```
Buhallin, line 4: If one is going to be a separate function, this one should be as well.  Reads very weird otherwise

src/playercards/Tarotcard.ttslua
```diff
@@ -0,0 +1,19 @@
...
+    self.setRotationSmooth(self.getRotation() + Vector(0, 180, 0))
+    rotatePreview()
+  end)
+end
+
+-- rotates the alt_view_angle
+function rotatePreview()
+  local angle = self.alt_view_angle
+  if angle.y == 0 then
```
Buhallin, line 13: Probably want a small range here in case the rotation gets imprecise on a save

src/playercards/Tarotcard.ttslua
Chr1Z93: Updated

src/playercards/Tarotcard.ttslua
Chr1Z93: Only coordinates are rounded onSave, this value can't have a value besides 0 (or 180 if changed via this script)

src/playercards/Tarotcard.ttslua
```diff
@@ -0,0 +1,19 @@
+-- context menu to manually fix rotation
+function onLoad()
+  self.addContextMenuItem("rotate preview", rotatePreview)
```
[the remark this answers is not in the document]
Chr1Z93: Updated

Tikatoy (approved):
Looks good!

Would be cool if the inverted tarot cards from the scripted bag have their previews automatically set to rotated, but that's just a "nice to have".

PR #280, opened 2023-05-03
Token Manager: out-commented log statements

Chr1Z93:
Right now playing an asset with uses pulls up the chat (if auto-hide is enabled) because there are log statements to the system console. This was not the case pre 3.1.0 and thus this PR disables the logs from the Token Manager.

### DIFF
--- a/src/core/token/TokenManager.ttslua
+++ b/src/core/token/TokenManager.ttslua
@@ -348,7 +348,8 @@ do
       for i, useInfo in ipairs(uses) do
         type = useInfo.type
         token = useInfo.token
-        tokenCount = (useInfo.count or 0) + (useInfo.countPerInvestigator or 0) * playArea.getInvestigatorCount()
+        tokenCount = (useInfo.count or 0)
+            + (useInfo.countPerInvestigator or 0) * playArea.getInvestigatorCount()
         if extraUses ~= nil and extraUses[type] ~= nil then
           tokenCount = tokenCount + extraUses[type]
         end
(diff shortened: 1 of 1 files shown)


src/core/token/TokenManager.ttslua
```diff
@@ -342,18 +342,17 @@ do
...
-      log("Spawning single use tokens for "..card.getName()..'['..card.getDescription()..']: '..tokenCount.."x "..token)
+      --log("Spawning single use tokens for "..card.getName()..'['..card.getDescription()..']: '..tokenCount.."x "..token)
       TokenManager.spawnTokenGroup(card, token, tokenCount)
     else
       for i, useInfo in ipairs(uses) do
         type = useInfo.type
         token = useInfo.token
-        tokenCount = (useInfo.count or 0)
-            + (useInfo.countPerInvestigator or 0) * playArea.getInvestigatorCount()
+        tokenCount = (useInfo.count or 0) + (useInfo.countPerInvestigator or 0) * playArea.getInvestigatorCount()
```
Buhallin, line 351: Please revert this formatting change

Buhallin:
As a general note here, log() is still visible in-game on another tab in the chat window, and we shouldn't be making these user-visible regardless of this specific issue

src/core/token/TokenManager.ttslua
Chr1Z93: Updated

Chr1Z93:
> As a general note here, log() is still visible in-game on another tab in the chat window, and we shouldn't be making these user-visible regardless of this specific issue

Yeah, I wouldn't mind the log statements if they wouldn't unhide the chat (even when the normal chat tab is selected). Quite the weird behaviour from TTS, but we shouldn't need these log statements outside of development/debugging anyway.

Buhallin (approved):

PR #285, opened 2023-05-11
Upgradesheets: Refactoring + QoL

### LINKED ISSUE #283: Summoned Servitor: Marks on Upgradesheets are shifted when importing from arkhamdb

Chr1Z93:
- refactored the upgradesheets to use a required file for shared code
- increased resolution for inputfields (by rendering larger and scaling down)
- added string conversion for "Grizzled"
- cleaned up related deck importer code
- fixed a bug with "Summoned Servitor" boxes being shifted (closes #283 )


src/playercards/customizable/UpgradeSheetLibrary.ttslua
```diff
@@ -0,0 +1,364 @@
+-- static values
+Y_VISIBLE    = 0.25
+Y_INVISIBLE  = -0.5
+
+-- used for Summoned Servitor and Living Ink
+VECTOR_COLOR = {
+  unselected = { 0.5, 0.5, 0.5, 0.75 },
+  selected   = { 0.597, 0.195, 0.796 }
```
Buhallin, line 8: I picked the purple color here because Ink was a mystic card, it won't be appropriate for others.  Let's key this by class (only need Mystic for now but at least have the option for the future)

src/playercards/customizable/UpgradeSheetLibrary.ttslua
```diff
@@ -0,0 +1,364 @@
...
+    end
+  end
+  return JSON.encode({ markedBoxes, inputValues })
+end
+
+-- Startup procedure
+function onLoad(savedData)
+  if savedData ~= "" and DEBUG == "" then
+    local loadedData = JSON.decode(savedData)
+    markedBoxes      = loadedData[1]
```
Buhallin, line 38: These really shouldn't be separate arrays.  It should be a single array of tables:
{
  xp = 0,
  text = ""
}

This keeps the boxes and field together so there's no need to correlate them, much less correlate them across different indexes which is very confusing.

I'd suggest a different name here as well, something more to what the data represents

Buhallin (changes requested):

src/playercards/customizable/UpgradeSheetLibrary.ttslua
```diff
@@ -0,0 +1,364 @@
+-- static values
+Y_VISIBLE    = 0.25
```
Buhallin, line 2: Can we replace this with transparency as well?

Chr1Z93:
Continued with https://github.com/argonui/SCED/pull/306

PR #289, opened 2023-05-12
added "take a clue" hotkey

Chr1Z93:
This hotkey moves the hovered object (if it is a clue) to your location. If you are instead hovering a location, it takes a clue from the location.


Tikatoy (approved):
Looks good!

PR #291, opened 2023-05-16
"Take a clue" hotkey: handling for stacked clues

### LINKED ISSUE #290: "Take a clue" hotkey: handle stacks of clues on a card
Right now it just looks for the first object with the correct memo and rotation. This can lead to "taking" a stack of clues.
(Note: normally there won't be stacks of clues on cards, but you never know...)

Chr1Z93:
Adds an additional check for the found object to handle stacks of clues.
closes #290


Tikatoy (approved):
Works!

PR #292, opened 2023-05-16
"Move card to victory display" hotkey: correctly rotates cards when placing them and remove tokens from it

Chr1Z93:
- correctly rotate the placed card (preserving the cards z orientation, so that the currently shown face stays the same)
- remove tokens from the card the hotkey was called on
- fixes missing `memo` fields for damage tokens and path tokens in the tokensource


Tikatoy:
Under what circumstances was the card not keeping its orientation when moved to the victory display?

Chr1Z93:
> Under what circumstances was the card not keeping its orientation when moved to the victory display?

Before this PR the card was simply not rotated at all. So a victory enemy from red's threat area would be placed horizontally (instead of up-right). Same deal for exhausted enemies.

Tikatoy (approved):

PR #293, opened 2023-05-20
"Auto-fail" exclusion for "Unrelenting (1)"

Chr1Z93:
This adds the missing exclusion for the "Auto-fail" to "Unrelenting", so that it doesn't show up as option in the context menu for sealing.


Tikatoy (approved):
But they can just... not seal the autofail 😆

PR #294, opened 2023-05-22
Token-Arranger: removes faulty onExit handling from MythosArea

Chr1Z93:
Removed code was added to "reset" the internal token data variable. As it turns out, this is also triggered when flipping the scenario card and thus stops the updating onFlip.

### DIFF
--- a/src/core/MythosArea.ttslua
+++ b/src/core/MythosArea.ttslua
@@ -12,9 +12,7 @@ local ENCOUNTER_DISCARD_AREA = {
   lowerRight = { x = 1.58, z = 0.38 },
 }
 
-local currentScenario
-local useFrontData
-local tokenData
+local currentScenario, useFrontData, tokenData
 
 local TRASHCAN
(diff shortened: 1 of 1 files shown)


src/core/MythosArea.ttslua
```diff
@@ -12,9 +12,7 @@ local ENCOUNTER_DISCARD_AREA = {
   lowerRight = { x = 1.58, z = 0.38 },
 }
 
-local currentScenario
-local useFrontData
-local tokenData
+local currentScenario, useFrontData, tokenData
```
Buhallin, line 15: Opinions vary on this so up to you, but unless these are tightly related variables one-per-line is usually more readable.  e.g. I might do "local x, y" but not much more than that.

Buhallin (approved):

src/core/MythosArea.ttslua
Chr1Z93: Updated

PR #295, opened 2023-05-22
updated scenario / campaign box placeholder art

Chr1Z93:
This is to match the updated containers with the downloadable versions.


Tikatoy (approved):
Placeholders match.

PR #297, opened 2023-06-11
New Tokens

Chr1Z93:
Including the amazing tokens Dran made!
Ammo, Charge, Secret and Supply (more to come?):

image

### DIFF
--- a/objects/AllPlayerCards.15bb07/18Derringer.101a41.gmnotes
+++ b/objects/AllPlayerCards.15bb07/18Derringer.101a41.gmnotes
@@ -10,7 +10,7 @@
     {
       "count": 2,
       "type": "Ammo",
-      "token": "resource"
+      "token": "ammo"
     }
   ],
   "cycle": "Investigator Packs"
--- a/objects/AllPlayerCards.15bb07/18Derringer2.f8a977.gmnotes
+++ b/objects/AllPlayerCards.15bb07/18Derringer2.f8a977.gmnotes
@@ -11,7 +11,7 @@
     {
       "count": 3,
       "type": "Ammo",
-      "token": "resource"
+      "token": "ammo"
     }
   ],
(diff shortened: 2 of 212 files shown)


Buhallin:
Rather than proliferating the number of token types, we should make these states.  The TokenSpawner can then set the state based on the metadata

Chr1Z93:
> Rather than proliferating the number of token types, we should make these states. The TokenSpawner can then set the state based on the metadata

PR is updated, tokens are states now.

src/core/token/TokenManager.ttslua
```diff
@@ -325,38 +333,28 @@ do
...
-    -- multi-spawn vs. single spawn otherwise.  TODO: Clean this up when 2.4.0 has been out long
-    -- enough that saved decks don't have old data
-    if uses.count != nil then
-      type = cardMetadata.uses.type
-      token = cardMetadata.uses.token
-      tokenCount = cardMetadata.uses.count
+    if uses == nil then return end
+
+    -- get the stateID for the additional resource tokens
+    local stateTable = {
```
Buhallin, line 339: Put this mapping at the top so it's easy to find

src/core/token/TokenManager.ttslua
```diff
@@ -219,8 +221,14 @@ do
       return
     end
 
+    -- this is used to load the correct state for additional resource tokens (e.g. "Ammo")
+    local callback = nil
+    if stateID ~= nil then
```
Buhallin, line 226: Check for the "resource" type here as well.  If we ended up with mismatched data (e.g. "clue"/"Charge") it would throw an error

Chr1Z93: Good idea! Updated

Buhallin (approved):

PR #300, opened 2023-06-11
Chaos Token Stat Tracker: sorted printing

Chr1Z93:
This sorts the print from the chaos token start tracker by prepopulating the table. Additionally, it will now omit the stat printing for players that haven't drawn any tokens (e.g. playmats that are not used).

image

### DIFF
--- a/src/core/Global.ttslua
+++ b/src/core/Global.ttslua
@@ -370,7 +370,7 @@ function handleStatTrackerClick(_, _, isRightClick)
         playerColor = playmatApi.getPlayerColor(MAT_GUID_TO_COLOR[key])
         playerName = Player[playerColor].steam_name or playerColor
 
-        local playerSquidCount = personalStats["Auto-fail"] or 0
+        local playerSquidCount = personalStats["Auto-fail"]
         if playerSquidCount > maxSquid then
           squidKing = playerName
           maxSquid = playerSquidCount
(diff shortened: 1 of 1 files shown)


src/core/Global.ttslua
```diff
@@ -353,42 +354,89 @@ end
...
+      local playerColor, playerName
+
+      if key == "Overall" then
+        playerColor = "White"
+        playerName = "Overall"
+      else
+        playerColor = playmatApi.getPlayerColor(MAT_GUID_TO_COLOR[key])
+        playerName = Player[playerColor].steam_name or playerColor
+
+        local playerSquidCount = personalStats["Auto-fail"] or 0
```
Buhallin, line 373: Nit: Your new initialization guarantees a set of 0s for all tokens, so the "or 0" isn't necessary

src/core/Global.ttslua
```diff
@@ -353,42 +354,89 @@ end
...
     end
-    printToAll("------------------------------")
-    printToAll(squidKing .. " is an auto-fail magnet.", {255, 0, 0})
+
+    -- detect if any player drew tokens
+    if foundAnyStats then
+      printToAll("------------------------------")
+      printToAll(squidKing .. " is an auto-fail magnet.", {255, 0, 0})
+    else
+      printToAll("Couldn't find any stats for chaos token draws.")
```
Buhallin, line 406: Maybe "No tokens have been drawn yet"?

src/core/Global.ttslua
```diff
@@ -353,42 +354,89 @@ end
...
+    end
+  end
+end
+
+-- resets the count for each token to 0
+function resetChaosTokenStatTracker()
+  local tokenNameList = {
```
Buhallin, line 413: This should be at the top as a constant, but also do we really not have this somewhere else already?

Buhallin (approved):

PR #298, opened 2023-06-11
Replacing featured content with placeholder boxes

Chr1Z93:
As already discussed, this replaces the featured fan-made content with placeholder boxes (mainly so that these can be updated remotely).
This additionally removes the release notes for version 3.1.0 in preparation for the next release.

PR #302, opened 2023-06-11
Added a Campaign Importer/Exporter!

Entrox-Licher:
Fully implemented a Campaign Importer/Exporter!  This allows players to save/export their campaign into a single object they can save into their Saved Objects and then use the tool to load/import their campaign into a fresh save of the mod.  Currently tracked items include:

- Campaign Box (Will run Place for the correct campaign box onto the table)
- Campaign Log (Will spawn in campaign log with all correct fields)
- Chaos Bag (Will spawn in chaos bag with all correct tokens)
- Trauma Values (Will tick up damage/horror according to trauma, following same campaign log order as the Cleanup Tool, for consistency)
- ArkhamDB DeckIDs (Will fill in the deck ids into the ArkhamDB Importer, but not actually import them)
- Investigator Counter (Will update main play area investigator counter for clue spawning)
- Campaign Guide (Will set the campaign guide to the page it was last open to)
- Options Panel! (Will load in ALL options from the options panel to whatever they were previously set to)

Please feel free to review the code, but I believe this implementation will make it extremely extensible (if more features want to be added), and require minimal upkeep with future campaigns (all that is needed is a single tag on the Campaign Box, Campaign Log, and Campaign Guide).  This also means it is very easy to add support for it for custom campaigns!

Potential Improvements (please provide feedback on these in addition to any of your own suggestions and feedback):

- Loading/Importing is currently handled with a snap point under the Campaign Importer/Exporter object that is surrounded by a scripting zone.  I think it would probably be better to just make it a single larger object with a 'slot' on the object itself, place the snap point on the object itself, and do away with the scripting zone altogether.  However, I don't have the graphical skills to make an image for that 😅.
- Some of these options should probably be toggleable (like trauma import is with the Cleanup Tool).  Which ones?
- I was also considering have the SCED Tour object be deleted if it was deleted in the saved campaign, but I was hesitant to make this change and undermine the cool new tour.
- Positioning on the table; I think next to the deck importer makes sense, but I'm not sure.
- Visibility; should it be hidden into the options menu like the cleanup tool or available on the table by default?  I think probably the former, but I'm not sure.
- Other options; none come to mind past toggling certain items, but are there other relevant options to control/customize the behaviour?

Thanks in advance for any feedback, and hope this feature is helpful!


objects/Fan-MadeScenariosCampaignsMiscellany.66e97c/Fan-MadeCampaigns.89c32e.json
```diff
@@ -43,7 +43,7 @@
     "CloseEncountersoftheLV-426Kind.4f5421",
     "CelticRising.4d305a",
     "CallofthePlaguebearer.613b64",
-    "BloodborneCityoftheUnseen.24fb2b",
+    "Bloodborne-CityoftheUnseen.24fb2b",
```
Chr1Z93, line 46: This should be removed from this PR since it stop the mod from bundling. Known issue that is resolved with https://github.com/argonui/SCED/pull/298

modsettings/SnapPoints.json
```diff
@@ -356,5 +356,12 @@
       "y": 1.481,
       "z": -56.165
     }
+  },
+  {
+    "Position": {
+      "x": -21.251,
+      "y": 1.481,
+      "z": 55.589
+    }
```
Chr1Z93, line 365: I don't think a snap point is needed. Just add a tag to the created coin (e.g. "SCED_Export") and use that to get the token.

objects/LeakedItems.42cd6e/AliceLuxley2.94f23b.json
Chr1Z93, line 1: Should be removed from this PR

objects/LeakedItems.42cd6e/DragonPole3.a20aef.json
Chr1Z93, line 1: Should be removed from this PR

objects/NavigationOverlayHandler.797ede.json
Chr1Z93, line 1: Should be removed from this PR

objects/ScriptingTrigger.ef3b5f.json
Chr1Z93, line 1: Scripting zone can be removed since you have a tag to detect the coin

src/arkhamdb/DeckImporterUi.ttslua
```diff
@@ -154,6 +154,17 @@ function whiteDeckChanged(_, _, inputValue) whiteDeckId = inputValue end
...
+  greenDeckId = idList[1]
+  redDeckId = idList[2]
+  whiteDeckId = idList[3]
+  orangeDeckId = idList[4]
+  self.editInput({index=0, value=greenDeckId})
+  self.editInput({index=1, value=redDeckId})
+  self.editInput({index=2, value=whiteDeckId})
+  self.editInput({index=3, value=orangeDeckId})
+end
+
```
Chr1Z93, line 167: Instead of using a table that depends on order, use a key-value-based table:
```
deckIDs = {
  white = "...",
  orange = "...",
  green = "...",
  red = "..."
}
```
Then you can get rid of LOC 158-161 too and just do:
`self.editInput({index=0, value=deckIDs.green})` etc.

src/core/Global.ttslua
```diff
@@ -983,6 +983,15 @@ function removeHelperObject(name)
...
 
+-- loads saved options
+function loadSettings(newOptions)
+  optionPanel = newOptions
+  updateOptionPanelState()
+  for id, state in pairs(optionPanel) do
+    applyOptionPanelChange(id, state)
+  end
+end
+
```
Chr1Z93, line 994: Not sure what the cause is, but in my testing the handhelper and search assistant were spawned despite the option not being ticked in the original save.

src/accessories/CampaignImporterExporter.ttslua
```diff
@@ -0,0 +1,302 @@
...
+    Transform = {
+      posX = 2.80123329,
+      posY = 1.681688,
+      posZ = -23.6498032,
+      rotX = -2.22745348E-06,
+      rotY = 269.999664,
+      rotZ = -1.6614282E-06,
+      scaleX = 2.00499868,
+      scaleY = 2.00499868,
+      scaleZ = 2.00499868
```
Chr1Z93, line 13: Nitpick: Clean this data up (round scale+position to 2 decimals, rotation to 0)

src/accessories/CampaignImporterExporter.ttslua
```diff
@@ -0,0 +1,302 @@
+local campaignTokenData = {
```
Chr1Z93, line 1: I think you can omit almost all of these entries and it should use the default values instead

src/accessories/CampaignImporterExporter.ttslua
```diff
@@ -0,0 +1,302 @@
...
+local DAMAGE_HORROR_GUIDS = {
+    "eb08d6"; "e64eec"; "1f5a0a"; "591a45";
+    "468e88"; "0257d9"; "7b5729"; "beb964";
+  }
+
+local campaignData = {}
+
+local tokenZone = getObjectFromGUID("ef3b5f")
+local deckImporter = getObjectFromGUID("a28140")
+local investigatorCounter = getObjectFromGUID("f182ee")
```
Chr1Z93, line 83: Don't set objects before `onLoad()` to avoid race conditions. Order of object spawning is not reliable. Just declare variable here and get reference in `onLoad()`

src/accessories/CampaignImporterExporter.ttslua
```diff
@@ -0,0 +1,302 @@
...
+    })
+    self.createButton({
+        click_function = "createCampaignToken",
+        function_owner = self,
+        label = "Export",
+        position = {x=1, y=0.2, z=0},
+        width = 350,
+        height = 150,
+        scale = {2, 1, 2},
+    })
```
Chr1Z93, line 108: The buttons are a bit pixelated. You could make them bigger and then set the scale to be below 1.

src/accessories/CampaignImporterExporter.ttslua
```diff
@@ -0,0 +1,302 @@
...
+end
+
+function findCampaignFromToken(_, _, _)
+    local coin = nil
+    for _, obj in ipairs(tokenZone.getObjects()) do
```
Chr1Z93, line 113: Why do you need the zone if you already have a tag?

src/accessories/CampaignImporterExporter.ttslua
```diff
@@ -0,0 +1,302 @@
...
+    end
+end
+-- checks scripting zone for chaos bag
+function findChaosBag()
+    for _, item in ipairs(getObjectFromGUID("83ef06").getObjects()) do
+      if item.getDescription() == "Chaos Bag" then
+        return item
+      end
+    end
+end
```
Chr1Z93, line 282: Probably better to call the function in `Global` instead

src/accessories/CampaignImporterExporter.ttslua
```diff
@@ -0,0 +1,302 @@
...
+        printToAll("Trauma values found in campaign log!", "Green")
+        for i = 1, 10, 3 do
+            traumaValues[1 + (i - 1) / 3] = counterData.counter[i].value
+            traumaValues[5 + (i - 1) / 3] = counterData.counter[i + 1].value
+        end
+    else
+        printToAll("Trauma values could not be found in campaign log!", "Yellow")
+        printToAll("Default values for health and sanity loaded.", "Yellow")
+    end
+    deckIds = {deckImporter.getInputs()[1].value, deckImporter.getInputs()[2].value, deckImporter.getInputs()[3].value, deckImporter.getInputs()[4].value}
```
Chr1Z93, line 230: Since you added a function to set the values, might as well also make one to get them 😅

src/accessories/CampaignImporterExporter.ttslua
```diff
@@ -0,0 +1,302 @@
...
+        function() createCampaignFromToken(importData) end,
+        function() return findCampaignLog() ~= nil end,
+        2,
+        function() broadcastToAll("Error placing campaign box") end
+    )
+end
+
+function createCampaignFromToken(importData)
+    findCampaignLog().destruct()
+    findChaosBag().destruct()
```
Chr1Z93, line 161: You could instead use the function in `Global` to set the chaos bag state with a string: https://github.com/argonui/SCED/blob/main/src/core/Global.ttslua#L516

This would make the importer/exporter more resilient to future changes to the chaos bag.

src/accessories/CampaignImporterExporter.ttslua
```diff
@@ -0,0 +1,302 @@
...
+local campaignData = {}
+
+local tokenZone = getObjectFromGUID("ef3b5f")
+local deckImporter = getObjectFromGUID("a28140")
+local investigatorCounter = getObjectFromGUID("f182ee")
+
+local campaignBoxGUID = ""
+local traumaValues = false
+local deckIds = false
+local campaignGuidePage = false
```
Chr1Z93, line 88: You're not using this as global variable. It's declared again later.

src/accessories/CampaignImporterExporter.ttslua
```diff
@@ -0,0 +1,302 @@
...
+
+local campaignData = {}
+
+local tokenZone = getObjectFromGUID("ef3b5f")
+local deckImporter = getObjectFromGUID("a28140")
+local investigatorCounter = getObjectFromGUID("f182ee")
+
+local campaignBoxGUID = ""
+local traumaValues = false
+local deckIds = false
```
Chr1Z93, line 87: This does not need to be global (to this tool).

src/accessories/CampaignImporterExporter.ttslua
```diff
@@ -0,0 +1,302 @@
...
+            traumaValues[5 + (i - 1) / 3] = counterData.counter[i + 1].value
+        end
+    else
+        printToAll("Trauma values could not be found in campaign log!", "Yellow")
+        printToAll("Default values for health and sanity loaded.", "Yellow")
+    end
+    deckIds = {deckImporter.getInputs()[1].value, deckImporter.getInputs()[2].value, deckImporter.getInputs()[3].value, deckImporter.getInputs()[4].value}
+    local campaignGuide = findCampaignGuide()
+    if campaignGuide == nil then
+        broadcastToAll("Campaign loguideg not found!", Color.Red)
```
Chr1Z93, line 233: Typo in the message

src/accessories/CampaignImporterExporter.ttslua
```diff
@@ -0,0 +1,302 @@
...
+  }
+
+local campaignData = {}
+
+local tokenZone = getObjectFromGUID("ef3b5f")
+local deckImporter = getObjectFromGUID("a28140")
+local investigatorCounter = getObjectFromGUID("f182ee")
+
+local campaignBoxGUID = ""
+local traumaValues = false
```
Chr1Z93, line 86: This does not need to be global (to this tool).

src/accessories/CampaignImporterExporter.ttslua
Chr1Z93, line 1: Some notes:
- add a description to the tool (and probably also a tooltip to the buttons)
- most likely not much of a difference unless some mass-exports their games: use a datestring at the start of the token name instead of at the end (e.g. 2023-06-11 <campaign name> Save)
- use 2 spaces as indent instead of a full tab
- a few more comments (mostly at the start of a function as general description) would be neat

Chr1Z93 (changes requested):

src/accessories/CampaignImporterExporter.ttslua
```diff
@@ -0,0 +1,302 @@
...
+    spawnObjectData({data = importData[2]}) 
+    --create chaos bag
+    spawnObjectData({data = importData[3]})
+    --populate trauma values
+    if importData[4] then
+        updateCounters(importData[4])
+    end
+    --populate ArkhamDB deck IDs
+    if importData[5] then
+        deckImporter.call("updateDeckIds", importData[5])
```
Buhallin, line 172: Avoid call() directly on other objects.  Use the API pattern instead:
https://docs.google.com/document/d/14ZX5JdIXx0rpGQ2HUQk91hWXBOqtZFY1xXDhwDRz4Z0/edit?usp=sharing

src/accessories/CampaignImporterExporter.ttslua
```diff
@@ -0,0 +1,302 @@
...
+    )
+end
+
+function createCampaignFromToken(importData)
+    findCampaignLog().destruct()
+    findChaosBag().destruct()
+    --create campaign log
+    spawnObjectData({data = importData[2]}) 
+    --create chaos bag
+    spawnObjectData({data = importData[3]})
```
Buhallin, line 165: Use a named table rather than an array for these as well

src/accessories/CampaignImporterExporter.ttslua
Buhallin, line 113: Any reason not to just use getObjectsWithTag here, and avoid both the zone and loop?

src/accessories/CampaignImporterExporter.ttslua
```diff
@@ -0,0 +1,302 @@
...
+        width = 350,
+        height = 150,
+        scale = {2, 1, 2},
+    })
+end
+
+function findCampaignFromToken(_, _, _)
+    local coin = nil
+    for _, obj in ipairs(tokenZone.getObjects()) do
+        if obj.hasTag("ImporterToken") then
```
Buhallin, line 114: Should add an error condition if multiple coins are found, as otherwise the behavior is unclear to the user

src/accessories/CampaignImporterExporter.ttslua
Buhallin, line 230: Don't access these directly, it creates a tight coupling between this and the importer which is both fragile and hard to see.  As suggested, make it a method (and put it in the API)

src/accessories/CampaignImporterExporter.ttslua
```diff
@@ -0,0 +1,302 @@
...
+    end
+    local campaignGuidePage = campaignGuide.Book.getPage()
+
+    campaignData = {
+        campaignBoxGUID,
+        campaignLog.getData(),
+        chaosBag.getData(),
+        traumaValues,
+        deckIds,
+        investigatorCounter.getVar("val"),
```
Buhallin, line 244: This is already exposed via the PlayAreaApi.
https://github.com/argonui/SCED/blob/fc68103f6748b4b85b045a8b0d70e1dd3c4d26cc/src/core/PlayAreaApi.ttslua#L8

Buhallin (dismissed):

Buhallin:
I'm a little concerned with the overall approach on this.  It's good as long as the coverage is there, but misses anything unique.  TSK comes to mind as the main offender, as it won't get the map, any keys that have been placed on the table, a pile of "haven't seen the last of", etc.  Tracked damage on the Harbinger of Valusia comes to mind as well.

I don't have any quick instincts on how to fix this, need to think on it some more

src/accessories/CampaignImporterExporter.ttslua
Buhallin: You can use the getUiState() function to get everything as a single object.  Will need to add one to set it back, but that'll be a lot cleaner than doing the individual values

Buhallin:
The other concern here is versioning.  We're going to have to be aware of what gets saved and how, because this will enable a whole lot of objects that would normally be locked into a specific version to cross that boundary.

The simplest answer here (and probably something I think needs to happen) would be to add a version into the saved data, and only restore a save if it's the same as the current version.

Entrox-Licher:
Thanks all for the comments!  Will dig into these throughout the week to implement your fixes and suggestions



> I'm a little concerned with the overall approach on this. It's good as long as the coverage is there, but misses anything unique. TSK comes to mind as the main offender, as it won't get the map, any keys that have been placed on the table, a pile of "haven't seen the last of", etc. Tracked damage on the Harbinger of Valusia comes to mind as well.
> 
> I don't have any quick instincts on how to fix this, need to think on it some more

This is a very good point.  The 'quick and dirty' solution to this would be to simply alert players to what isn't saved (maybe an alert that pops up specifically when you save and it's TSK that's laid out).  I do believe the TSK map is the only offender here, because almost all other instances I can think of you are specifically instructed to track in the campaign log (including Harbinger damage and TSK keys, and others like EOTE expedition team), though admittedly I haven't played through everything, so I can't say for sure.  It does raise concerns for future campaigns that break from the mold as much as TSK though.

Alternatively, there is a world where I bake in a custom solution for TSK (I don't think it would be too challenging), but it might set a precedent, and I don't necessarily want to add another thing on the already lengthy set of steps involved in getting a new campaign up and running.  I'll put some more thought into this as well throughout the week.


> The other concern here is versioning. We're going to have to be aware of what gets saved and how, because this will enable a whole lot of objects that would normally be locked into a specific version to cross that boundary.
> 
> The simplest answer here (and probably something I think needs to happen) would be to add a version into the saved data, and only restore a save if it's the same as the current version.

This is a valid concern, but I wouldn't want to lock people from loading in saves from previous versions, since I see that as one of the primary features of the tool: being able to upgrade to a new version of the mod in the middle of a campaign.  Additionally I think the risk here is minimal: in the current version the only two objects that could 'cross the boundary' so to speak are the chaos bag and the campaign log.  The campaign log is incredibly unlikely to be an issue, so it really just comes down to the chaos bag, which I believe will be fixed by the above suggestion to use the new tools to adjust the contents of the existing bag, rather than respawn it entirely.  Therefore, I don't believe this will be a significant issue.  We can still add in a pop-up warning that you're loading in an old version, but I don't think further action is necessary on this front.

Buhallin:
> almost all other instances I can think of you are specifically instructed to track in the campaign log (including Harbinger damage and TSK keys, and others like EOTE expedition team)
One of the nice things about rolling saves is you don't have to record them in the campaign log, and can just leave things out on the table.  I don't know how many players do that, but it's what my group does and I'd be very surprised if it's not a common practice.  This will be especially bad if you assume players are tracking it in the campaign log but they aren't.

I'm not sure I have a good solution to this yet, still thinking on it, but it's something I think is important to at least try and address.

Charon-Coder:
You make some good arguments.  That being said, I would imagine players are roughly split down the middle (continuing save vs fresh save, track on table vs track on log, etc.).  To that end, while I obviously want this tool to be as generally applicable as possible, I feel like this tool exists largely as an alternative to the Cleanup Helper; so ideally, this catches use cases the Cleanup Helper doesn't, and vice versa.  In other words, I feel like this tool exists largely to facilitate those using fresh saves, while Cleanup Helper exists largely to help those with continuing saves.  While I would absolutely love to have a tool that enables players to 'convert' to fresh saves instead of rolling saves, I think covering all the edge cases for how people track things on a table are just not feasible. Even for the sake of my greatly desired use case of importing/exporting to a new SCED version mid-campaign, I think the edge cases would be simply too numerous in the way you describe them.

My current opinion is leaning more towards just being very clear about communicating what the tool is and isn't able of tracking. (But obv still eager to hear you all's thoughts and opinions/continue the conversation)

src/accessories/CampaignImporterExporter.ttslua
Entrox-Licher: Wow this made a huge difference, thanks for the suggestion!

src/arkhamdb/DeckImporterUi.ttslua
Entrox-Licher: Totally agree the key-value table will be more legible, but I can't get rid of the 4 lines of code you mention because the `editInput` function does not trigger the input field's callback function (`colorDeckChanged` in this case), so the script will fail to load a deck because it isn't tracking the ID properly.  I could abstract that functionality out into a separate function if you think it would be clearer though.

Chr1Z93: Ah that's a good point. Then just keep those lines, but I'd still turn it into a key-value table

src/accessories/CampaignImporterExporter.ttslua
Entrox-Licher: Good suggestion, this is the solution I ended up going with!

src/core/Global.ttslua
Entrox-Licher: Yeah it was a some weirdness due to the fact those two are tracked as tables rather than booleans in the `optionPanel` table, so I added a simple check for it and it seems to be golden now.

Chr1Z93 (commented):
Looks good to me (at least I couldn't spot any blockers). Would keep this open until we have the other PR with the function for chaos token returning merged.

objects/CampaignImporterExporter.334ee3.json
```diff
@@ -0,0 +1,57 @@
...
+      "Stretch": true,
+      "Thickness": 0.2,
+      "Type": 0
+    },
+    "ImageScalar": 1,
+    "ImageSecondaryURL": "",
+    "ImageURL": "http://cloud-3.steamusercontent.com/ugc/254843371583173230/BECDC34EB4D2C8C5F9F9933C97085F82A2F21AE3/",
+    "WidthScale": 0
+  },
+  "Description": "Saves the state of the table to enable loading the campaign into a new save and keep all current progress.\n\nThis tool will track which campaign you're playing, the entire contents of your campaign log, the contents of your chaos bag, update your health/sanity according to trauma, your ArkhamDB deck IDs, the number of investigators, the page of your campaign guide, and any options you have selected in the options panel.",
```
Buhallin, line 25: Unless I'm missing something with the tagging, this is only going to work for official campaigns and not anything fanmade, correct?  Should make this very clear in the description, and possibly enhance various "not found" error messages with that.

src/core/OptionPanelApi.ttslua
```diff
@@ -0,0 +1,16 @@
...
+  local OptionPanelApi = {}
+
+  -- loads saved options
+  ---@param options Table New options table
+  OptionPanelApi.loadSettings = function(options)
+    return Global.call("loadSettings", options)
+  end
+
+  -- returns option panel table
+  OptionPanelApi.getOptionPanel = function()
```
Buhallin, line 11: Let's just make this getOptions() or getPreferences().  getOptionPanel() sounds like you're retrieving the UI element itself

objects/CampaignImporterExporter.334ee3.json
Entrox-Licher: Well, this will work for a custom campaign provided they properly tag all their components (box, log, guide), and they have the campaign box on the table when the game is imported.  I've added that to the description

src/accessories/CampaignImporterExporter.ttslua
```diff
@@ -0,0 +1,291 @@
...
+local deckImporterApi
+
+local campaignBoxGUID
+
+function onLoad(save_state)
+  chaosBagApi = require("chaosbag/ChaosBagApi")
+  playAreaAPI = require("core/PlayAreaApi")
+  deckImporterApi = require("arkhamdb/DeckImporterApi")
+  optionPanelApi = require("core/OptionPanelApi")
+  blessCurseApi = require("chaosbag/BlessCurseManagerApi")
```
Chr1Z93, line 61: I'm not sure if this actually makes a difference functionally or performance-wise, but let's move these out of `onLoad()` (e.g. directly to line 49-52), since that's the syntax we've used everywhere else.

src/accessories/CampaignImporterExporter.ttslua
```diff
@@ -0,0 +1,291 @@
...
+      function()
+        if #campaignBox.getObjects() > 0 then
+          placeCampaignFromToken(importData)
+        else
+          createCampaignFromToken(importData)
+        end
+      end,
+      function()
+        local obj = getObjectFromGUID(campaignBoxGUID)
+        if  obj == nil then
```
Chr1Z93, line 116: Nitpick: there is a double space after `if` (instead of a single one)

src/accessories/CampaignImporterExporter.ttslua
```diff
@@ -0,0 +1,272 @@
...
+  )
+  broadcastToAll("Campaign successfully imported!", Color.Green)
+end
+
+
+-- Creates a campaign token with save data encoded into GM Notes based on the current state of the table
+function createCampaignToken(_, _, _)
+  local campaignBoxGUID = ""
+  -- find active campaign
+  for _, obj in ipairs(getObjectsWithTag("CampaignBox")) do
```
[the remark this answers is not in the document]
Chr1Z93: @Entrox-Licher Can you explain how you solved this? I don't really understand how the `campaignBoxGUID` works 😅

src/accessories/CampaignImporterExporter.ttslua
```diff
@@ -0,0 +1,291 @@
...
+  broadcastToAll("Campaign successfully exported! Save coin object to import on a fresh save", Color.Green)
+end
+
+
+-- helper functions
+
+function findCampaignLog()
+  if getObjectsWithTag("CampaignLog") then
+    if #getObjectsWithTag("CampaignLog") == 1 then
+    return getObjectsWithTag("CampaignLog")[1]
```
Chr1Z93, line 250: Instead of calling `getObjectsWithTag("CampaignLog")` 3x, assign it to a variable and use it for the other two occasions.
Also I don't think you need the outer `if`.

Same principle applies to the campaign guide below. Might actually be worth to turn these into a function that takes the searchtag as parameter to avoid the repetition.

src/accessories/CampaignImporterExporter.ttslua
```diff
@@ -0,0 +1,291 @@
...
+  for i, guid in ipairs(DAMAGE_HORROR_GUIDS) do
+    local TOKEN = getObjectFromGUID(guid)
+  
+    if TOKEN ~= nil then
+    TOKEN.call("updateVal", tableOfNewValues[i])
+    else
+    printToAll(": No. " .. i .. " could not be found.", "Yellow")
+    end
+  end
+  end
```
Chr1Z93, line 291: Indentation seems off for these lines

src/accessories/CampaignImporterExporter.ttslua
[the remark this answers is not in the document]
Buhallin: I'd probably still check to see if you hit two, and abort the process with a message if you do.  If someone just cleans up the old campaign you could easily have two empty boxes. It should be straightforward to get the names and print them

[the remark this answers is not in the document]
Entrox-Licher: (Lines 182-196)
Basically, it will iterate through all the objects that have the CampaignBox tag.  This will return all the campaign boxes, both the downloaded and not.  On each one, it will check to make sure it's downloaded (i.e. is now a Bag) and that it's placed on the table (i.e. empty Bag).  If both of those things are true, it will update the `campaignBoxGUID`,  However, if there is already a value set in that variable, it means more than one has been detected, so the error on 187 is thrown, and the process is aborted.

Pretty sure this was already functioning correctly with my previous push?  Regardless, I still think it is an extreme edge case; it would require a user manually deleting all the scenario boxes instead of recalling (since cleanup helper doesn't touch scenario boxes), and then placing a new campaign.  Still, I agree it's good to account for it and not leave it ambiguous. Makes the code neater and more predictable.

Entrox-Licher:
Screenshot 2023-08-03 103520

BootleggerFinn (approved):
Looks great to me! What a fantastic feature!

src/accessories/CampaignImporterExporter.ttslua
```diff
@@ -0,0 +1,292 @@
...
+      end,
+      function()
+        findCampaignGuide().Book.setPage(importData["guide"])
+        return findCampaignGuide().Book.getPage() == importData["guide"]
+      end,
+      1,
+      function()
+        --log("Campaign Guide import failed!")
+      end
+    )
```
Chr1Z93, line 167: I'd split these function out of the Wait statement into separate functions for readability

src/accessories/CampaignImporterExporter.ttslua
```diff
@@ -0,0 +1,292 @@
...
+      function()
+        --log("Campaign Guide import successful!")
+      end,
+      function()
+        findCampaignGuide().Book.setPage(importData["guide"])
+        return findCampaignGuide().Book.getPage() == importData["guide"]
+      end,
+      1,
+      function()
+        --log("Campaign Guide import failed!")
```
Chr1Z93, line 165: Seems like this should either be un-commented or removed

src/accessories/CampaignImporterExporter.ttslua
```diff
@@ -0,0 +1,292 @@
...
+  if importData["decks"] then
+    deckImporterApi.setUiState(importData["decks"])
+  end
+  --set investigator count
+  playAreaApi.setInvestigatorCount(importData["clueCount"])
+  --set campaign guide page
+  if findCampaignGuide() then
+    Wait.condition(
+      function()
+        --log("Campaign Guide import successful!")
```
Chr1Z93, line 157: Seems like this should either be un-commented or removed

src/accessories/CampaignImporterExporter.ttslua
```diff
@@ -0,0 +1,292 @@
...
+    updateCounters(importData["trauma"])
+  end
+  --populate ArkhamDB deck IDs
+  if importData["decks"] then
+    deckImporterApi.setUiState(importData["decks"])
+  end
+  --set investigator count
+  playAreaApi.setInvestigatorCount(importData["clueCount"])
+  --set campaign guide page
+  if findCampaignGuide() then
```
Chr1Z93, line 154: I'd save the return value from `findCampaignGuide` to a variable, so you don't have to call it multiple times

src/core/PlayAreaApi.ttslua
```diff
@@ -77,5 +85,13 @@ do
     return getObjectFromGUID(PLAY_AREA_GUID).call("isInPlayArea", object)
   end
 
+  PlayAreaApi.getSurface = function()
+    return getObjectFromGUID(PLAY_AREA_GUID).getCustomObject().image
+  end
+
+  PlayAreaApi.updateSurface = function(url)
+    return getObjectFromGUID(IMAGE_SWAPPER).call("updateSurface", url)
+  end
```
Chr1Z93, line 94: Turn the image updating into a single function

src/accessories/CampaignImporterExporter.ttslua
Entrox-Licher: I'd rather keep these anonymous, but I did add some comments to improve legibility

src/core/PlayAreaApi.ttslua
Entrox-Licher: Agreed, but will be handled in the API Updates PR

Chr1Z93 (approved):

PR #303, opened 2023-06-12
added language selection to option panel

Chr1Z93:
This adds a language selection to the option panel:
image

This will point all download requests to the respective subfolder in the loadable-objects repo.

Implementation to automatically update the player cards bag is not yet included (@Buhallin).


lriuui0x0:
Maybe it's a good time to pack the Chinese font as part of this PR as well.

Chr1Z93:
> Maybe it's a good time to pack the Chinese font as part of this PR as well.

That is already merged with https://github.com/argonui/SCED/pull/299

lriuui0x0:
Ok since the font is already merged, we probably should set the splash screen font when the language is switched to Chinese.

Chr1Z93:
> Ok since the font is already merged, we probably should set the splash screen font when the language is switched to Chinese.

Good point. I'll keep that to a separate PR once we have sorted this out. Am I right in the assumption that we only need to do this for chinese and can keep the regular font for other languages?

lriuui0x0:
> Good point. I'll keep that to a separate PR once we have sorted this out. Am I right with the assumption that we only need to do this for chinese and can keep the regular font for other languages?

Yes, we should keep Teutonic font since that's the font used for most other languages. We probably need separate fonts for things like Korean and Russian, but that's outside of my knowledge.

Buhallin:
I tried to look through this, and there's just way too much going on.  Option panel.  Content move and rename.  Apparently also decomposition and assembly tests leading to a bunch of unrelated value rounding in the objects.

Can we please split this out and get a new PR with just the option panel at least?  The files are almost entirely unrelated so outside of a few URL tweaks in Global you could start a new branch and copy just the 2-3 files which relate to the options panel.  Ignore the addition of the language to the URL, that's easy to add later as part of the migration to the new structure.

Chr1Z93:
> I tried to look through this, and there's just way too much going on. Option panel. Content move and rename. Apparently also random position cleanup? This is at least 3 different PRs, and probably 4.
> 
> Can we please split this out and get a new PR with just the option panel at least? The files are almost entirely unrelated so outside of a few URL tweaks in Global you could start a new branch and copy just the 2-3 files which relate to the options panel. Ignore the addition of the language to the URL, that's easy to add later as part of the migration to the new structure.

Split PR can be found here: https://github.com/argonui/SCED/pull/319

argonui (changes requested):
Whole lot of changes that have nothing to do with language. Please pull those out

Chr1Z93:
> Whole lot of changes that have nothing to do with language. Please pull those out

Fully agreed. This is very outdated and not at all intended to be merged in this state. I'll close this once we have closure on whether we will include translations or not.

PR #304, opened 2023-06-12
Fix ArkhamDB id for parallel cards

lriuui0x0:


Chr1Z93 (changes requested):
If we update these IDs, the deck importer needs to be updated

lriuui0x0:
Thanks for the explanation! @Chr1Z93 @Buhallin

PR #306, opened 2023-06-13
Customs refactor

Buhallin:
Large refactoring of customizable card implementation.  This builds on pr/285 (which it replaces):
- Moves customizable card to require() structure with shared code (thanks Chr1Z!)
- Unifies UI definition into a single table
- Makes input (text) fields associate with the row
- Separates UI and data into separate fields
- Enhance button clicks to select up to the clicked column
- Updates the Deck Importer to present the new structure on load


Chr1Z93 (approved):
Great update, looked through and tested ingame

PR #307, opened 2023-06-13
Upkeep: ready face-down cards, removed legacy code

Chr1Z93:
Face-down cards will now ready and legacy code for the old metadata has been removed.

### DIFF
--- a/src/playermat/Playmat.ttslua
+++ b/src/playermat/Playmat.ttslua
@@ -245,32 +245,26 @@ function doUpkeep(_, clickedByColor, isRightClick)
 
   -- unexhaust cards in play zone, flip action tokens and find forcedLearning
   local forcedLearning = false
+  local rot = self.getRotation()
   for _, v in ipairs(searchAroundSelf()) do
     local obj = v.hit_object
     if obj.getDescription() == "Action Token" and obj.is_face_down then
       obj.flip()
-    elseif obj.tag == "Card" and not obj.is_face_down and not inArea(self.positionToLocal(obj.getPosition()), INVESTIGATOR_AREA) then
+    elseif obj.tag == "Card" and not inArea(self.positionToLocal(obj.getPosition()), INVESTIGATOR_AREA) then
       local cardMetadata = JSON.decode(obj.getGMNotes()) or {}
       if not doNotReady(obj) then
-        obj.setRotation(self.getRotation())
+        -- set correct rotation for face-down cards
+        rot.z = obj.is_face_down and 180 or 0
+        obj.setRotation(rot)
       end
       if cardMetadata.id == "08031" then
         forcedLearning = true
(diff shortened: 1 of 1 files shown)


src/playermat/Playmat.ttslua
```diff
@@ -245,32 +245,26 @@ function doUpkeep(_, clickedByColor, isRightClick)
...
   for _, v in ipairs(searchAroundSelf()) do
     local obj = v.hit_object
     if obj.getDescription() == "Action Token" and obj.is_face_down then
       obj.flip()
-    elseif obj.tag == "Card" and not obj.is_face_down and not inArea(self.positionToLocal(obj.getPosition()), INVESTIGATOR_AREA) then
+    elseif obj.tag == "Card" and not inArea(self.positionToLocal(obj.getPosition()), INVESTIGATOR_AREA) then
       local cardMetadata = JSON.decode(obj.getGMNotes()) or {}
       if not doNotReady(obj) then
-        obj.setRotation(self.getRotation())
+        -- set correct rotation for face-down cards
```
Buhallin, line 256: This does work, but I have no idea why it works.  Generally, comments should explain why you're doing something rather than just what, since that can be seen from the code itself.  In this case, what changes about the coordinate space that makes the z-axis invert?

src/playermat/Playmat.ttslua
```diff
@@ -245,32 +245,26 @@ function doUpkeep(_, clickedByColor, isRightClick)
...
-        -- a single entry or an array.  TODO: Clean this up when 2.4.0 has been out long
-        -- enough that saved decks don't have old data
-        if cardMetadata.uses.count ~= nil then
-          count = cardMetadata.uses.count
-          replenish = cardMetadata.uses.replenish
-        else
-          count = cardMetadata.uses[1].count
-          replenish = cardMetadata.uses[1].replenish
-        end
+        -- TODO: support for cards with multiple uses AND replenish (as of yet, no official card needs that)
```
Buhallin, line 265: As always, try and keep unrelated changes in separate PRs.

Buhallin (approved):

src/playermat/Playmat.ttslua
Chr1Z93: Since the `setRotation()` needs a full vector as input, the z-value needs to be set according to the cards orientation. I could have also used `getRotation()` on each card and manipulated the x- and y-value, but the would be more steps.

Buhallin: My point was that should go in the comment :)

Probably not worth changing now, but even if it's a little more code it's generally good practice not to mutate data unless you're actually intending to change it.  "Preserving the current Z value" is both clearer and safer than "Assume we know what the Z value should be."  I don't see any harm in this case though.

PR #310, opened 2023-06-14
Victory Display: trigger update from clue hotkey

Chr1Z93:
Adds the VictoryDisplayApi, which contains just one function for now. This function triggers an update and is used to keep the VD updated if clues are taken via hotkey (since they don't trigger an update through the existing events).

### DIFF
--- a/src/core/VictoryDisplayApi.ttslua
+++ b/src/core/VictoryDisplayApi.ttslua
@@ -1,0 +1,12 @@
+do
+  local VictoryDisplayApi = {}
+  local VD_GUID = "6ccd6d"
+
+  -- triggers an update of the Victory count
+  ---@param delay Number Delay in seconds after which the update call is executed
+  VictoryDisplayApi.update = function(delay)
+    getObjectFromGUID(VD_GUID).call("startUpdate", delay)
+  end
+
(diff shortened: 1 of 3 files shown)


Buhallin:
A simpler approach here would be to set the y-value on the moved cards a bit high, so they then fall and trigger the collision event

Chr1Z93:
> A simpler approach here would be to set the y-value on the moved cards a bit high, so they then fall and trigger the collision event

The issue is not that moved cards don't trigger the update.
Clues that are taken from locations via Hotkey need to trigger an update (because they could be the last clue from that location).

src/core/VictoryDisplayApi.ttslua
```diff
@@ -0,0 +1,12 @@
+do
```
Buhallin, line 1: For some reason it won't let me add the comment, but as long as you're creating the API here might as well add placeCard as well (GameKeyHandler:50).

(Note this is related since this can be considered a general victory display update/refactoring)

Buhallin (commented):
Ah, I see.  That's what I get for a quick morning read :)

Buhallin (approved):

PR #313, opened 2023-06-17
Update player mat handling of customizable cards

Buhallin:
- Fixes broken data retrieval to get the selected upgrades
- Moves "is this customization active" logic into the UpgradeSheetLibrary
- Adds logic to update customizable cards in play if a selection is made on the upgrade sheet
- Extends the area covered by searchAroundSelf() to extend farther in (and past) the set aside card areas
- Made searchAroundSelf() dynamic, removing the need for explicit position and areas


Chr1Z93 (approved):
LGTM

PR #316, opened 2023-06-18
Upkeep: keep state for replenishing cards

Chr1Z93:
This updates the "Upkeep" funtionality of the playmers to keep the state for resource tokens on cards with "Replenish" like "Runic Axe" or "Physical Training (4)".

### DIFF
--- a/src/core/token/TokenManager.ttslua
+++ b/src/core/token/TokenManager.ttslua
@@ -109,6 +109,7 @@ do
 
   -- stateIDs for the multi-stated resource tokens
   local stateTable = {
+    ["Resource"] = 1,
     ["Ammo"] = 2,
     ["Charge"] = 3,
     ["Secret"] = 4,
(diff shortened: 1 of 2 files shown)


src/playermat/Playmat.ttslua
```diff
@@ -546,17 +546,30 @@ function replenishTokens(card, count, replenish)
...
   local search = searchArea(cardPos, { 2.5, 0.5, 3.5 })
   local clickableResourceCounter = nil
   local foundTokens = 0
 
   for _, obj in ipairs(search) do
     local obj = obj.hit_object
-    if obj.getCustomObject().image == "http://cloud-3.steamusercontent.com/ugc/1758068501357192910/11DDDC7EF621320962FDCF3AE3211D5EDC3D1573/" then
+    local memo = obj.getMemo()
+
+    if (stateTable[memo] or 0) > 0 then
```
Buhallin, line 568: This is going to be pretty inconsistent if a card ends up with different token types, since it'll just take the last one found.  I think it'll be better to pass the type in, and spawn that explicitly?

src/core/token/TokenManager.ttslua
```diff
@@ -231,7 +231,7 @@ do
 
     -- this is used to load the correct state for additional resource tokens (e.g. "Ammo")
     local callback = nil
-    if tokenType == "resource" and stateID ~= nil then
+    if tokenType == "resource" and stateID ~= nil and stateID ~= 1 then
```
Buhallin, line 234: Up to you if it's worth changing, but looking at this again now I think it would have been better to pass the type string all the way to this level, then convert it to the state number just before spawning (or even in the callback method).  Doing it at a higher level (and taking the state number here) means that the conversion code and state translation table are duplicated, and they're currently done so inconsistently (compare Playmat:550 to TokenManager:111).

Approving, but I think this would be worth cleaning up before submit.

Buhallin (approved):

src/core/token/TokenManager.ttslua
Chr1Z93: Valid objection, I'll change that

Chr1Z93: @Buhallin I've updated it, but can't get rid of the table in the playmat since I need to check the memo of the tokens. Are you fine with merging it like this?

Chr1Z93: Finale update, moved the replenishing function to token manager to avoid repetion

Issue #315, opened 2023-06-18
Add Function to Check for Sealed Tokens


Entrox-Licher:
Add a function that can be called by Clean Up Helper or Campaign/Importer Exporter that will check for any cards with sealed tokens on them (and ideally release them back to the bag before continuing operations).  Where would the best place for this function be? (CardsThatSealTokens.ttslua?)

PR #317, opened 2023-06-18
Clean Up Helper: return sealed tokens to chaos bag

Chr1Z93:
- Adds a new function that returns chaos tokens that are on the table to the chaos bag. Intended usage if for the Clean Up Helper and Campaign Importer / Exporter. Closes https://github.com/argonui/SCED/issues/315
- uses the existing blesscursemanagerapi for the Clean Up Helper instead of a direct call
- minor update to bless / curse manager printing (yes, this is functionally unrelated... 🙈)

### DIFF
--- a/src/core/Global.ttslua
+++ b/src/core/Global.ttslua
@@ -646,11 +646,15 @@ function emptyChaosBag()
   end
 end
 
--- find all chaos tokens that are on the table to the chaos bag (e.g. sealed on cards)
+-- move all regular chaos tokens that are on the table to the chaos bag (e.g. sealed on cards)
 function findAllChaosTokens()
   local chaosbag = findChaosBag()
   for _, obj in ipairs(getObjects()) do
-    if tokenChecker.isChaosToken(obj) and not obj.hasTag("tempToken") then
+    local name = obj.getName()
+    if tokenChecker.isChaosToken(obj) and
+        not obj.hasTag("tempToken") and
+        name ~= "Bless" and
+        name ~= "Curse" then
       chaosbag.putObject(obj)
     end
   end
--- a/src/core/Global.ttslua
+++ b/src/core/Global.ttslua
@@ -646,8 +646,8 @@ function emptyChaosBag()
   end
 end
 
--- return chaos tokens that are on the table to the chaos bag (e.g. sealed on cards)
-function returnChaosTokens()
+-- find all chaos tokens that are on the table to the chaos bag (e.g. sealed on cards)
+function findAllChaosTokens()
   local chaosbag = findChaosBag()
   for _, obj in ipairs(getObjects()) do
     if tokenChecker.isChaosToken(obj) and not obj.hasTag("tempToken") then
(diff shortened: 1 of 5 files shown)


src/core/Global.ttslua
```diff
@@ -645,6 +646,16 @@ function emptyChaosBag()
   end
 end
 
+-- find all chaos tokens that are on the table to the chaos bag (e.g. sealed on cards)
+function findAllChaosTokens()
```
Buhallin, line 650: This may be a broader feature for the cleanup helper, but should it clean up Bless/Curse tokens since those are removed at the end of a scenario?

src/core/Global.ttslua
```diff
@@ -645,6 +646,16 @@ function emptyChaosBag()
...
+  local chaosbag = findChaosBag()
+  for _, obj in ipairs(getObjects()) do
+    if tokenChecker.isChaosToken(obj) and not obj.hasTag("tempToken") then
+      chaosbag.putObject(obj)
```
Buhallin, line 654: If this finds Bless/Curse and puts them in the bag, will it leave the B/C manager in an incorrect state?

src/core/Global.ttslua
Chr1Z93: Those are already handled (via a call to the bless/curse manager who has the GUIDs stored) before this function is called ;)

src/core/Global.ttslua
Chr1Z93: The bless/curse manager indexes these tokens `onLoad()`. But if this is called it would probably leave it in a wrong state 🤔

Chr1Z93: Updated it to exclude bless / curse tokens (since these are temporary)

src/core/Global.ttslua
```diff
@@ -645,6 +646,20 @@ function emptyChaosBag()
   end
 end
 
+-- move all regular chaos tokens that are on the table to the chaos bag (e.g. sealed on cards)
+function findAllChaosTokens()
```
Buhallin, line 650: The name here is unclear - find() implies no mutation, just a lookup.  Maybe returnAllChaosTokens()?

src/core/Global.ttslua
```diff
@@ -645,6 +646,20 @@ function emptyChaosBag()
...
+  local chaosbag = findChaosBag()
+  for _, obj in ipairs(getObjects()) do
```
Buhallin, line 652: If a user has other tokens on the table, this will put them in the bag.  I don't know about other groups, but when I have a campaign with a highly mutable bag through scenarios I tend to leave the extra copies on the table.

src/core/Global.ttslua
```diff
@@ -645,6 +646,20 @@ function emptyChaosBag()
...
+    local name = obj.getName()
+    if tokenChecker.isChaosToken(obj) and
+        not obj.hasTag("tempToken") and
+        name ~= "Bless" and
+        name ~= "Curse" then
+      chaosbag.putObject(obj)
```
Buhallin, line 658: Cards that seal tokens track those tokens locally.  Just pulling them will leave the state on those incorrect.

Buhallin (changes requested):

src/core/Global.ttslua
Chr1Z93: Limited this function to those sealed tokens

src/core/Global.ttslua
Chr1Z93: Good point. Now it will only return sealed tokens

src/core/Global.ttslua
Chr1Z93: Name is changed now

Buhallin (approved):

PR #318, opened 2023-06-19
Add camera_ignore tags and reduce X_BUFFER

Chr1Z93:
Purpose of this PR is to fix the Navigation Overlay zooming on the victory display / campaign log / table dividers since they are now getting hit by the playmat's `searchAroundSelf`.

This is a draft since I am not yet sure what the best solution to this issue is.

### DIFF
--- a/objects/TableDivider.612072.json
+++ b/objects/TableDivider.612072.json
@@ -29,6 +29,11 @@
   "Nickname": "Table Divider",
   "Snap": true,
   "Sticky": true,
+  "Tags": [
+    "CameraZoom_ignore",
+    "CleanUpHelper_ignore",
+    "displacement_excluded"
+  ],
   "Tooltip": false,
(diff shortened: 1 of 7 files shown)


Buhallin:
This would be the other impacts I was asking about with the bound search change 🤣 

Easiest answer here seems to be cutting the search area back down to avoid the other elements.

Chr1Z93:
> This would be the other impacts I was asking about with the bound search change 🤣
> 
> Easiest answer here seems to be cutting the search area back down to avoid the other elements.

Yeah, at that point I didn't think of the Navigation Overlay.

Reducing the search area is what I did. Additionally, I added the exclusion tags (for Navigation Overlay, CleanUp Helper and Displacement Helper) for good measure since the table elements should never be manipulated like that.

Buhallin (approved):

PR #319, opened 2023-06-21
Option Panel: added dropdown list to select language

Chr1Z93:
This is a partial split from https://github.com/argonui/SCED/pull/303 for better reviewability. Contains just the code for a dropdown menu in the option panel. The selected option does not do anything (for now).


src/core/Global.ttslua
```diff
@@ -46,6 +46,27 @@ local SOURCE_REPO = 'https://raw.githubusercontent.com/chr1z93/loadable-objects/
 local library, requestObj, modMeta, notificationVisible
 local acknowledgedUpgradeVersions = {}
 
+-- language selection
+local LANGUAGE_CODES = {
```
Buhallin, line 50: Should avoid correlated arrays.  Instead:
local LANGUAGES = {
  [1] = {
    code = "en",
    name = "English"
  },
...
}

objects/LuaScriptState.luascriptstate
```diff
@@ -1 +1 @@
-{"acknowledgedUpgradeVersions":[],"optionPanel":{"playAreaSnapTags":true,"showAttachmentHelper":false,"showChaosBagManager":false,"showCleanUpHelper":false,"showCustomPlaymatImages":false,"showCYOA":false,"showDisplacementTool":false,"showDrawButton":false,"showHandHelper":[],"showSearchAssistant":[],"showTitleSplash":true,"showTokenArranger":false,"useClueClickers":false,"useSnapTags":true}}
+{"acknowledgedUpgradeVersions":[],"optionPanel":{"cardLanguage":3,"playAreaSnapTags":true,"showAttachmentHelper":false,"showChaosBagManager":false,"showCleanUpHelper":false,"showCustomPlaymatImages":false,"showCYOA":false,"showDisplacementTool":false,"showDrawButton":false,"showHandHelper":[],"showSearchAssistant":[],"showTitleSplash":true,"showTokenArranger":false,"useClueClickers":false,"useSnapTags":true}}
```
Buhallin, line 1: Is this actually making the default German, or is it a weird 0-indexed case?

Regardless, storing an index to the array ties it directly to that array, so if we were to add or reorder languages it could break things.  Better to store the language code directly.

src/core/Global.ttslua
```diff
@@ -838,15 +859,23 @@ function onClick_toggleOption(_, id)
...
 end
 
+-- called by the language selection dropdown
+function languageSelected(_, selectedIndex, id)
+  optionPanel[id] = tonumber(selectedIndex)
+end
+
 -- sets the option panel to the correct state (corresponding to 'optionPanel')
 function updateOptionPanelState()
   for id, enabled in pairs(optionPanel) do
```
Buhallin, line 869: enabled has become disconnected as it describes a boolean but can actually be pretty much any type.  Consider a name change for clarity.  value or optionValue?

src/core/Global.ttslua
```diff
@@ -838,15 +859,23 @@ function onClick_toggleOption(_, id)
...
     if (type(enabled) == "boolean" and enabled)
         or (type(enabled) == "string" and enabled)
         or (type(enabled) == "table" and #enabled ~= 0) then
-      self.UI.setAttribute(id, "isOn", true)
+      UI.setAttribute(id, "isOn", true)
+    -- hard-coded updating of the language selection dropdown
+    elseif id == "cardLanguage" and type(enabled) == "number" then
```
Buhallin, line 875: For clarity in a conditional block it's usually better to but more specific branches first, and more general ones later.

xml/OptionPanel.xml
```diff
@@ -101,6 +110,30 @@
...
+              <VerticalLayout class="text-column">
+                <Text class="option-header">Card language</Text>
+                <Text class="description">Downloading a supported campaign or importing a deck will use this language for the cards.</Text>
+              </VerticalLayout>
+            </Cell>
+            <Cell class="option-dropdown">
+              <Panel padding="0 17 13 13">
+                <Dropdown id="cardLanguage"
+                  onValueChanged="languageSelected(selectedIndex)">
+                  <Option>简体中文</Option>
```
Buhallin, line 125: Can these options can be created dynamically from the language name list?

objects/LuaScriptState.luascriptstate
Chr1Z93: It's 0-indexed, default is English. Will update to store language directly

xml/OptionPanel.xml
Chr1Z93: I tried it, but after hitting my head against a wall for an hour, I give up. I assume dropdown menu works differently than other XML.

src/core/Global.ttslua
```diff
@@ -838,15 +848,33 @@ function onClick_toggleOption(_, id)
...
 end
 
+-- called by the language selection dropdown
+function languageSelected(_, selectedIndex, id)
+  optionPanel[id] = LANGUAGES[tonumber(selectedIndex) + 1].name
+end
+
+function returnLanguageId(name)
+  for index, tbl in ipairs(LANGUAGES) do
+    for key, value in pairs(tbl) do
```
Buhallin, line 858: You know what the elements in the table are, seems like you could just check name and return code rather than having the inner loop?  Or am I missing something (it's late)

objects/LuaScriptState.luascriptstate
```diff
@@ -1 +1 @@
-{"acknowledgedUpgradeVersions":[],"optionPanel":{"playAreaSnapTags":true,"showAttachmentHelper":false,"showChaosBagManager":false,"showCleanUpHelper":false,"showCustomPlaymatImages":false,"showCYOA":false,"showDisplacementTool":false,"showDrawButton":false,"showHandHelper":[],"showSearchAssistant":[],"showTitleSplash":true,"showTokenArranger":false,"useClueClickers":false,"useSnapTags":true}}
+{"acknowledgedUpgradeVersions":[],"optionPanel":{"cardLanguage":"English","playAreaSnapTags":true,"showAttachmentHelper":false,"showChaosBagManager":false,"showCleanUpHelper":false,"showCustomPlaymatImages":false,"showCYOA":false,"showDisplacementTool":false,"showDrawButton":false,"showHandHelper":[],"showSearchAssistant":[],"showTitleSplash":true,"showTokenArranger":false,"useClueClickers":false,"useSnapTags":true}}
```
Buhallin, line 1: This should save the code, rather than the name.  It's shorter, harder to get wrong if we have to update it manually, and doesn't require unicode to store.

xml/OptionPanel.xml
```diff
@@ -101,6 +110,30 @@
             </Cell>
           </Row>
 
+          <!-- Option: card language -->
+          <Row class="option-text">
+            <Cell class="option-dropdowntext">
+              <VerticalLayout class="text-column">
+                <Text class="option-header">Card language</Text>
+                <Text class="description">Downloading a supported campaign or importing a deck will use this language for the cards.</Text>
```
Buhallin, line 118: Might be worth pointing out that cards already loaded won't be affected, but not sure how much room there is to work with

src/core/Global.ttslua
Chr1Z93: Good catch. Updated it

xml/OptionPanel.xml
Chr1Z93: I can't think of a concise way to convey this. We only have two lines 😓

BootleggerFinn (approved):

PR #323, opened 2023-06-22
Upkeep: rotate cards to the next multiple of 90° towards 0°

Chr1Z93:
The "Upkeep" function will now rotate cards to the next closest multiple of 90° towards 0° (local rotation to the playmat).
This can be used to mark cards as "do not ready during the next upkeep phase" by simply rotating them to something between 90° and 270°.

Example of Rotation Change in Upkeep
before -> after
45 -> 0
90 -> 0
135 -> 90
180 -> 90 (had to apply some precedence here and chose 90 over 270)
225 -> 270
270 -> 0

### DIFF
--- a/src/playermat/Playmat.ttslua
+++ b/src/playermat/Playmat.ttslua
@@ -280,9 +280,23 @@ function doUpkeep(_, clickedByColor, isRightClick)
     elseif obj.tag == "Card" and not inArea(self.positionToLocal(obj.getPosition()), INVESTIGATOR_AREA) then
       local cardMetadata = JSON.decode(obj.getGMNotes()) or {}
       if not doNotReady(obj) then
+        local cardRotation = round(obj.getRotation().y, 0) - rot.y
+        local yRotDiff = 0
+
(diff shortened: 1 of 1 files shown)


Buhallin:
I'm a little concerned about weird behavior here if a user has their rotation set to something other than 90.  The double-tap concept is fine as long as we can maintain it, but if we rotate in 90-degree intervals it's going not going to mesh well with someone who uses 45.

Chr1Z93:
> I'm a little concerned about weird behavior here if a user has their rotation set to something other than 90. The double-tap concept is fine as long as we can maintain it, but if we rotate in 90-degree intervals it's going not going to mesh well with someone who uses 45.

I agree that this will feel a bit out-of-place for the people that rotate less than 90°. Nevertheless I think it would be a good addition since it is fully optional. During regular gameplay cards won't be rotate over 90° so it will go unnoticed until someone goes out of their way to use this.

BootleggerFinn (approved):
Let's put this change in but be open to feedback if it winds up causing problems for or confusing people

PR #325, opened 2023-06-26
Addition of downloadable content

Chr1Z93:
This adds the downloadable content (that is currently hosted on https://github.com/Chr1Z93/loadable-objects) to this repository. Only initially the `generated` folder will be manually edited. Once the `decomposed` folder is set-up, the `generated` folder should not be edited manually.

Here is an overview of the file structure:
image


content/decomposed/campaigns/placeholder.txt
```diff
@@ -0,0 +1 @@
+<to be removed once the automation is in place>
```
Buhallin, line 1: I don't think we need these.  The new directories can be added as part of getting the decomposition running

Buhallin (approved):

content/decomposed/campaigns/placeholder.txt
Chr1Z93: Alright, removed them

PR #326, opened 2023-06-26
Content-Downloading: change links to new repo

Chr1Z93:
This updates the content-downloading to point to this repo. For official campaigns there is now a dynamically created URL instead of having them in the gm_notes.

The removal of the existing gm_notes is not included in this PR for easier reviewability.
Additionally, there will be the need for a new tag (or memo) for campaign boxes to give the DownloadBox file the correct size for the download buttons).


src/core/DownloadBox.ttslua
```diff
@@ -15,27 +13,20 @@ function onLoad()
...
-    buttonParameters.width = 3400
-    buttonParameters.font_size = 700
   end
 
   self.createButton(buttonParameters)
 end
 
 function buttonClick_download()
-  Global.call('placeholder_download', { url = self.getGMNotes(), replace = self.guid })
+  Global.call('placeholder_download', { name = self.getName(), url = self.getGMNotes(), replace = self.guid })
```
Buhallin, line 31: Were you planning to add the URL conversion from the name in this one?  Or even better, the library is keyed by name so there shouldn't be any need to provide a URL at all - just look it up from there based on the name.

src/core/Global.ttslua
```diff
@@ -42,7 +42,7 @@ local navigationOverlayApi = require("core/NavigationOverlayApi")
 
 -- online functionality related variables
 local MOD_VERSION = "3.1.0"
-local SOURCE_REPO = 'https://raw.githubusercontent.com/chr1z93/loadable-objects/main'
+local SOURCE_REPO = 'https://raw.githubusercontent.com/argonui/SCED/main'
```
Buhallin, line 45: I don't love this pointing to the top level of the repo.  As mentioned in the other PR, let's make this look to content

src/core/Global.ttslua
```diff
@@ -650,14 +650,34 @@ end
...
+function onClick_select(_, params)
+  if type(params) == "string" then
+    params = JSON.decode(urldecode(params))
+  end
+
+  local url = SOURCE_REPO .. '/content/'
+  if params.url ~= nil and params.url ~= "" then
+    -- TODO: change local folder based on selection in option panel
+    -- TODO: better detection of fancreations
+    if string.match(params.url, "............") == "fancreations" then
```
Buhallin, line 667: This change should be an update in the library.json file, not done in code.  Especially with version mismatch concerns it will be very hard to change.

src/core/Global.ttslua
Chr1Z93: Updated

src/core/DownloadBox.ttslua
Chr1Z93: Updated it to use the library for URLs of community content

content/library.json
Chr1Z93, line 1: - removed file extension
- removed "fancreations/" prefix

Chr1Z93, line 1: - removed file extension
- removed "fancreations/" prefix

Chr1Z93:
Turned this into a draft to avoid accidental merging until we have changed the URL to the new repo once it is set up

PR #327, opened 2023-06-26
move of modversion file

Chr1Z93:
This copies the existing modversion file (which holds metadata about the mod for the update notification) to this repo.


Buhallin:
I don't think we want this at the top level.  Maybe a new version directory?  We could use that for the player cards version as well.

Buhallin:
Though possibly in content/versions.  Looking at the link update, it doesn't feel right to point that at the top level of the mod (which inherently points to src, xml, objects, etc).

Chr1Z93:
> Though possibly in content/versions. Looking at the link update, it doesn't feel right to point that at the top level of the mod (which inherently points to src, xml, objects, etc).

I think we should keep this at the same directory as the library file since both of these contain metadata about the repository.

Buhallin (approved):

PR #329, opened 2023-07-20
New bounty and evidence tokens

Chr1Z93:
added the two new token types (to the token source and the 3 resource bags):
image


Tikatoy (approved):
Looks nice!

PR #331, opened 2023-07-20
Living Ink bugfix

Chr1Z93:
This adds an empty entry to the customizations table for Living Ink to account for the empty row with the skill selection (without this the additional charges effect is falsely applied to the 2nd row instead of the 3rd)


BootleggerFinn (approved):
Thanks for the fix!

PR #332, opened 2023-07-20
Bugfix for unprovided subtype (token spawning)

Chr1Z93:
This fixes an error when the subtype for a token is not provided.

VERY IMPORTANT since without this clue-spawning from custom data helper is broken!


BootleggerFinn (approved):
Thanks for the fix for this too!

PR #335, opened 2023-07-21
added numerical display and update button to cards with sealing

Chr1Z93:
Still WIP, unsure about the position / visual of the button

image


argonui:
Can you add more context to why this change is needed?

Chr1Z93:
> Can you add more context to why this change is needed?

Sure!

There are multiple cards that seal 5 or more tokens (Nephthys, Holy Spear, Rite of Sanctification etc.). For these cards you don't really know how many tokens exactly you have unless you count them manually (which can be a bit tedious). So this displays was meant as quick reference. The "click to update" ability is not really needed and could be removed if we don't value that.

Chr1Z93:
@argonui I think I have a better solution: A tooltip that is displayed when hovering the card.

argonui:
That sounds better. Balancing new features with existing players comforts is a tough one

Chr1Z93:
I'll close this for now - I don't think there is demand for this

PR #340, opened 2023-07-30
Clean Up Helper: updated trauma getting

Chr1Z93:
This change utilizes the "returnTrauma" function that is included in the campaign logs instead of directly accessing their data (results in the same behaviour, but allows support for unconventional campaign logs like Dark Matter with its Memory stat)

### DIFF
--- a/src/accessories/CleanUpHelper.ttslua
+++ b/src/accessories/CleanUpHelper.ttslua
@@ -254,15 +254,19 @@ function getTrauma()
     printToAll("Default values for health and sanity loaded.", "Yellow")
     return
   end
+  campaignLog = campaignLog.hit_object
+
+  -- set campaign log to first state
+  local stateId = campaignLog.getStateId()
+  if stateId ~= 1 then
+    campaignLog.setState(1)
+  end
 
   -- get data from campaign log if possible
-  local counterData = campaignLog.hit_object.getVar("ref_buttonData")
-  if counterData ~= nil then
+  local trauma = campaignLog.getVar("returnTrauma")
+  if trauma ~= nil then
     printToAll("Trauma values found in campaign log!", "Green")
-    for i = 1, 10, 3 do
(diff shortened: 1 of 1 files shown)


src/accessories/CleanUpHelper.ttslua
```diff
@@ -248,21 +248,32 @@ function getTrauma()
...
-  local counterData = campaignLog.hit_object.getVar("ref_buttonData")
-  if counterData ~= nil then
+  -- set campaign log to first state
+  local stateId = campaignLog.getStateId()
+  if stateId ~= 1 then
+    campaignLog = campaignLog.setState(1)
+  end
+
+  -- small delay to account for potential state change
+  Wait.time(loadTrauma, 0.1)
```
argonui, line 266: Add units either in function name, comment, or set .1 as a variable

src/accessories/CleanUpHelper.ttslua
```diff
@@ -248,21 +248,32 @@ function getTrauma()
   end
 
   -- get campaign log
-  local campaignLog = findObjects(6)[1]
+  campaignLog = findObjects(6)[1]
```
argonui, line 251: What are these constants?

argonui (changes requested):
How would I test this?

Chr1Z93:
> How would I test this?

1) Load any official campaign (and place the campaign log)
2) Set trauma values (Player 1 is White etc.)
3) Optionally swap campaign log to second page
4) Spawn CleanUp Helper from option panel
5) Hit the clean up button
6) Check if health/sanity match what you set in 2)

src/accessories/CleanUpHelper.ttslua
Chr1Z93: - `campaignLog` is the object reference to the campaign log
- `findObjects` is a function that uses a cast to find an object (the first object from that search being the campaign log in this case)
- the 6 stands for the type of search to perform

BootleggerFinn (approved):

PR #341, opened 2023-08-08
API Updates

Entrox-Licher:
Adding support across the codebase to utilize new and existing APIs following the API pattern.

The code here is reliant on the ChaosBagApi implemented by the Campaign Importer/Exporter, so will need that to be merged in first.


src/core/GameKeyHandler.ttslua
```diff
@@ -98,14 +99,18 @@ function takeClueFromLocation(playerColor, hoveredObject)
...
+  local matColor = playmatApi.getMatColor(playerColor)
+  local pos = nil
+  if clickableClues then
+    pos = {x = 0.49, y = 2.66, z = 0.00}
+    playmatApi.updateClueClicker(playerColor, playmatApi.getClueCount(clickableClues, playerColor) + 1)
+  else
+    pos = playmatApi.transformLocalPosition({x = -1.12, y = 0.05, z = 0.7}, matColor)
   end
+  
+  local rot = playmatApi.returnRotation(matColor)
```
Chr1Z93, line 113: I think you need to add a replacement effect to the physical move of the clue token - see LOC 116-121 in your code (like moving it to a trash can instead).

src/core/GameKeyHandler.ttslua
```diff
@@ -98,14 +99,18 @@ function takeClueFromLocation(playerColor, hoveredObject)
...
 
   -- check if found clue is a stack or single token
```
Chr1Z93, line 115: The following lines need handling for clickable clues

src/accessories/CleanUpHelper.ttslua
```diff
@@ -8,6 +8,7 @@ local soundCubeApi              = require("core/SoundCubeApi")
 local playmatApi                = require("playermat/PlaymatApi")
 local blessCurseManagerApi      = require("chaosbag/BlessCurseManagerApi")
 local chaosBagApi               = require("chaosbag/ChaosBagApi")
+local playAreaAPI               = require("core/PlayAreaApi")
```
Chr1Z93, line 11: "API" should be "Api" for consistent naming

src/core/PlayAreaApi.ttslua
```diff
@@ -83,5 +84,30 @@ do
...
+    return getObjectFromGUID(PLAY_AREA_GUID).getCustomObject()
+  end
+
+  PlayAreaApi.setCustomObject = function(customInfo)
+    return getObjectFromGUID(PLAY_AREA_GUID).setCustomObject(customInfo)
+  end
+
+  PlayAreaApi.reload = function()
+    return getObjectFromGUID(PLAY_AREA_GUID).reload()
+  end
```
Chr1Z93, line 108: Why do we need these as functions in the API?

src/core/GameKeyHandler.ttslua
Chr1Z93: True, I missed that you already implemented that

src/core/PlayAreaApi.ttslua
Chr1Z93: Usually we add an Api if we need to access resources (variables or functions) of another object. So for this case I'd say it would be better if you just add a new function that takes the new image as parameter instead of 3 functions that are useless on their own.

src/core/GameKeyHandler.ttslua
```diff
@@ -98,14 +99,18 @@ function takeClueFromLocation(playerColor, hoveredObject)
...
   end
 
+  local clickableClues = optionPanelApi.getOptions()["useClueClickers"]
   local playerName = Player[playerColor].steam_name
-  local matColor = playmatApi.getMatColor(playerColor) 
-  local pos = playmatApi.transformLocalPosition({x = -1.12, y = 0.05, z = 0.7}, matColor)
-  local rot = playmatApi.returnRotation(matColor)
-
-  if cardName == "" or cardName == nil then
-    cardName = "nameless card"
```
Chr1Z93, line 107: Why was this removed?

src/core/PlayArea.ttslua
```diff
@@ -504,20 +504,6 @@ function shiftContents(playerColor, direction)
...
-  return investigatorCounter.getVar("val")
-end
-
--- Updates the current value of the investigator counter from the playmat
----@param count Number of investigators to set on the counter
-function setInvestigatorCount(count)
-  local investigatorCounter = getObjectFromGUID("f182ee")
-  return investigatorCounter.call("updateVal", count)
-end
-
```
Chr1Z93, line 520: Can you confirm that you checked whether this is referenced by any other scripts?

src/core/PlayAreaApi.ttslua
```diff
@@ -2,19 +2,20 @@ do
   local PlayAreaApi = { }
 
   local PLAY_AREA_GUID = "721ba2"
+  local INVESTIGATOR_COUNTER_GUID = "f182ee"
 
   local IMAGE_SWAPPER = "b7b45b"
```
Chr1Z93, line 7: While you are at it, this variable should get a `_GUID` suffix too

src/core/PlayAreaSelector.ttslua
```diff
@@ -70,8 +71,7 @@ function none() end
 
 -- main function (can be called by other objects)
 function updateSurface(newURL)
-    local playArea = getObjectFromGUID("721ba2")
-    local customInfo = playArea.getCustomObject()
+    local customInfo = playAreaAPI.getCustomObject()
```
Chr1Z93, line 74: Instead of adding a single-use function to the Api, let's add the "updateSurface" functionality as an Api function

src/playermat/Playmat.ttslua
```diff
@@ -772,7 +774,7 @@ end
 ---------------------------------------------------------
 
 function drawChaosTokenButton(_, _, isRightClick)
-  Global.call("drawChaosToken", {self, DRAWN_CHAOS_TOKEN_OFFSET, isRightClick})
+  chaosBagApi.drawChaosToken({self, DRAWN_CHAOS_TOKEN_OFFSET, isRightClick})
```
Chr1Z93, line 777: Advantage of the Api files is that we don't have to restrict us to a single parameter (and thus make multiple ones into a single table). So'd split these into three separate parameters.

src/playercards/AllCardsBagApi.ttslua
Chr1Z93, line 1: While you can only pass a single parameter to a function call on another object, we don't have that limitation in the Api's itself. Thus I think it would be better if we split those up. Example:
```
before:
AllCardsBagApi.getCardsByName = function(params)
  return getObjectFromGUID(ALL_CARDS_BAG_GUID).call("getCardsByName", params)
end

after:
AllCardsBagApi.getCardsByName = function(name, exact)
  return getObjectFromGUID(ALL_CARDS_BAG_GUID).call("getCardsByName", {name = name, exact = exact)
end
```
Of course this also requires a change to the places that use the Api function.

src/arkhamdb/DeckImporterUi.ttslua
```diff
@@ -218,8 +220,7 @@ function loadDecks()
   -- Method in DeckImporterMain, visible due to inclusion
 
   -- TODO: Make this use the configuration ID for the all cards bag
```
Chr1Z93, line 222: can remove the "TODO" line at this point

Chr1Z93 (changes requested):
See separate comments

src/core/GameKeyHandler.ttslua
Entrox-Licher: Because it conflicts with the if statement a few lines down (LOC 123-127).  Of the two options, I thought the later one made more sense, since stating you received a clue from a 'nameless card' is not very useful, so you may as well just say nothing about the card at all.  I can switch it back if preferred though, but I still think we should have only one or the other for code clarity.

src/core/PlayArea.ttslua
Entrox-Licher: Yep!  It only existed for the API, so I just put it in there for clarity

src/playermat/Playmat.ttslua
Entrox-Licher: Yep, makes perfect sense.  I was trying to just make them behave the same as the original function for consistency, but I agree it's just much more usable to split them all up

src/playercards/AllCardsBagApi.ttslua
Entrox-Licher: Agreed!  My only concern is they become slightly inconsistent (i.e. when used internally they are referenced differently to when they are used externally, since the internal function still needs to only take a single parameter in order to make the call possible in the API at all).  That being said, since these functions are used in so many other places, and the difference is relatively minor, I don't have an issue with it

Entrox-Licher: Note: latest commit has implemented this change, however due to TTS being a little ridiculous with their `Attempt to perform operations with resources owned by different scripts` error, the `getCardById` function still needs to be a params table, even though it's only one parameter; it will trigger the aforementioned error otherwise, for reasons known only to The Creators

src/arkhamdb/ArkhamDb.ttslua
```diff
@@ -166,18 +166,20 @@ do
...
-      local weaknessId = allCardsBagApi.getRandomWeaknessId()
-      slots[weaknessId] = 1
-      slots[RANDOM_WEAKNESS_ID] = nil
-      internal.maybePrint("Random basic weakness added to deck", playerColor)
+    if randomWeaknessAmount ~= 0 then
+      for i=1, randomWeaknessAmount do
+        local weaknessId = allCardsBagApi.getRandomWeaknessId()
+        slots[weaknessId] = 1
+        slots[RANDOM_WEAKNESS_ID] = nil
+      end
```
Chr1Z93, line 181: Now that I read this again, I think this will result in too few weaknesses if the random weakness rolls the same one multiple times 😅
We're gonna need to check if the weaknessId was already added to the deck (`if slots[weaknessId] == 1`) and get a new one if it is already in there

Chr1Z93: Hm and I'm not certain whether this will take the weakness count into consideration when importing multiple decks. Theoretically, it should since they are load one by one, but I am not sure if there is enough delay for the allcardsbag to pick up on the weaknesses in other peoples decks (and also doesn't work properly if a unique weakness is in a deck that will be loaded later)

Chr1Z93: I'm probably going to target this in a separate PR since this one is already huge

Chr1Z93 (changes requested):

Chr1Z93 (approved):

PR #343, opened 2023-08-11
Tool promotion: Chaos Bag Manager and Token Arranger

Chr1Z93:
- Chaos Bag Manager and Token Arranger are now placed on the table by default to declutter the option panel
- The Token Arranger starts without placed tokens, but will be updated automatically when adding/removing tokens or loading a scenario
- The temporary tokens are now locked to avoid confusion for new players
- A rounding error for the cumulative percentages was fixed

Table Preview (relevant part):
image


argonui:
Clutter in an options menu is not a bad thing. I am much more nervous about cluttering the board. 

What resources exist to explain the tools? 

Can you link a screenshot of the options panel today?

Chr1Z93:
Relevant section of the Option Panel

Before:
image

After:
image

Explanation is given through description and the GUI itself:
image
image

We could add them to the Tour to explain them in more detail.

BootleggerFinn (approved):
Looks good!

Chr1Z93:
> Looks good! I think we can consider adding them to the tour if there is confusion, but there's no need to just yet.

Good idea! Let's keep this in mind but wait for public reception first.

PR #349, opened 2023-08-21
Underworld Market Helper

bankey1443:
Adds a new helper for the [Underworld Market](https://arkhamdb.com/card/09077) card that can be used to help manipulate the market deck.

When the market deck is placed in the helper, the helper checks for a legal deck before splitting the deck and adding the individual cards to the object. When it is your turn, you can click 'Reveal' to reveal the first two cards of the market deck. After purchase (or no purchase) you click 'Finish' and the cards are placed back into the helper. You can click 'Swap' to swap the positions of the revealed cards before clicking 'Finish'.

If anyone would be interested in finding some better quality art or design for the helper, let me know! :)


Chr1Z93:
Thanks for the contribution!
I am down to approve and merge this to add it to the fan-made accessories (like you proposed).

If we want this to be the default solution that is also spawned by the deckimporter, I'd prefer a smaller solution (just revealing two cards below the deck and returning them to the bottom).

objects/Fan-MadeAccessories.aa8b38/UnderworldMarketHelper.6572d0.json
Chr1Z93, line 1: Some general comments:
- "Custom Tile" is generally the better object type if you have a regular shape (better performance)
- background image seems to be a bit blurry
- in-game description would be useful

bankey1443:
I am in the process of remaking the helper as a smaller object similar to the attachment helper. Hopefully will be done soon :)

Chr1Z93:
We decided on putting the Underworld Market Helper into the "Fan-made Accessories" for now. I went ahead and did that since we want to release tonight! See here: https://github.com/argonui/SCED/pull/360

PR #356, opened 2023-08-25
Deck Builder!

Entrox-Licher:
Adds a Deck Builder to the game!

Current Features:
* Loading in cards accessible to each investigator
* Full support for fringe deckbuilding requirements/restrictions, like subclasses, Vincent/Carolyn, parallel Wendy etc.
* Displaying cards in deck
* Adding/Removing cards from deck by drag and drop from display
* Exporting deck and importing deck back to deck builder or to play mat

While not in this PR yet (want to finalize the formatting first), the final update would require adding the `deckbuilding_requirements` and `deckbuilding_options` fields to all present and future investigator cards' metadata.  I can add documentation to the official metadata document if needed.  In the meantime, I have manually added the necessary metadata to 9 card copies I have on the table to be used for testing purposes (these won't be in the final PR of course)

Keep in mind in the current version there are a few bugs still (customizable cards not supported yet, an error when importing to deck builder because of metadata update not happening yet, etc.).  Also missing from this PR is updates relating to the latest changes to the API, which will also be addressed for the final submission.


src/playercards/DeckBuilder.ttslua
```diff
@@ -0,0 +1,543 @@
...
+  toggleChoiceButton(5, alt_click)
+end
+
+function toggleChoiceButton6(_, _, alt_click)
+  toggleChoiceButton(6, alt_click)
+end
+
+function toggleChoiceButton7(_, _, alt_click)
+  toggleChoiceButton(7, alt_click)
+end
```
Chr1Z93, line 275: You can append a parameter to click_functions, either via using `setVar` or `_G`. Example:
https://github.com/argonui/SCED/blob/fea12454dee7979913fb6b8d69c38d798ad14288/src/accessories/ChaosBagManager.ttslua#L41C1-L44C47

src/playercards/DeckBuilder.ttslua
```diff
@@ -0,0 +1,543 @@
...
+
+function placeCardsUpgraded(_,_, alt_click)
+  placeCards(true, alt_click)
+end
+
+function placeCards(isUpgraded, alt_click)
+  if (alt_click) then
+    --Clear all cards
+    playerCardPanelApi.deleteAll("middle")
```
Chr1Z93, line 432: Instead of having the whole functions in this big if statement, add a return to the first "path":
```
if (alt_click) then
    --Clear all cards
    playerCardPanelApi.deleteAll("middle")
    return
end
...
```

src/playercards/AllCardsBag.ttslua
```diff
@@ -3,6 +3,108 @@ local classAndLevelIndex = { }
 local basicWeaknessList = { }
 local uniqueWeaknessList = { }
 local cycleIndex = { }
+local traitIndex = { }
+local specialCriteriaIndex = { }
+
+local specialCriteriaManualOverride = {
```
Chr1Z93, line 9: This will introduce a lot of additional maintenance 😩

Entrox-Licher:
Still left to do:

- Update APIs to comply with new GUID manager
- Add metadata to alternate art investigators
- Update wiki with new metadata features


That being said, there is hopefully plenty here to review, starting with why the ModManager is failing to build it ><

src/playermat/Playmat.ttslua
```diff
@@ -620,12 +621,40 @@ function spawnTokensFor(object)
...
+    
+    -- Add Summon Servitor minicard
+    if deckData.slots["09080"] ~= nil then
+      deckData.slots["09080-m"] = 1
+    end
+    -- Add On The Mend
+
+
+    deckImporterApi.loadCards(deckData)
+end
```
Chr1Z93, line 646: This code should be part of the deckimporter instead of having it on each playmat

src/playercards/PlayerCardPanelApi.ttslua
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  PlayerCardPanelApi.spawnCardList = function(params)
+    return getPlayerCardPanel().call("spawnCardList", params)
+  end
+
+  PlayerCardPanelApi.getPlacedCards = function(targetState)
+    return getPlayerCardPanel().call("getPlacedCards", targetState)
+  end
+
```
Chr1Z93, line 26: Documentation (at least short descriptions) missing

src/playercards/CardFilterSearchApi.ttslua
```diff
@@ -0,0 +1,22 @@
...
+    for _, token in ipairs(getCardFilterSearch().call("tokenizeString", filterString)) do
+      table.insert(passthroughList, token)
+    end
+    return passthroughList
+  end
+
+  CardFilterSearchApi.createFilterCardList = function(tokenList, startingCardSet)
+    return getCardFilterSearch().call("createFilterCardList", {tokenList = tokenList, startingCardSet = startingCardSet})
+  end
+
```
Chr1Z93, line 20: Documentation missing

objects/Search-A-Card.24051a.json
```diff
@@ -42,9 +42,9 @@
   "Sticky": true,
   "Tooltip": true,
   "Transform": {
-    "posX": 60,
+    "posX": 62.953,
     "posY": 1.531,
-    "posZ": 56,
+    "posZ": 57.123,
```
Chr1Z93, line 47: Nitpick: Use rounded values

src/playercards/AllCardsBag.ttslua
```diff
@@ -131,70 +254,179 @@ function buildSupplementalIndexes()
...
-          table.insert(classAndLevelIndex["Mystic"..upgradeKey], cardMetadata.id)
+        if (string.match(cardMetadata.class, "Survivor")) then
+          table.insert(classIndex["survivor"], cardMetadata.id)
         end
-        if (isRogue) then
-          table.insert(classAndLevelIndex["Rogue"..upgradeKey], cardMetadata.id)
+        if (string.match(cardMetadata.class, "Neutral")) then
+          table.insert(classIndex["neutral"], cardMetadata.id)
+        end
+      end
```
Chr1Z93, line 284: This should be simplified to just using the values instead of 6 if-clauses. E.g. use the `|` as separator to split the string into X substrings, loop through them and insert the card id into each respective classindex

src/playercards/AllCardsBag.ttslua
```diff
@@ -131,70 +254,179 @@ function buildSupplementalIndexes()
...
+      local cycleName = cardMetadata.cycle
+      if cycleName ~= nil then
+        cycleName = string.lower(cycleName)
+        if string.match(cycleName, "return") then
+          cycleName = string.sub(cycleName, 11)
         end
-        if (isSurvivor) then
-          table.insert(classAndLevelIndex["Survivor"..upgradeKey], cardMetadata.id)
+        if cycleName == "the night of the zealot" then
+          cycleName = "core"
```
Chr1Z93, line 293: This would be a good opportunity to get rid of this hard-coded remapping

src/playercards/AllCardsBag.ttslua
```diff
@@ -131,70 +254,179 @@ function buildSupplementalIndexes()
...
+        table.insert(iconIndex.combat[0], cardMetadata.id)
+      end
+      if cardMetadata.wildIcons then
+        if not iconIndex.wild[cardMetadata.wildIcons] then
+          iconIndex.wild[cardMetadata.wildIcons] = { }
+        end
+        table.insert(iconIndex.wild[cardMetadata.wildIcons], cardMetadata.id)
+      else
+        table.insert(iconIndex.wild[0], cardMetadata.id)
+      end
```
Chr1Z93, line 368: This could be simplified too

src/playercards/AllCardsBag.ttslua
```diff
@@ -131,70 +254,179 @@ function buildSupplementalIndexes()
...
+        if not usesIndex[usesCount] then
+          usesIndex[usesCount] = { }
+        end
+        table.insert(usesIndex[usesCount], cardMetadata.id)
+        if not usesIndex[string.lower(cardMetadata.uses[1].type)] then
+          usesIndex[string.lower(cardMetadata.uses[1].type)] = { }
+        end
+        table.insert(usesIndex[string.lower(cardMetadata.uses[1].type)], cardMetadata.id)
+      end
+      if bonded then
```
Chr1Z93, line 399: This is a repetition of LOC 314: https://github.com/argonui/SCED/pull/356/files#diff-5025e54fb68163d3f0c0619249100efca58320292451f20d76c5250e2d27d2deR314

src/playercards/AllCardsBag.ttslua
```diff
@@ -233,6 +494,24 @@ function getCardById(params)
   return cardIdIndex[params.id]
 end
 
+function getFilteredIndexList(cardList, includeOthers)
+  if cardList then
```
Chr1Z93, line 498: Instead of nesting ifs this deep, account for an early "escape":
```
if not cardList then return {} end
...
```

src/playercards/AllCardsBag.ttslua
```diff
@@ -241,26 +520,119 @@ end
...
+
+function getCardsByUses(params)
+  if (not indexingDone) then
+    broadcastToAll("Still loading player cards, please try again in a few seconds", {0.9, 0.2, 0.2})
+    return { }
+  end
+  if type(params.index) == "number" then
+    return getFilteredIndexList(usesIndex[tonumber(params.index)], params.includeOthers)
+  elseif usesIndex[string.lower(params.index)] then
+    return getFilteredIndexList(usesIndex[string.lower(params.index)], params.includeOthers)
```
Chr1Z93, line 588: These functions should be turned into a single one that takes the respective index as parameter. If you end up repeating yourself this much, there is always a better approach.

src/playercards/AllCardsBagApi.ttslua
```diff
@@ -58,17 +58,69 @@ do
...
+    return getAllCardsBag().call("getCardsByLevel", {level = level, includeOthers = includeOthers})
+  end
+  
+  AllCardsBagApi.getCardsByUses = function(index, includeOthers)
+    return getAllCardsBag().call("getCardsByUses", {index = index, includeOthers = includeOthers})
+  end
+
+  AllCardsBagApi.getPermanents = function(includeOthers)
+    return getAllCardsBag().call("getPermanents", includeOthers)
+  end
```
Chr1Z93, line 99: Combine these into a single function that takes the requested index as parameter

Chr1Z93 (commented):
First batch of review comments to give you something to chew on

src/playercards/AllCardsBag.ttslua
Entrox-Licher: Agreed it does not look pretty as is.  What would be the best way to do what you're suggesting though?  Is the parameter a string and the single function is just one big chain of `elseif`s?

src/arkhamdb/DeckImporterApi.ttslua
```diff
@@ -58,5 +58,9 @@ do
     return getDeckImporter().call("loadCardsHelper", params)
   end
 
+  DeckImporterApi.spawnDeckFromToken = function(coin, playerColor)
+    return getDeckImporter().call("spawnDeckFromToken", {coin = coin, playerColor = playerColor})
+  end
+
   return DeckImporterApi
 end
```
Chr1Z93, line 66: missing end of file newline

src/playercards/AllCardsBag.ttslua
```diff
@@ -112,6 +112,24 @@ local specialCriteriaManualOverride = {
...
+  k = "traits",
+  o = "cost",
+  w = "willpowerIcons",
+  c = "combatIcons",
+  i = "intellectIcons",
+  a = "agilityIcons",
+  d = "wildIcons",
+  t = "type",
+  p = "level",
+  --z = "slot",
```
Chr1Z93: Is this inteionally commented out?

src/playercards/CardFilterSearch.ttslua
```diff
@@ -34,19 +34,27 @@ buttonParameters.font_size      = 75
...
   u = "uses",
-  q = "special"
+  q = "specialCriteria"
+}
+
+local iconOptions = {
+  w = "willpower",
+  c = "combat",
+  i = "intellect",
+  a = "agility",
```
Chr1Z93, line 56: Why are these split into a separate list instead of using the table above?

src/playercards/PlayerCardPanelApi.ttslua
```diff
@@ -6,20 +6,32 @@ do
     return guidReferenceApi.getObjectByOwnerAndType("Mythos", "PlayerCardPanel")
   end
 
+
```
Chr1Z93, line 9: Double new line

src/playercards/PlayerCardPanelApi.ttslua
```diff
@@ -6,20 +6,32 @@ do
...
+  ---@param targetState The section of the table to delete cards from
   PlayerCardPanelApi.deleteAll = function(targetState)
     local targetState = targetState or "default"
     getPlayerCardPanel().call("swapState", targetState)
     return getPlayerCardPanel().call("deleteAll")
   end
 
+  -- 
+  ---@param 
+  ---@return
```
Chr1Z93, line 20: missing documentation

src/playercards/PlayerCardPanelApi.ttslua
```diff
@@ -6,20 +6,32 @@ do
...
+  -- 
+  ---@param 
+  ---@return 
   PlayerCardPanelApi.spawnCardsByType = function(params)
     return getPlayerCardPanel().call("spawnCardsByType", params)
   end
 
+  -- 
+  ---@param 
+  ---@return
```
Chr1Z93, line 27: missing documentation

src/arkhamdb/DeckImporterMain.ttslua
```diff
@@ -53,6 +53,30 @@ function buildDeck(playerColor, deckId)
...
+    
+    -- Add Summon Servitor minicard
+    if deckData.slots["09080"] ~= nil then
+      deckData.slots["09080-m"] = 1
+    end
+    -- Add On The Mend
+
+
+    loadCardsHelper(deckData)
+end
```
Chr1Z93, line 78: I think it would be more future-proof to combine this with the existing spawn deck function instead of repeating the same process

src/playercards/CardFilterSearch.ttslua
Entrox-Licher: The table above needs to contain the exact name of the -`Index` variable from the `AllCardsBag` script.  For that reason, it has to be `icon`.  However, in order to further index into that index, we need a secondary variable, which is what this table is for.

Technically, I could refactor the icon indices to be standalone if we wanted this to be slightly cleaner, but I don't know if that then looks weird on the `AllCardsBag` side?

src/arkhamdb/DeckImporterMain.ttslua
Entrox-Licher: The issue with this is that it would require formatting the deck data into the horrible way that it is formatted when receiving it from ArkhamDB, which is something I would rather avoid 😅

Chr1Z93: It could also mean to format the horrible data into this format ;)
We should just avoid having two "data handling" methods that need updating (like the addition of "On the Mend", "Summoned Servitor" etc.)

src/playercards/AllCardsBag.ttslua
Entrox-Licher: How would you suggest?  Unlike classes, all the icons are separate metadata entries, so I can't think of any way around manually checking for each one.

src/playercards/PlayerCardPanel.ttslua
```diff
@@ -755,3 +788,122 @@ function spawnRandomWeakness()
...
+end
+
+-- @param params table:
+--    cardList: 
+--    name:
+--    startPos:
+--    rotation:
+--    spread:
+--    spreadCols:
+function spawnCardsByType(params)
```
Chr1Z93, line 821: This function looks very repetitive to me. Would it be possible to rework it for easier readability and maintainability?

src/playermat/Playmat.ttslua
```diff
@@ -630,12 +631,16 @@ function spawnTokensFor(object)
   tokenManager.spawnForCard(object, extraUses)
 end
 
-function onCollisionEnter(collisionInfo)
-  local object = collisionInfo.collision_object
+function onCollisionEnter(collision_info)
```
Chr1Z93, line 634: Let's stick to the camelCase we had before

src/playermat/Playmat.ttslua
```diff
@@ -630,12 +631,16 @@ function spawnTokensFor(object)
...
+  local object = collision_info.collision_object
 
   -- only continue if loading is completed
   if not collisionEnabled then return end
 
+  if object.hasTag("DeckBuilderToken") then
+    deckImporterApi.spawnDeckFromToken(object, playerColor)
```
Chr1Z93, line 641: There should probably be a `return` after this

src/arkhamdb/DeckImporterMain.ttslua
```diff
@@ -53,6 +53,18 @@ function buildDeck(playerColor, deckId)
...
 end
 
+-- Modelled after onDeckResult from arkhamdb/ArkhamDb.ttslua, but with some necessary changes
+function spawnDeckFromToken(params)
+  local deckData = JSON.decode(params.coin.getGMNotes())
+  destroyObject(params.coin)
+  deckData["playerColor"] = params.playerColor
+  -- Add Investigator Cards
+  deckData["slots"][deckData.investigatorId] = 1
+  deckData["slots"][string.sub(deckData.investigatorId, 1, 5) .. "-m"] = 1
```
Chr1Z93, line 63: Can you explain why this is not part of the original saved deck data?

objects/CardFilterSearch.2a4eec.json
```diff
@@ -0,0 +1,57 @@
...
+  "MeasureMovement": false,
+  "Name": "Custom_Token",
+  "Nickname": "Card Filter Search",
+  "Snap": true,
+  "Sticky": true,
+  "Tooltip": true,
+  "Transform": {
+    "posX": 51.654,
+    "posY": 1.581,
+    "posZ": 52.183,
```
Chr1Z93, line 47: Nitpick: I prefer rounded X and Z values :P

objects/AllPlayerCards.15bb07/ZoeySamaras.98a0e1.gmnotes
```diff
@@ -7,5 +7,37 @@
...
+        "max": 5
+      }
+    },
+    {
+      "level": {
+        "min": 0,
+        "max": 0
+      },
+      "limit": 5,
+      "error": "You cannot have more than 5 cards that are not Guardian or Neutral"
```
Chr1Z93, line 40: Is this going to work with Versatile?

Entrox-Licher: This doesn't do anything currently, this is more laying the groundwork for eventual deck validation.

src/arkhamdb/DeckImporterMain.ttslua
Entrox-Licher: In order to be able to import the deck to the deck builder.  The deck builder looks at the slots as well and populates a deck with it.  Don't want the investigator card included in that.  It needs to be to be compatible with the ArkhamDB code though, so it's added in manually here.

Entrox-Licher: I think it's easier just to add it in here.  In order to exclude it on the deck builder I'd have to either search through the list or add a check into a for loop.  This on the other hand is an instruction that gets executed once.  I feel like it's preferable.

src/playercards/AllCardsBag.ttslua
```diff
@@ -1,8 +1,144 @@
...
 local basicWeaknessList = { }
 local uniqueWeaknessList = { }
-local cycleIndex = { }
+cycleIndex = { }
+traitsIndex = { }
+willpowerIconsIndex = { }
+intellectIconsIndex = { }
+combatIconsIndex = { }
+agilityIconsIndex = { }
+wildIconsIndex = { }
```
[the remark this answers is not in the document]
Entrox-Licher: Haha that is how I had it in a previous version.  I changed it in response to your comments XD.  This exists this way in order to be able to trigger the global variable namespace in the `getCardsByData` function further down.  That's not possible if you have to index deeper in the way it was previously set up.

src/arkhamdb/DeckImporterApi.ttslua
```diff
@@ -38,5 +38,36 @@ do
...
+  DeckImporterApi.spawnDeckFromToken = function(coin, playerColor)
+    return getDeckImporter().call("spawnDeckFromToken", {coin = coin, playerColor = playerColor})
+  end
+
+  -- Callback function for investigator cards and minicards to set the correct state for alt art
+  ---@param card Object Card which needs to be set the state for
+  ---@param loadAltInvestigator String Contains the name of alternative art for the investigator ("normal", "revised" or "promo")
+  DeckImporterApi.loadAltArt = function(params)
+    return getDeckImporter().call("loadAltArt", params)
+  end
```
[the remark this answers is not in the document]
Entrox-Licher: We don't, apparently!  The need for this must have been removed when I redid the import function, good catch!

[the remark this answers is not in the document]
Entrox-Licher: This is called in the Deck Builder on LOC 507

src/playercards/AllCardsBag.ttslua
```diff
@@ -1,8 +1,144 @@
 local cardIdIndex = { }
-local classAndLevelIndex = { }
+local otherIndex
```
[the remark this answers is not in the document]
Entrox-Licher: For the same reason as the iconIndex idea above, nesting important subindices like this breaks other functionality.  Additionally, these cards are tracked totally differently (set-style, rather than map style), so it would make things very messy to mix them  (as explained more in the other comment below on the otherIndex).

src/playercards/AllCardsBag.ttslua
```diff
@@ -214,16 +491,45 @@ end
...
   end
   if (card1.data.Nickname ~= card2.data.Nickname) then
     return card1.data.Nickname < card2.data.Nickname
   end
   return card1.data.Description < card2.data.Description
 end
 
+function sortCardsTable (cardsTable)
+  return table.sort(cardsTable, cardComparator)
+end
```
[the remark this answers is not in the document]
Entrox-Licher: Good catch, I should be!

src/playercards/AllCardsBag.ttslua
```diff
@@ -141,70 +306,182 @@ function buildSupplementalIndexes()
...
+      --       local trait = string.lower(string.gsub(traitString, "[.]", ""))
+      --       if not checkedTraits[trait] then
+      --         if traitsIndex[trait] == nil then
+      --           traitsIndex[trait] = { }
+      --         end
+      --         table.insert(traitsIndex[trait], cardMetadata.id)
+      --       end
+      --     end
+      --   end
+      -- end
```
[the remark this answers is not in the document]
Entrox-Licher: Good point.  Still sad about that taboo reversal though XD

src/playercards/AllCardsBag.ttslua
```diff
@@ -141,70 +306,182 @@ function buildSupplementalIndexes()
...
+        if string.match(cycleName, "return") then
+          cycleName = string.sub(cycleName, 11)
+        end
+        if cycleName == "the night of the zealot" then
+          cycleName = "core"
         end
-        if (isGuardian) then
-          table.insert(classAndLevelIndex["Guardian"..upgradeKey], cardMetadata.id)
+        if cycleIndex[cycleName] == nil then
+          cycleIndex[cycleName] = { }
```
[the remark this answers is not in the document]
Entrox-Licher: Oooo nice, will do!

src/playercards/AllCardsBag.ttslua
```diff
@@ -28,20 +164,49 @@ end
...
+  slotIndex["arcane x2"] = { }
+  slotIndex["ally"] = { }
+  slotIndex["accessory"] = { }
+  slotIndex["body"] = { }
+  slotIndex["tarot"] = { }
+  slotIndex["hand"] = { }
+  slotIndex["hand x2"] = { }
+  slotIndex[1] = { }
+  slotIndex[2] = { }
+  slotIndex[3] = { }
```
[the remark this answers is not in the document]
Entrox-Licher: Maybe?  The only scenario I can think of is if someone tries to `ipairs` it, which shouldn't ever really happen.  The whole point of this index is that it isn't ever modified or read except by the functions explicitly defined in this file, so I think the danger is relatively minimal.

src/arkhamdb/ArkhamDb.ttslua
```diff
@@ -56,12 +56,6 @@ do
       loadNewest,
       loadInvestigators,
       callback)
-    -- Get a simple card to see if the bag indexes are complete. If not, abort
```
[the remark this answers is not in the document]
Entrox-Licher: A misconception on my part.  I assumed it was redundant given my changes but I see now that it isn't.  Adding it back in!

src/playercards/DeckBuilder.ttslua
```diff
@@ -0,0 +1,682 @@
...
+  }
+end
+
+function processCardSpec(specList, filterType, isFirst)
+  local filterSubString = ""
+  if not isFirst then
+    if filterType == "t" then
+      filterSubString = " &"
+    else
+      filterSubString = " |"
```
Chr1Z93, line 666: Why are there spaces on top of the delimiter? And why does "type" get a different one?

src/playercards/DeckBuilder.ttslua
```diff
@@ -0,0 +1,682 @@
...
+    end
+  end
+  if #specList > 1 then
+    filterSubString = filterSubString .. " ("
+  end
+  for _, instance in ipairs(specList) do
+    if _ ~= 1 then
```
Chr1Z93, line 673: Give the loop index a name if you are using it

src/playercards/PlayerCardPanel.ttslua
```diff
@@ -417,9 +433,24 @@ end
 
 -- Deletes all cards currently placed on the table
 function deleteAll()
+  if currentState == "default" then
+    getObjectFromGUID('6d55fe').setVar("deckPlaced", false)
```
Chr1Z93, line 437: Avoid GUID reference

src/playercards/PlayerCardPanelApi.ttslua
```diff
@@ -0,0 +1,49 @@
...
+
+  -- Spawn a given list of cards, but seperated into groups based on their card type (asset, event, or skill)
+  ---@param cardlist List containing sublists for assets, events, and skills (and optionally limited versions of each), each sublist containing a list of card ids of cards to spawn
+  ---@param name Uniquely identifying name for the group of cards to be spawned
+  ---@param startPos String for the desired start position, based on those available in the PlayerCardPanel
+  ---@param rotation The rotation for the cards to be spawned
+  ---@param spread Boolean of whether to spread out the cards (true) or stack them in a deck (false)
+  ---@param spreadCols Int of how many cards should be placed before starting a new row
+  ---@param targetState String of what section of the table to spawn the cards, based on the option in PlayerCardPanel
+  PlayerCardPanelApi.spawnCardsByType = function(params)
```
Chr1Z93, line 25: Can have the parameters as actual separate arguments and then forward them as table like other Apis do

src/playercards/AllCardsBag.ttslua
```diff
@@ -243,34 +549,93 @@ function getCardById(params)
...
+  end
+
+  -- If not including non-player cards, filter them out and then return
+  local returnList = {}
+  for _, cardId in ipairs(cardList) do
+    if not otherIndex[cardId] then
+      table.insert(returnList, cardId)
+    end
+  end
+  return returnList
```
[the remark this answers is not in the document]
Chr1Z93: Ah, thanks for the clarification.

src/playercards/CardFilterSearchApi.ttslua
```diff
@@ -0,0 +1,29 @@
...
+  -- Turns a string into a list of individual tokens
+  ---@param filterString String to be broken down into tokens
+  ---@return List of tokens, broken down from the given string
+  CardFilterSearchApi.tokenizeString = function(filterString)
+    local passthroughList = {}
+    for _, token in ipairs(getCardFilterSearch().call("tokenizeString", filterString)) do
+      table.insert(passthroughList, token)
+    end
+    return passthroughList
```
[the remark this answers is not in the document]
Entrox-Licher: TTS's "cannot use resources from other files" error.  I'll add a comment to that effect

dscarpac:
Played around with the latest version a bit this morning with totally fresh eyes, so these comments are hopefully indicative of someone who is new to the feature (even more so for someone new to the mod/TTS). In general, I like the overall functionality and it will make a nice addition to the mod! (I especially see the use for custom investigators!)

1) Wasn't immediately obvious what to do. Eventually I got that it needs to be used in conjunction with spawning investigator cards to the left, but in general the objects on the table of this sort haven't interacted with each other much in the past. Visually, maybe the "player cards object" art needs to align better with the new deck builder/card search to make it more clear that these are all working together. The "spawn a card" object might need to be a lot bigger given the size of the "card filter search" object. Definitely could use some much more obvious directions, perhaps a "place investigator card here" graphic to get people started.

2) The "Place Deck On Table" button drops the deck fairly far out of view if you don't know what's supposed to happen. We should alert viewers to the location of the deck.

3) The card filter search is great, but a little confusing with the instructions. There is a lot of information in the GM Notes, but it's impossible to read it all as a tool tip as it goes off the screen. We should highlight the differences in the features compared to arkhamDB: namely, that you cannot search by card text (obvious to us why this is impossible, but not to users), and that chaining works differently (the GM notes states this, but an example to illustrate would be nice)

4) I was a little surprised that the card filter search didn't filter based on the investigator I had selected. E.g. "k:illicit" with Skids gives me Survivor and Guardian illicit cards. I get that you might not want to do this all the time, but I think it's an important feature to add, or at least clue users into what is going on. (again, this is another expectation I had coming from arkhamDB).

5) Turning the deck into a coin is cool, but it also wasn't obvious what to do with it afterwards.

6) Idea: how about showing how much XP is in the deck as you build it? Might be useful for building standalones, or keeping track of XP for deck upgrading or building with Kymani/Father Mateo/etc. Unsure how this would work with customizables (does the customizable sheet go into the deck? similar question about intended placement of permanents or bonded cards) but maybe worth a shot.

Entrox-Licher:
> Played around with the latest version a bit this morning with totally fresh eyes, so these comments are hopefully indicative of someone who is new to the feature (even more so for someone new to the mod/TTS). In general, I like the overall functionality and it will make a nice addition to the mod! (I especially see the use for custom investigators!)
> 
> 1. Wasn't immediately obvious what to do. Eventually I got that it needs to be used in conjunction with spawning investigator cards to the left, but in general the objects on the table of this sort haven't interacted with each other much in the past. Visually, maybe the "player cards object" art needs to align better with the new deck builder/card search to make it more clear that these are all working together. The "spawn a card" object might need to be a lot bigger given the size of the "card filter search" object. Definitely could use some much more obvious directions, perhaps a "place investigator card here" graphic to get people started.
> 2. The "Place Deck On Table" button drops the deck fairly far out of view if you don't know what's supposed to happen. We should alert viewers to the location of the deck.
> 3. The card filter search is great, but a little confusing with the instructions. There is a lot of information in the GM Notes, but it's impossible to read it all as a tool tip as it goes off the screen. We should highlight the differences in the features compared to arkhamDB: namely, that you cannot search by card text (obvious to us why this is impossible, but not to users), and that chaining works differently (the GM notes states this, but an example to illustrate would be nice)
> 4. I was a little surprised that the card filter search didn't filter based on the investigator I had selected. E.g. "k:illicit" with Skids gives me Survivor and Guardian illicit cards. I get that you might not want to do this all the time, but I think it's an important feature to add, or at least clue users into what is going on. (again, this is another expectation I had coming from arkhamDB).
> 5. Turning the deck into a coin is cool, but it also wasn't obvious what to do with it afterwards.
> 6. Idea: how about showing how much XP is in the deck as you build it? Might be useful for building standalones, or keeping track of XP for deck upgrading or building with Kymani/Father Mateo/etc. Unsure how this would work with customizables (does the customizable sheet go into the deck? similar question about intended placement of permanents or bonded cards) but maybe worth a shot.

Thanks so much for your feedback!  Here's my thoughts

1. Good point!  Some of this can be facilitate with the tour, but a short explainer in the description will be helpful, similar to what was done in the Card Filter Search.
2. Another great point!  I'll add in a ping, maybe even change the location to be slightly closer
3. Heh, so the problem is there's too much information but also not enough information XD?  In all seriousness, I intentionally took a more verbose approach here to make it explicit as to the syntax that can be used.  I would venture to say most users of ArkhamDB aren't even aware of most of the features you can use there (for ORs or ANDs, etc.).  So I wanted to just have a clean explanation of our rules, rather than comparing to ArkhamDB too much.  That being said, I agree there should probably be a callout that card text specifically cannot be searched.  As far as there being a lot of text: it's true!  No real way around it other than offloading it to the GitHub wiki, which would be terrible (usability information outside of the mod is bad).  The notebook could be an option, but honestly, no one really checks it (and that's partially our fault perhaps).
4. I considered this, but ultimately decided against it and implemented the 'Filter Placed' option instead.  I definitely didn't want it to be on be default, since that could be very confusing.  I could maybe try implementing it as a toggle though, I'll play around with that.
5. Fair point!  I'll add a description to the coin object.
6. Already in development!  But this will be part of a future version of the tool, since anything that strays into the realm of deck validation is going to require a lot more work to get things working properly.  This PR is already huge, so trying to keep it manageable.

Thanks again for the feedback!

dscarpac:
> > Played around with the latest version a bit this morning with totally fresh eyes, so these comments are hopefully indicative of someone who is new to the feature (even more so for someone new to the mod/TTS). In general, I like the overall functionality and it will make a nice addition to the mod! (I especially see the use for custom investigators!)
> > 
> > 1. Wasn't immediately obvious what to do. Eventually I got that it needs to be used in conjunction with spawning investigator cards to the left, but in general the objects on the table of this sort haven't interacted with each other much in the past. Visually, maybe the "player cards object" art needs to align better with the new deck builder/card search to make it more clear that these are all working together. The "spawn a card" object might need to be a lot bigger given the size of the "card filter search" object. Definitely could use some much more obvious directions, perhaps a "place investigator card here" graphic to get people started.
> > 2. The "Place Deck On Table" button drops the deck fairly far out of view if you don't know what's supposed to happen. We should alert viewers to the location of the deck.
> > 3. The card filter search is great, but a little confusing with the instructions. There is a lot of information in the GM Notes, but it's impossible to read it all as a tool tip as it goes off the screen. We should highlight the differences in the features compared to arkhamDB: namely, that you cannot search by card text (obvious to us why this is impossible, but not to users), and that chaining works differently (the GM notes states this, but an example to illustrate would be nice)
> > 4. I was a little surprised that the card filter search didn't filter based on the investigator I had selected. E.g. "k:illicit" with Skids gives me Survivor and Guardian illicit cards. I get that you might not want to do this all the time, but I think it's an important feature to add, or at least clue users into what is going on. (again, this is another expectation I had coming from arkhamDB).
> > 5. Turning the deck into a coin is cool, but it also wasn't obvious what to do with it afterwards.
> > 6. Idea: how about showing how much XP is in the deck as you build it? Might be useful for building standalones, or keeping track of XP for deck upgrading or building with Kymani/Father Mateo/etc. Unsure how this would work with customizables (does the customizable sheet go into the deck? similar question about intended placement of permanents or bonded cards) but maybe worth a shot.
> 
> Thanks so much for your feedback! Here's my thoughts
> 
> 1. Good point!  Some of this can be facilitate with the tour, but a short explainer in the description will be helpful, similar to what was done in the Card Filter Search.
> 2. Another great point!  I'll add in a ping, maybe even change the location to be slightly closer
> 3. Heh, so the problem is there's too much information but also not enough information XD?  In all seriousness, I intentionally took a more verbose approach here to make it explicit as to the syntax that can be used.  I would venture to say most users of ArkhamDB aren't even aware of most of the features you can use there (for ORs or ANDs, etc.).  So I wanted to just have a clean explanation of our rules, rather than comparing to ArkhamDB too much.  That being said, I agree there should probably be a callout that card text specifically cannot be searched.  As far as there being a lot of text: it's true!  No real way around it other than offloading it to the GitHub wiki, which would be terrible (usability information outside of the mod is bad).  The notebook could be an option, but honestly, no one really checks it (and that's partially our fault perhaps).
> 4. I considered this, but ultimately decided against it and implemented the 'Filter Placed' option instead.  I definitely didn't want it to be on be default, since that could be very confusing.  I could maybe try implementing it as a toggle though, I'll play around with that.
> 5. Fair point!  I'll add a description to the coin object.
> 6. Already in development!  But this will be part of a future version of the tool, since anything that strays into the realm of deck validation is going to require a lot more work to get things working properly.  This PR is already huge, so trying to keep it manageable.
> 
> Thanks again for the feedback!

re: 3. Just to be clear, I like the amount of text! The problem is just that the tooltip makes it impossible to read on the screen. What do you think of using an XML pop-up or side panel to provide the instructions? Something like the settings panel on the Clean Up Helper? It'd be easi(er) to keep some documentation there, and even allow for a scroll bar. I only found out about the Notebook a few days ago, so I agree that that's not a good option.

Chr1Z93:
Closed until someone wants to pick this back up

PR #358, opened 2023-08-25
Adding initial "The Feast of Hemlock Vale" cards

Chr1Z93:
image

This includes an update to the player card panel image:
image


Tikatoy:
When these cards are in hand, Pitchfork front shows up as the player back image.

Chr1Z93:
> When these cards are in hand, Pitchfork front shows up as the player back image. Screen Shot 2023-08-29 at 9 20 40 PM

Huh that's weird. I'll fix that. Thanks for catching that!

Chr1Z93:
@Tikatoy fixed the Pitchfork back issue

Tikatoy:
Looks good!

Tikatoy:
Were you thinking of adding "I'll Pay You Back!" as well? There's only screenshots from the live stream so it would be low quality.

Tikatoy (approved):

Chr1Z93:
Note to reader: "I'll Pay You Back!" will be added with the next batch of spoilered cards

PR #365, opened 2023-08-28
Token Arranger: more resilience to latency

Chr1Z93:
- By pushing back the spawning and moving into the same loop, we have more resilience to high latency clients in multiplayer sessions. Without this, it's possible for all tokens to complete spawning before getting moved and thus "exploding" because of the collisions.
- Additionally, spawning them directly at the correct position should avoid issues with "ghost tokens"
- Removal of percentage buttons updated
- click-/input-function setting updated


Tikatoy:
I'm starting a new 4P campaign tomorrow so should be a good test case for this.

BootleggerFinn (approved):

PR #366, opened 2023-08-28
Resource Hotkey: load correct state

Chr1Z93:
The hotkey for spawning resources (numpad 9 by default) now spawns the correct type for the below card.


Entrox-Licher (approved):
Really cool change!

PR #368, opened 2023-08-29
Scripted Well Connected

bankey1443:
Adds scripting to [Well Connected](https://arkhamdb.com/card/05028) as well as its [upgrade](https://arkhamdb.com/card/54006) to add a context menu item that, when toggled, displays the skill boost on the card as an alternative to having to do math.

Changes include adding a new function `getResourceCount()` to the `playermat/playmat` API that is used to get the resource count for an individual player by color.

image


src/playermat/PlaymatApi.ttslua
```diff
@@ -178,6 +178,15 @@ do
...
 
+  -- Returns the resource counter amount for the requested playermat
+  PlaymatApi.getResourceCount = function(matColor)
+    local resources = 0
+    for _, mat in ipairs(internal.getMatForColor(matColor)) do
+      resources = resources + mat.call("getResourceCount")
+    end
+    return resources
+  end
+
```
Chr1Z93, line 189: Is there a specific reason for returning the sum of all resources instead of the specific values of the mats?

src/playercards/cards/WellConnected.ttslua
Chr1Z93, line 1: Imo it would be better if both cards shared the same script file since they are so similar. You could use the `id` from the gm notes to discern between them (e.g. check the id in onload and set `modValue` accordingly). That would make this more resilient by avoiding duplicate code.

src/playercards/cards/WellConnected.ttslua
```diff
@@ -0,0 +1,52 @@
...
+    position = {-0.25,1,-0.75},
+    font_color = {1,1,1,100},
+    font_size = 250,
+    width = 0,
+    height = 0
+}
+
+function onLoad()
+  self.addContextMenuItem('Toggle Counter', toggleCounter)
+end
```
Chr1Z93, line 21: This needs an `onSave()` function to remeber the context menu selection after saving & loading.

src/playercards/cards/WellConnected.ttslua
```diff
@@ -0,0 +1,52 @@
...
+
+local b_display = {
+    index = 0,
+    click_function = 'none',
+    function_owner = self,
+    position = {-0.25,1,-0.75},
+    font_color = {1,1,1,100},
+    font_size = 250,
+    width = 0,
+    height = 0
```
Chr1Z93, line 16: I think the button should get a background and be moved into the upper right corner. See here for inspiration: image

Chr1Z93 (changes requested):

src/playercards/cards/WellConnected.ttslua
```diff
@@ -0,0 +1,70 @@
...
+    if notes.id == "54006" then -- hardcoded card id for upgraded Well Connected (3)
+        modValue = 4 -- Well Connected (3)
+    end
+
+    if saved_data != '' then
+        local loaded_data = JSON.decode(saved_data)
+        display = not loaded_data.saved_display
+
+        toggleCounter()
+    end
```
Chr1Z93, line 30: This causes an error if you save & load when the display is not enabled since it does not have a loopId defined, but tries to stop the loop

Chr1Z93 (changes requested):

Chr1Z93 (approved):

Issue #371, opened 2023-09-02
Underworld Market Helper breaks after saving & loading


Chr1Z93:
@bankey1443

PR #376, opened 2023-09-09
Fix: Underworld Market Helper breaks after saving & loading

bankey1443:
Fixes issue #371 with the Underworld Market Helper where saving and loading would break the helper.


Chr1Z93 (approved):
Looks good to me, thanks for fixing!

PR #378, opened 2023-09-21
Uses spawning: additional option

Chr1Z93:
Added a new option to the "clickable resource counters" to spawn clickable counters only on "0 uses" + similar assets (since those tend to change their count often)

Relevant section of the option panel (new look):
image

Showcase of options (enabled, custom, disabled from top to bottom):
image


Entrox-Licher:
I think it'd be helpful to add a brief explanation of what "Chef's Selection" does to the options menu itself.

Chr1Z93:
> I think it'd be helpful to add a brief explanation of what "Chef's Selection" does to the options menu itself.

Added some more description:
image

BootleggerFinn (approved):

PR #381, opened 2023-09-26
Removal of log function overrides

Chr1Z93:
This removes the overriding of the native `log()` function (bad practice).


Entrox-Licher (approved):
Strongly agree, overriding the built-in log will only ever lead to confusion and inconsistency I think.

PR #385, opened 2023-09-29
Addition of GUID reference handler and api cleanup

Chr1Z93:
This adds the "GUID Reference Handler". This object holds all GUID references sorted by the objects owner (playermat or "Mythos"). This allows to clean up api code and enables future dynamic updates of these GUID references (for example spawning a new playmat).


Buhallin:
This is going to have very negative impacts on performance.  GUIDs may be somewhat fragile but they're O(1) for lookups.  This replaces those object lookups with an O(n) linear search through all objects on the table.

The fragility in GUIDs comes primarily from deletion/recreation risk, as the object may exist but doesn't have the expected GUID.  If that is a problem (though it's not one that has cropped up before) it can be managed by maintaining a global lookup table of object name->GUID to ensure the lookups remain O(1).  But even that is complex, and should be undertaken to fix an existent issue.

Chr1Z93:
> This is going to have very negative impacts on performance. GUIDs may be somewhat fragile but they're O(1) for lookups. This replaces those object lookups with an O(n) linear search through all objects on the table.
> 
> The fragility in GUIDs comes primarily from deletion/recreation risk, as the object may exist but doesn't have the expected GUID. If that is a problem (though it's not one that has cropped up before) it can be managed by maintaining a global lookup table of object name->GUID to ensure the lookups remain O(1). But even that is complex, and should be undertaken to fix an existent issue.

That's a valid concern. Underlying reason for this change is ultimately the idea to support different playermat layouts (including removal of mats). I wanted to avoid adding more tags, but I think you are right about the performance hit. I did some tests and this is the result:
image
Script used if you are interested:
```lua
  local startTime = os.time()
  local obj = getObjectFromGUID("721ba2")
  local duration = os.time() - startTime
  log("GUID time: " .. duration)
  --
  local startTime = os.time()
  local obj = getObjectFromMemo({matColor = "Mythos", type = "PlayArea"})
  local duration = os.time() - startTime
  log("Memo time: " .. duration)
  --
  local startTime = os.time()
  local obj = getObjectsWithTag("PlayArea")[1]
  local duration = os.time() - startTime
  log("Tag time: " .. duration)
  --
  local startTime = os.time()
  local obj = getObjectsWithAllTags({"Playermat", "Red"})[1]
  local duration = os.time() - startTime
  log("Tags time: " .. duration)
```
Apparently the native functions to TTS are all "indexed" and basically instantenous and while looping through the objects is fast (especially considering that the loop starts with the "oldest" object), it is still introducing a delay.
So I'll change this refactor to use tags instead.

Buhallin:
os.time() is seconds resolution - 7ms for a single object lookup is actually pretty atrocious, and it would only get worse on a table full of cards and tokens.

A few (mostly rhetorical) questions:
- If the main reasoning for this is playmat layout why is it being applied to static and even hidden objects like the All Cards Bag and DeckImporter, which have nothing to do with playmats?
- If the concern is about nonexistent objects, the lookup technique won't have anything to do with it.  Callers will have to be adapted to handle nil or failed responses no matter how you don't find an expected object
- If the issue is with objects that will be deleted and then recreated after a layout change, the GUIDs could be stored for reference.  Playmat already did this, I believe.  Pros and cons to either this or tags

Chr1Z93:
> os.time() is seconds resolution - 7ms for a single object lookup is actually pretty atrocious, and it would only get worse on a table full of cards and tokens.
> 
> A few (mostly rhetorical) questions:
> 
> * If the main reasoning for this is playmat layout why is it being applied to static and even hidden objects like the All Cards Bag and DeckImporter, which have nothing to do with playmats?
> * If the concern is about nonexistent objects, the lookup technique won't have anything to do with it.  Callers will have to be adapted to handle nil or failed responses no matter how you don't find an expected object
> * If the issue is with objects that will be deleted and then recreated after a layout change, the GUIDs could be stored for reference.  Playmat already did this, I believe.  Pros and cons to either this or tags

1) You're correct. It's better to stick to GUIDs for those static objects.

2) Yeah, it's not about non-existent objects.

3) This is the thing. Imagine a bag with a "playmat set" which contains all playmat related objects. This can be used to respawn playmats after they were destroyed (by checking out a 2-player layout for example). The respawned mats won't have the same GUID as before.

Chr1Z93:
I think I get what you mean with 3) and I like that approach. I don't think we have a storage for GUIDs yet, but I think that would be sensible.

Buhallin:
I believe you can provide the GUID when spawning an object. So long as the GUID doesn't actively exist when you do it should work fine.

Chr1Z93:
> I believe you can provide the GUID when spawning an object. So long as the GUID doesn't actively exist when you do it should work fine.

Yeah, that's a good idea. I added all the GUID references to a single object that's now going to handle them.
When something changes (e.g. spawning a new playmat), the spawn method can tell the reference handler the new guid and it can store it (this part is not yet implemented).

Entrox-Licher:
I'm pretty sure I know the answer to this, but I think it's important to document it/be on the same page:

Why did you elect to have the GUIDReferenceHandler not follow the API pattern?

Chr1Z93:
> I'm pretty sure I know the answer to this, but I think it's important to document it/be on the same page:
> 
> Why did you elect to have the GUIDReferenceHandler not follow the API pattern?

It needs to be an actual object so that we canuse it for storage of an updated table. Imagine this:
- Playmat GUID changes (was respawned for whatever reason)
- spawning method can now tell the reference handler the new GUID and it can update it's internal table
- it can also store a table with these changed GUIDs and load them
- additionally, it would be quite the overkill to require such a file everywhere

Entrox-Licher:
Yeah it definitely has to be an object, I was more thinking along the lines of the last bullet point (since all the APIs are based around actual objects anyway). It still leaves a slight vulnerability/inconsistency, but I think it makes sense for this context.

Chr1Z93:
> Yeah it definitely has to be an object, I was more thinking along the lines of the last bullet point (since all the APIs are based around actual objects anyway). It still leaves a slight vulnerability/inconsistency, but I think it makes sense for this context.

Ah, now I see what you mean. An Api for the accessing of the GUIDs. We can do that and might need one either way for storing data. Let's keep this in mind and get back to that in the future.

src/core/GUIDReferenceHandler.ttslua
```diff
@@ -0,0 +1,100 @@
...
+function getObjectsByType(type)
+  local objList = {}
+  for owner, objects in pairs(GuidReferences) do
+    local obj = getObjectFromGUID(objects[type])
+    if obj then
+      objList[owner] = obj
+    end
+  end
+  return objList
+end
```
Entrox-Licher, line 89: I'm a little concerned about just how intuitive it will be to use this for the 'Mythos' owner, since arguably at least half of those aren't really applicable at all, it's just a catch-all.  Would it make sense to maybe refactor the `getObjectsByType` method to return a lone object instead of a list if only 1 is found, or maybe create a new method specifically for these singletons?

src/playermat/ClueCounter.ttslua
```diff
@@ -67,20 +65,8 @@ function findValidItemsInSphere()
   return retval
 end
 
-function removeAllClues()
```
Entrox-Licher, line 70: The changes made here to the clue clicker break it; specifically when switching between clue tokens to clue clicker, the clue tokens fail to delete and throw an error: 
`function <removeAllClues>: cannot convert a table to a clr type LuaGameObjectScript`

Entrox-Licher (commented):
Just a pair of minor comments, only able to find 1 bug, pretty impressive considering the scope of the changes XD

src/core/token/TokenSpawnTracker.ttslua
```diff
@@ -39,31 +28,20 @@ function resetTokensSpawned(cardGuid)
   spawnedCardGuids[cardGuid] = nil
 end
 
-function resetAllAssetAndEvents()
```
Entrox-Licher, line 42: The changes here also make it so putting a card into a hand zone no longer resets its ability to spawn tokens

src/playermat/ClueCounter.ttslua
Entrox-Licher: The new changes do fix the behaviour, but their is a new error thrown now:

`function <onClick_toggleOption>: attempt to index nil value`

PR #386, opened 2023-10-02
Added placeholder boxes

Chr1Z93:
- The Invisible Man by The Beard
- Return to the Wendigo by Myriad
- Return to Consternation on the Constellation by Myriad


Entrox-Licher (approved):
More fun content, looks good!

PR #387, opened 2023-10-03
Add "Asset" tag to Enchant Weapon for snapping

Chr1Z93:
While "Enchant Weapon" is not an asset per-se, it very much behaves like one since it basically takes up an arcane slot. This adds the tag to support snapping.

### DIFF
--- a/objects/AllPlayerCards.15bb07/EnchantWeapon3.33455f.json
+++ b/objects/AllPlayerCards.15bb07/EnchantWeapon3.33455f.json
@@ -42,6 +42,7 @@
   "Snap": true,
   "Sticky": true,
   "Tags": [
+    "Asset",
     "PlayerCard"
   ],
   "Tooltip": true,


Buhallin:
Enchant Weapon is literally the poster child for why snapping was limited to assets in the first place.

If someone wants to put it in a slot when it doesn't snap they can, with minor mental trauma because it doesn't align.

If someone wants to attach it to the weapon (as the text instructs) it is nearly impossible to do so when it's snapping.

Chr1Z93:
> Enchant Weapon is literally the poster child for why snapping was limited to assets in the first place.
> 
> If someone wants to put it in a slot when it doesn't snap they can, with minor mental trauma because it doesn't align.

Disabling snapping for cards like "Reliable" makes perfect sense.
"Enchant Weapon" is different since it exhausts on its own and the weapon now takes an additional arcane slot, so you cna just place "Enchant Weapon" in that slot.

> 
> If someone wants to attach it to the weapon (as the text instructs) it is nearly impossible to do so when it's snapping.
Do you mean the attach tool in TTS to make it a long card? Then it can't be exhausted on its own.

Buhallin:
> Do you mean the attach tool in TTS to make it a long card? Then it can't be exhausted on its own.
No, I mean just placing Enchant Weapon so it overlaps the weapon it's attached to, likely crossing slot bounds.

The card instructs you to attach it to a weapon.  The weapon it's attached to is relevant game state info.  Placing it in a slot loses that visual connection and I don't know why anyone would do it, but that's not relevant.

Some will want to put it in the slot.  Those who want it in the slot can still do so even if it doesn't snap.  Some people will want to put EW with the weapon.  Those who want it with the weapon CANNOT do so if the card snaps (short of disabling snapping for that one card, which few know to do).

As it is both approaches work, with a minor loss of convenience for those who want it in the separate slot.  Snapping it forces players into the slot approach, and there's not enough impact on slot-without-snap to do so.

Chr1Z93:
> > Do you mean the attach tool in TTS to make it a long card? Then it can't be exhausted on its own.
> > No, I mean just placing Enchant Weapon so it overlaps the weapon it's attached to, likely crossing slot bounds.
> 
> The card instructs you to attach it to a weapon. The weapon it's attached to is relevant game state info. Placing it in a slot loses that visual connection and I don't know why anyone would do it, but that's not relevant.
> 
> Some will want to put it in the slot. Those who want it in the slot can still do so even if it doesn't snap. Some people will want to put EW with the weapon. Those who want it with the weapon CANNOT do so if the card snaps (short of disabling snapping for that one card, which few know to do).
> 
> As it is both approaches work, with a minor loss of convenience for those who want it in the separate slot. Snapping it forces players into the slot approach, and there's not enough impact on slot-without-snap to do so.

Are you aware that you can right-click while holding a card to place it down without snapping?

Buhallin:
I was not, but I expect the vast majority of users don't know that either.  I spent a lot of time searching for that capability before going to Asset-snap route, with nothing.  A couple of searches now turns up nothing on it.  The TTS UI is very unintuitive as it shows the shadow where the card will snap.

"Impossible" is not meaningfully different from "Possible as long as the user knows a super-secret trick".

Entrox-Licher:
Don't know how universal my experience is, but for what it's worth, I already have to move cards off to the side to attach them, since I prefer to overlap the asset on top of the event (so it takes up much less space, otherwise it covers another slot).  Asset snapping already makes that impossible without the trick Chr1z mentioned, so I don't think this is a great loss personally.

That, combined with the fact that Enchant Weapon needs to be exhausted separate from the asset, makes me feel this is an OK change.

PR #388, opened 2023-10-03
Bugfix for clue counts > 12

Chr1Z93:
Since there are locations with more than 12 clues for 4 player playthroughs, this enables spawning clues for them.
Example location:
image


Entrox-Licher (approved):
Approving for now, but 9+ clue spawning is an open issue (#390) that we want to improve upon in the future.

Issue #406, opened 2023-10-11
improved download window


Chr1Z93:
Potential new features to add:
- tabs (e.g. Campaigns / Scenarios / Fan-made Content)
- search bar / filter
- splashart preview on hover

PR #410, opened 2023-10-13
Revamped Download Window

Chr1Z93:
Preview:
image

- added tab buttons to swap content types
- added automatic refresh onLoad() and tab swap
- added documentation
- added preview window
- removed legacy code

closes https://github.com/argonui/SCED/issues/406


Entrox-Licher:
I'd recommend making it so the preview window pops up on a single click, and the content spawns on a double click (or add a spawn button in place/below the percentage bar).  The hovering to show the preview window can be quite finnicky and error prone, particularly as it is loading in a lot of new images at once that the player doesn't necessarily want/need.

Entrox-Licher:
Additionally, the navigation overlay does not seem to be appearing anymore.  The button is still present and can be clicked on, but does not actually display the overlay

Entrox-Licher:
Love the new changes!  You mentioned removing the placeholder boxes altogether, is that something you want to do in this PR?

Also, any chance you could add a feature that pinged the location of the thing spawned in?  And do we maybe want to close the window after something is spawned in?  (Not strong opinions on these ones, just musings)

Chr1Z93:
> Love the new changes!  You mentioned removing the placeholder boxes altogether, is that something you want to do in this PR?
> 
> Also, any chance you could add a feature that pinged the location of the thing spawned in?  And do we maybe want to close the window after something is spawned in?  (Not strong opinions on these ones, just musings)

I think if we want to remove them that can be a separate PR (need to discuss this in the group first).

The other two suggestions are good! I think we could also unify the spawn position.

src/core/Global.ttslua
```diff
@@ -619,176 +638,285 @@ end
...
 end
 
-function onClick_select(player, params)
-  params = JSON.decode(urldecode(params))
-  local url = SOURCE_REPO .. '/' .. params.url
-  local request = WebRequest.get(url, function (request) complete_obj_download(request, params) end )
-  requestObj = request
+-- click function for the items in the download window
+-- updates backgroundcolor for row panel and fontcolor for list item
+function onClick_select(_, _, identificationKey)
```
Entrox-Licher, line 658: Unsure what it is, but you have an error case here you're not catching (specifically throwing the error `attempt to concatenate a nil value`).  You can try selecting "Bad Batch" under Fan-Made Player Cards to test

Entrox-Licher: Additionally, you might want to consider adding the clicking sound effect to these (but I don't feel strongly about it)

src/core/Global.ttslua
```diff
@@ -619,176 +638,285 @@ end
...
+    end
+  end
+
+  UI.setAttribute("panel" .. currentListItem, "color", "grey")
+  UI.setAttribute(contentToShow .. "_" .. currentListItem, "color", "black")
+  updatePreviewWindow()
+end
+
+-- click function for the download button in the preview window
+function onClick_download()
```
Entrox-Licher, line 684: Should definitely add the click sound effect to this one

Entrox-Licher (commented):
Only one error found, the rest is stylistic changes.  Functionality is awesome!  Super stoked to have this in the mod, a massive improvement!

src/core/Global.ttslua
Chr1Z93: Done!

src/core/Global.ttslua
```diff
@@ -619,176 +638,285 @@ end
...
 end
 
-function downloadCoroutine()
-  while requestObj do
-    UI.setAttribute('download_progress', 'percentage', requestObj.download_progress * 100)
-    coroutine.yield(0)
+-- updates the preview window
+function updatePreviewWindow()
+  local item = library[contentToShow][currentListItem]
+
```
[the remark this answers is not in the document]
Chr1Z93: added an empty placeholder image until next image is loaded

Entrox-Licher (approved):
Looks good, let's get it in!

PR #412, opened 2023-10-15
Navigation Overlay: set max angle to 89

Chr1Z93:
This avoids looking at objects directly from above which can produce artifacts if they have badly aligned scripting buttons/fields.


Entrox-Licher:
Merge conflicts!  Yay!

Issue #404, opened 2023-10-11
Short Supply error messages


Chr1Z93:
Supposedly triggering Short Supply does sometimes show these error messages (but still executes as intended). Couldn't replicate this yet.

![image](https://github.com/argonui/SCED/assets/97286811/77e5d073-d57c-4316-a3c5-61b766d76edf)

Chr1Z93:
@Entrox-Licher I think we can close this for now sicne it should be fixed. If for some reason this gets reported again, we can re-open it.

PR #419, opened 2023-10-17
Bugfix for discarding from hands

Chr1Z93:
Since hands don't allow smooth moving objects out of them, this temporarily disables hands for the object.
Can test this via "discard random" button from the Hand Helper.

This should fix https://github.com/argonui/SCED/issues/404

### DIFF
--- a/src/playermat/Playmat.ttslua
+++ b/src/playermat/Playmat.ttslua
@@ -245,39 +245,37 @@ function placeOrMergeIntoDeck(obj, pos, rot)
   if not pos then return end
 
   local offset = 0.5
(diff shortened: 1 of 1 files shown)


Chr1Z93:
> Why does this fix #404?

I assume the error messages came from the multi-discard causing some objects merging by physics before the `putObject()` could grab them.

Entrox-Licher:
Hmm that could be it, but it still seems weird to me that it would propagate to all the playermats.  I think I rather leave #404 open until we get confirmation this fixes it.

Entrox-Licher (approved):

Issue #390, opened 2023-10-05
9+ Clue Spawning


Entrox-Licher:
When 9+ clues are spawned on a location, the location text is significantly obscured.  While several solutions were discussed, consensus seemed to be around stacking any clues beyond 8, and having clue counter stacks display their count on the top.  This way we maintain location card readability while still making it easy to see how many clues are on a location at a glance.

PR #424, opened 2023-10-18
Token Group Spawning: fixed shiftdown vector

Chr1Z93:
This fixes the spawning of multiple token groups. The shiftdown needs to be in the same rotation as the card.

Example:
image


Entrox-Licher:
I'd probably take advantage of this opportunity to tackle #390 if you're up for it

Chr1Z93:
> I'd probably take advantage of this opportunity to tackle #390 if you're up for it

This might be loosely related, but #390 is a much bigger change, so I'd rather resolve that separately.

Entrox-Licher (approved):

PR #428, opened 2023-10-19
Removal of obsolete placeholder boxes for community content

Chr1Z93:
Because of the new download menu, we don't need these placeholders anymore.


Entrox-Licher (commented):
Are the secret objectives/ultimatums in the download menu?

Chr1Z93:
> Are the secret objectives/ultimatums in the download menu?

Oh that's a very good catch. Since they aren't, how about putting them in the fan-made accessories?

Entrox-Licher:
Sounds good to me!

Chr1Z93:
> Sounds good to me!

Updated!

Entrox-Licher (approved):

Issue #417, opened 2023-10-17
Add Skill Timing Rules Reference to other side of "Detailed phase Reference" sheet


Jacklifear:
Story: 
The most common pages of the rules reference we check out is the "Skill Test Timing" in page 26-27. I realized that it would be rather neat to have that as a single page reference as well as a good quality of life improvement to the game.  There are existing cards that are lookup cards, but they work for people who are comfortable with the game and are written in the same fashion as an FAQ and is hard to understand for people who are learning the game.  

Image suggestion:
Page 26, swap left and right columns.  Take 2 parts on Page 27 and put it where the image is on 26 (Even as a opaque square above it even)

Image Placement: 
Could be added to the opposite side of the existing timing sheet, and have both of them side by side, flipped, just like the action reference cards. 
![Page 26 next to the Detailed Phase Reference](https://github.com/argonui/SCED/assets/10401883/ca4941bb-0ffe-43e2-adeb-9b28759eb8ae)
![Page 27, only need to pull the top left corner's data](https://github.com/argonui/SCED/assets/10401883/a46f12cc-0109-4407-b0a9-5964c937d026)

PR #430, opened 2023-10-19
updated rules reference to revised core version

Chr1Z93:
Since skill test timing is now on a single page, this closes https://github.com/argonui/SCED/issues/417


Entrox-Licher (approved):
Good update, but I do not think this solves the entire issue described here #417, since there is talk about having the pages beside the existing timing sheet.

Chr1Z93:
> Good update, but I do not think this solves the entire issue described here #417, since there is talk about having the pages beside the existing timing sheet.

I actually talked to them privately and this does satisfy their request :9

Issue #421, opened 2023-10-18
VP Highlighting Fails to Update


Entrox-Licher:
For the VP Tracker, if a location has all of its clues removed but then later receives another clue it will correctly update the number on the VP display, but the highlighting will stay on forever regardless of no longer being eligible for VP (even after updating/toggling VP highlighting)

PR #431, opened 2023-10-19
Victory Display: reset highlighted locations

Chr1Z93:
closes https://github.com/argonui/SCED/issues/421


Entrox-Licher:
Still not working, broken in a new way now.

Before, the location would toggle between on/off regardless of whether clues had been placed on it again.

Now, a VP location that has had clues replaced on it will _always_ stay highlighted, regardless of toggle

Entrox-Licher (changes requested):
Still not working, broken in a new way now.

Before, the location would toggle between on/off regardless of whether clues had been placed on it again.

Now, a VP location that has had clues replaced on it will always stay highlighted, regardless of toggle

Entrox-Licher (approved):
Works perfectly!

Issue #426, opened 2023-10-19
Clickable clues options should better hide text counter


Entrox-Licher:
At certain angles the text is still visible, just change slightly to hide better
![Screenshot 2023-10-19 111336](https://github.com/argonui/SCED/assets/8006667/482d3b62-4c7a-48c4-b436-10787093edea)

PR #437, opened 2023-10-21
Playermat Hiding!

Chr1Z93:
Adds a new tool, the "Playmat Hider". It sits in the fan-made accessories for now. Features a button for each playmat, which can toggle its visibility.
Also slightly edited the Y values for the clickable clue counters, which should fix https://github.com/argonui/SCED/issues/426

image


Entrox-Licher:
Buttons are still very clickable, as well as states modifiable when playmats are hidden.  It might be inevitable, but it would be worth pushing the y values down a bit further to see if it helps.

Chr1Z93:
> Buttons are still very clickable, as well as states modifiable when playmats are hidden.  It might be inevitable, but it would be worth pushing the y values down a bit further to see if it helps.

Unfortunately, this is inevitable since the table surface is "uninteractable". We could change that since that might be the lesser evil.

Entrox-Licher:
Aaah I see. Worth thinking about yeah, but beyond the scope of what's being done here, so I'll approve here and the rest can be discussed separately.

Entrox-Licher (approved):

Issue #436, opened 2023-10-20
Feature request: Hotkey Token spawning


dscarpac:
I really like these, either in the numpad or the programmable ones. But especially with the automatic connections and the deck searching built into TTS, might there be room for some other types of token spawning? I'd really like a damage counter with state 2 already on it, given that most weapons deal 2 damage.

PR #438, opened 2023-10-21
Numpad Keys: increase same counters instead if hovered

Chr1Z93:
- damage and horror key will set the hovered object to the next state
- resourceCounter key will add 1 to the hovered counter

closes https://github.com/argonui/SCED/issues/436


Entrox-Licher:
This is really cool!  What’s the deck cutter that’s getting cut?

Chr1Z93:
> This is really cool!  What’s the deck cutter that’s getting cut?

Damn, totally forgot to mention that in the description. It's an older tool that was triggered on num 1/2 and would take the top 3/6 cards from a deck. Since we now have way more search functions (the assistant and the built-in search X) it's obsolete.

Entrox-Licher (approved):

Issue #420, opened 2023-10-17
Norman Withers QoL


Chr1Z93:
Upkeep does respect the top card being flipped, but does not automatically flip the next card of the deck after drawing a flipped card from it.

PR #439, opened 2023-10-21
Norman Withers QoL // stateless deck area searching

Chr1Z93:
- added more QoL for Norman Withers: automatically flipping the top card after a face-up top card has been drawn
- refactor the deck area searching to be stateless: now returns obj references instead of setting global variables

closes https://github.com/argonui/SCED/issues/420


src/accessories/SearchAssistant.ttslua
```diff
@@ -124,59 +127,68 @@ function startSearch(messageColor, number)
...
+  if not isRightClick then
+    Wait.time(function()
+      local deckAreaObjects = playmatApi.getDeckAreaObjects(matColor)
+      if deckAreaObjects.draw then
+        deckAreaObjects.draw.shuffle()
+      end
+    end, #handCards * 0.1)
+  end
+
+  -- Norman Withers handling
```
Entrox-Licher, line 188: I think this should be added in to a be a Norman-specific check, as with others like Patrice.  I think players might leave cards face up for other reasons, and won't necessarily want this autoflipping to happen all the time, unless playing with Norman

Entrox-Licher (changes requested):

src/accessories/SearchAssistant.ttslua
Chr1Z93: Good point. Will add that as separate PR since [the](https://github.com/argonui/SCED/pull/458) uses this code too.

Entrox-Licher (approved):

Issue #414, opened 2023-10-17
Feature Request: Discard hotkey


Chr1Z93:
Just like the discard buttons on the playermats, this would move a card to the respective discard pile and move objects on it to the trashcan

Glitchy-Tozier:
If possible, it should also work on cards outside the playmats (e.g. on unengaged enemies, treacheries attached to locations or commited cards placed in front of other players' playmats).

Entrox-Licher:
This should mostly be possible, but on the note of skill cards on other playmats: there is no way to track which deck a card came from. You _can_ track who pressed the discard hotkey, and send the card to their discard pile though. But worth noting the difference

Chr1Z93:
I'm thinking that this should be the most convenient way:
- object is on playmat: trigger it for that playmat
- object is not on playmat: trigger it for the playmat of the player that pressed the hotkey

Entrox-Licher:
Well for what it's worth that _also_ doesn't work for the commiting cards to other players scenario

Glitchy-Tozier:
I like @Chr1Z93's proposal. Seems the most intuitive.
For the record: In my playgroups, we place commited cads in front of (instead of _on_) other player's playmats.

PR #440, opened 2023-10-21
Added discard hotkey & remove one use hotkey

Chr1Z93:
Discard hotkey:
This hotkey will behave like the "Discard" buttons on the playermats.
It will discard objects to the trash / discard pile of the nearest playermat (unless the object is not near a playmat, then it uses the triggering player).

closes https://github.com/argonui/SCED/issues/414

Remove one use hotkey:
Removes a single use from a card (matching the metadata). If a token is directly hovered, that will be discarded instead.

This also allows spending clues 1 by 1: https://github.com/argonui/SCED/issues/461


Chr1Z93:
Will update this to use the triggering player's discard if the hovered object is in the main playarea.

Chr1Z93:
Updated!

Entrox-Licher:
Not sure if we have a clean way of detecting if an object is on a playmat?  Because to me it would make more sense to have it such that:

If on playmat: hotkey discards to that playmat
If not on playmat: hotkey discards to player who pressed key

Not a big fan of the 'nearest playmat' standard, since it can create confusing scenarios in the places between playmats and the area where the acts & agendas go.

Chr1Z93:
> Not sure if we have a clean way of detecting if an object is on a playmat? Because to me it would make more sense to have it such that:
> 
> If on playmat: hotkey discards to that playmat If not on playmat: hotkey discards to player who pressed key
> 
> Not a big fan of the 'nearest playmat' standard, since it can create confusing scenarios in the places between playmats and the area where the acts & agendas go.

Just to be clear, the current behaviour is this:
1) object in playarea and card: discard to triggering player
2) object in playarea and not a card: discard to mythos trash
3) object outside of playarea: discard to nearest playmat

And you basically want 3 to be split into this:
3) object outside of playarea, but not near playmat: discard to triggering playmat
4) object outside of playarea, but near playmat: discard to nearest playmat

Entrox-Licher:
Excellent breakdown. There would probably need to be an agreement as far as what counts as ‘near’, but I think what I would like instead of the current behaviour as you describe is:

1. Object near playmat: discard to nearby playmat
2. Object not near playmat: discard to triggering playmat

I don’t see the value in differentiating by playarea, since encounter cards will be discarded correctly regardless, and I don’t think it’s particularly important which trash can things like clues or health tokens go to.

Chr1Z93:
> Excellent breakdown. There would probably need to be an agreement as far as what counts as ‘near’, but I think what I would like instead of the current behaviour as you describe is:
> 
> 1. Object near playmat: discard to nearby playmat
> 2. Object not near playmat: discard to triggering playmat
> 
> I don’t see the value in differentiating by playarea, since encounter cards will be discarded correctly regardless, and I don’t think it’s particularly important which trash can things like clues or health tokens go to.

That's a good point. I've removed the playarea differentiation and updated it according to the above conversation.

Entrox-Licher (approved):
Great solution

Issue #422, opened 2023-10-18
Feature Request: Discard tokens from cards entering hand / leaving play


dscarpac:
Example: Card with damage token on it returns to hand via Calling in Favors or Protecting the Anirniq, all tokens fall onto the table below the hand.

Could be fixed by sending tokens that fall onto this space to the trash (much like what happens with the discard pile).

Entrox-Licher:
Worthwhile idea, but not as easy as it seems to implement.  There's no great way (that I can think of) to track the tokens that fall beneath the hand, without adding a lot of needless complexity.

Will require some more thought to see if something clicks

dscarpac:
In the meantime I've just made a giant copy of the Token Remover and placed it underneath the hand, and that does work, although sending tokens to the Trash is more user friendly IMO.

Entrox-Licher:
The main issue with this is that @Chr1Z93 has been working on a feature to dynamically move playmats, and adding another layer of complexity to that I think would be undesirable

Chr1Z93:
I think we could just use handzones for this detection. If they are properly sized, tokens should never enter the hand zone on accident. So then we can just move entering objects (that are tokens) into the player's trash.

PR #441, opened 2023-10-22
implemented token removal below hand zone

Chr1Z93:
Ever moved a card back to your hand that had tokens on it? (like Calling in Favors)
Tokens would fall to the ground below the hand zone and stay there...

Now this PR adds ascripting zone t to detect these tokens and put them into your trash.
Also features an equalization of the hand zone positions / sizes.

Token Remover is now obsolete and was thus removed.

closes https://github.com/argonui/SCED/issues/422


Chr1Z93:
Noticed during playtesting, that the token remover is not actually obsolete.
At the very least you still need to it for locations that leave play (and still have clues).
Will restore it!

src/core/Global.ttslua
```diff
@@ -186,6 +186,19 @@ function tryObjectEnterContainer(container, object)
   return true
 end
 
+-- TTS event for objects that enter zones
+-- used to detect the "token discard zones" beneath the hand zones
+function onObjectEnterZone(zone, enteringObj)
+  if zone.getName() ~= "TokenDiscardZone" then return end
+  if tokenChecker.isChaosToken(enteringObj) then return end
+  
+  if enteringObj.type == "Tile" and enteringObj.getMemo() and enteringObj.getLock() == false then
```
Entrox-Licher, line 195: Not strictly related to this PR, but I wonder if the memo field is the best way to track tokens like this?  What's the rationale to using it instead of a tag and the GMNotes?

Chr1Z93: Historically, the memo field was implemented for tokens to discern them from objects with gmnotes since a lot of code assume the gmnotes to be card metadata

Entrox-Licher: Hmm interesting.  I would have thought it shouldn't matter much since tokens should never really have metadata.  Moreover, I don't love the idea of checking a token or card is of a certain type just by the mere _presence_ of gm notes or memos (restricts a lot of the versatility of those fields).

To me a tag is the cleaner way to go.  That being said, not really germane to this PR though.  I'll approve and maybe open an issue to have this conversation

Entrox-Licher (approved):

PR #442, opened 2023-10-22
Option Panel: moved descriptions to tooltips

Chr1Z93:
This is easier on the eyes since it doesn't put so much text on the user at once:

image


argonui (approved):

Entrox-Licher (commented):
Are we sure we want this change? I think this is a step down, much preferred it the way before where everything was easy to parse at a glance

PR #452, opened 2023-10-30
Added parallel Zoey

Chr1Z93:


src/arkhamdb/ArkhamDb.ttslua
```diff
@@ -198,7 +198,7 @@ do
       local altArt     = { front = "normal", back = "normal" }
 
       -- translating front ID
-      if altFrontId > 90000 and altFrontId < 90050 then
```
argonui, line 201: This is a pretty serious nightmare of maintenance. Are there other ways to determine parallel or not?

argonui (dismissed):

objects/AllPlayerCards.15bb07/ZoeySamarasParallel.98a0e2.json
```diff
@@ -0,0 +1,62 @@
...
+    "posZ": -73.088,
+    "rotX": 0,
+    "rotY": 180,
+    "rotZ": 0,
+    "scaleX": 1.15,
+    "scaleY": 1,
+    "scaleZ": 1.15
+  },
+  "Value": 0,
+  "XmlUI": ""
```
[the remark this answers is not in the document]
Chr1Z93: good catch, fixed

src/arkhamdb/ArkhamDb.ttslua
Chr1Z93: Very valid concern and this was indeed a pain in the ass to implement, but unfortunately the ArkhamDB Api just provides a single altId. We have to use this to discern whether this means it's a parallel investigator, revised core alt art or promo art.

Entrox-Licher (approved):
Valid maintainability concerns raised, but as far as the new cards themselves, they look good to me

PR #456, opened 2023-11-04
Unified custom content spawn position

Chr1Z93:
Custom content (that doesn't have a replacement box on the table) will now spawn directly below the tentacle stand in the first empty position (from left to right, max. 3 rows).

image


Entrox-Licher (changes requested):
There should probably be a consistent behaviour defined for what happens when all 3 rows are filled (like a broadcast that there is no room?), rather than just defaulting back to the random spawn positions of each given box

Chr1Z93:
> There should probably be a consistent behaviour defined for what happens when all 3 rows are filled (like a broadcast that there is no room?), rather than just defaulting back to the random spawn positions of each given box

Fair point! Very unlikely, but I'll add a broadcast

Entrox-Licher:
I think it would be better to just block spawning altogether if full.  It's unlikely enough to happen that the inconvenience isn't huge, and it provides a consistent behaviour, rather than the boxes spawning all over the table and potentially falling into other bags, leading to more confusion.

Entrox-Licher (approved):

PR #458, opened 2023-11-04
Added scripting to Scroll of Secrets to draw bottom cards of a deck

Chr1Z93:
This adds a context menu entry to Scroll of Secrets to draw the bottom card of a deck.
Drawing from the top and returning cards to the top / bottom will stay unscripted.

Note: This requires https://github.com/argonui/SCED/pull/439 since it shares code for deck detection.


src/playercards/cards/ScrollofSecrets.ttslua
```diff
@@ -0,0 +1,92 @@
...
+    createContextMenu(notes.class)
+  else
+    print("Missing metadata for Scroll of Secrets!")
+  end
+end
+
+function createContextMenu(class)
+  local colorList = Player.getAvailableColors()
+
+  if class == "Seeker|Mystic" then
```
Entrox-Licher, line 18: This should probably be checking IDs rather than classes

src/playercards/cards/ScrollofSecrets.ttslua
```diff
@@ -0,0 +1,92 @@
...
+
+  -- show the target selection dialog
+  Player[playerColor].showOptionsDialog("Select target deck", options, _, 
+    function(owner)
+      drawCardsFromBottom(playerColor, owner, amount)
+    end
+  )
+end
+
+function drawCardsFromBottom(playerColor, owner, amount)
```
Entrox-Licher, line 43: Are we sure we want to be _drawing_ the cards?  Since this is a look operation that can result in the card going to any of 4 places it seems the least likely place for a card to end up is the hand of the player activating the ability.  I think this may be better envisioned as a camera refocus and let players manually draw the card and choose where it goes.  But honestly at that point, I question if it's even worth adding the feature at all...?

Entrox-Licher (dismissed):

src/playercards/cards/ScrollofSecrets.ttslua
Chr1Z93: We could add a broadcast statement to make this clear, but indeed the shoe point is simplifying drawing cards from the bottom.

src/playercards/cards/ScrollofSecrets.ttslua
Chr1Z93: Fair point. Updated

src/playercards/cards/ScrollofSecrets.ttslua
Chr1Z93: Added an additional print instruction on how to proceed with the drawn cards.

BootleggerFinn (approved):

PR #462, opened 2023-11-05
Bugfix for Family Inheritance: stopped overriding resources

Chr1Z93:
Previously, the call to the playmatApi had the modifier at the wrong position and thus that would be used as new value instead of modifier.

### DIFF
--- a/src/playercards/cards/FamilyInheritance.ttslua
+++ b/src/playercards/cards/FamilyInheritance.ttslua
@@ -1,5 +1,5 @@
-local tokenManager = require("core/token/TokenManager")
 local playmatApi = require("playermat/PlaymatApi")
+local tokenManager = require("core/token/TokenManager")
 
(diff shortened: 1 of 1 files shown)


src/playercards/cards/FamilyInheritance.ttslua
```diff
@@ -1,5 +1,5 @@
-local tokenManager = require("core/token/TokenManager")
 local playmatApi = require("playermat/PlaymatApi")
+local tokenManager = require("core/token/TokenManager")
```
argonui, line 2: Does this reordering matter?

Chr1Z93: Functionally, it does not. We decided on ordering these requires alphabetically (already the case for all main script files) and I just noticed that they were unordered in this one while fixing the bug.

PR #464, opened 2023-11-07
Added uses spawning for Makeshift Trap and Shrine of the Moirai in the playarea

Chr1Z93:


argonui (commented):
How do you know how many tokens to spawn?

Chr1Z93:
The cards already do have the metadata in their GM_Notes for spawning uses. This just lets them pass the condition for the spawn process on the playarea too (instead of just on the playermats).

Entrox-Licher (approved):

PR #466, opened 2023-11-08
Update note changes

BootleggerFinn:
Put in the changes/reviews to the update notes, mostly to highlight the biggest "features" first and to match the formatting of previous notes. As I noted in the first commit's notes, it would have been nice to include a special thanks callout, but there was no room for it without just making a third notecard, which didn't seem right. I set the included release date to November 9th since that seemed the most convenient.


Chr1Z93:
Thanks for the editing! Looks good to me

Chr1Z93 (approved):

PR #473, opened 2023-11-18
Added better mod version comparison

Chr1Z93:
The update notification will now only show if you are on an outdated version instead of also displaying when you are on a newer version (e.g. dev).


BootleggerFinn (approved):
Nice change!

PR #485, opened 2023-11-23
Split the custom content from the "download all" button into separate bags

tarogers:
One bag is created for each contentType, in a column on the right side of the table


Chr1Z93 (approved):
Clever implementation of variables in the JSON string!

PR #499, opened 2023-12-07
Wait time between cards for Short Supply

dscarpac:
Allows users to see which cards were discarded without having to search through Discard pile.


Chr1Z93:
> Allows users to see which cards were discarded without having to search through Discard pile.

Neat idea. I just tested it and it feels mostly good. I'd personally remove the delay for the first card and increase the delay slightly: `delay = 0.7 * (i-1`

Chr1Z93 (approved):
LGTM!

PR #506, opened 2023-12-15
Added startsInPlay metadata to more cards

Entrox-Licher:


Chr1Z93 (changes requested):
The data looks good to me! Now we need a handling for multiple cards getting placed in the "BlankTop" zone (see function `getDefaultCardZone` in `DeckImporterMain`). I'd go for this: Define the other 5 blank spots in the Zones file and then split the cards to the spots. So that if you have like 4 cards in the "BlankTop" (should be renamend) zone, that the upper 2 and lower 2 slots each get one card. For the edge case that you actually have more than 6 cards in your deck that want to go directly on the playmat, we should probably return the 7th to nth card to the regular permanent pile and broadcast a notification.

Chr1Z93 (approved):
Really cool solution! Looks good to me

PR #521, opened 2023-12-31
More Hemlock cards!

Chr1Z93:


objects/AllPlayerCards.15bb07/CatMask.847ed6.gmnotes
```diff
@@ -0,0 +1,11 @@
...
+  "id": "10084",
+  "type": "Asset",
+  "class": "Mystic",
+  "cost": 1,
+  "level": 0,
+  "traits": "Item. Charm. Mask.",
+  "combatIcons": 1,
+  "willpowerIcons": 1,
+  "cycle": "The Feast of Hemlock Vale"
+}
```
Entrox-Licher, line 11: Missing Uses (offerings)

Entrox-Licher (changes requested):
Looks good!  Only missing one minor change regarding the Cat Mask (and thinking about how we want to handle cost metadata for X cost cards?)

Entrox-Licher (approved):

PR #524, opened 2024-01-01
Turned playermat hider into a dropdown option

Chr1Z93:
Need to keep an open eye for any issues that stem from a removed mat.
Respawning a mat might eventually be added.

- updated GUID Reference Handler to allow edits to the index
- turned playermat hider into an option

closes https://github.com/argonui/SCED/issues/536


Entrox-Licher (changes requested):
image
Produces error on startup.  Most likely an issue with script state?

Chr1Z93:
> image Produces error on startup. Most likely an issue with script state?

I can't replicate that. The interceptor should be fully gone. How did you get that error?

Entrox-Licher:
It's gone now!  Looks like the main merge fixed it?  (Or maybe I did something weird with Git haha?)

Entrox-Licher (changes requested):
- Option should probably be under Playermat Settings
- Might want to change button (maybe even delete it?) once delete operation has been performed?

Entrox-Licher (approved):
Awesome!

PR #533, opened 2024-01-08
Campaign Exporter Improvements

Entrox-Licher:
Added support for Additional Player Cards bag (still needs latency improvements), custom trauma getting, importing on collision, and fixed the tour bug


src/accessories/CampaignImporterExporter.ttslua
```diff
@@ -75,44 +64,43 @@ end
...
+  local objectData = info.collision_object.getData()
+  if info.collision_object.hasTag("ImporterToken") then
+    findCampaignFromToken(info.collision_object)
+  end
+end
 
-    local importData = JSON.decode(coin.getGMNotes())
-    campaignBoxGUID = importData["box"]
+-- Identifies import token, determines campaign box and downloads it (if needed)
+function findCampaignFromToken(coin)
```
Chr1Z93, line 75: I'd give this function a better fitting name. Something like `ImportFromToken()`?

src/accessories/CampaignImporterExporter.ttslua
```diff
@@ -75,44 +64,43 @@ end
...
-        if obj == nil then
-          return false
-        else
-          return obj.type == "Bag" and obj.getLuaScript() ~= ""
-        end
-      end,
-      2,
-      function() broadcastToAll("Error loading campaign box") end
-    )
+  campaignBoxGUID = importData["box"]
```
Chr1Z93, line 80: Avoid global variable

src/accessories/CampaignImporterExporter.ttslua
```diff
@@ -224,16 +216,21 @@ function createCampaignToken(_, playerColor, _)
...
+
   local campaignData = {
     box = campaignBox.getGUID(),
     log = campaignLog.getData(),
     bag = chaosBagApi.getChaosBagState(),
     trauma = traumaValues,
     decks = deckImporterApi.getUiState(),
     clueCount = playAreaApi.getInvestigatorCount(),
-    guide = campaignGuide.Book.getPage(),
+    playmat = playAreaApi.getSurface(),
```
Chr1Z93, line 230: This should not be named playmat (because it could be confused with the actual playermats)

src/accessories/CampaignImporterExporter.ttslua
```diff
@@ -202,20 +194,20 @@ function createCampaignToken(_, playerColor, _)
...
-      traumaValues[1 + (i - 1) / 3] = counterData.counter[i].value
-      traumaValues[5 + (i - 1) / 3] = counterData.counter[i + 1].value
+    trauma = campaignLog.call("returnTrauma")
+    for _, val in ipairs(trauma) do
+      table.insert(traumaValues, val)
     end
   else
     printToAll("Trauma values could not be found in campaign log!", "Yellow")
-    printToAll("Default values for health and sanity loaded.", "Yellow")
+    return
```
Chr1Z93, line 210: Why is there a "return" here? Do you not want to finish the import?

Chr1Z93 (changes requested):

src/accessories/CampaignImporterExporter.ttslua
```diff
@@ -47,23 +47,12 @@ local campaignTokenData = {
...
-    scale = { 0.5, 1, 0.5 },
-  })
   self.createButton({
     click_function = "createCampaignToken",
     function_owner = self,
     label = "Export",
-    tooltip = "Create a campaign save token!\n\n(Ensure all chaos tokens have been unsealed!)",
-    position = { x = 1, y = 0.2, z = 0 },
+    tooltip = "Create a campaign save token!",
+    position = { x = -1, y = 0.2, z = 0 },
```
Chr1Z93, line 55: Nitpick: The "y"-value should be 0.11 since the thickness is 0.2, but this is calculated from the center. (0.10 will cause flickering since the asset and the button will "fight" for the same position)

Entrox-Licher: Assuming you meant .21, but yes!

Entrox-Licher:
This should be ready now!  Any tips on how to implement the coroutine we discussed to reduce freezing/stuttering?

src/accessories/CampaignImporterExporter.ttslua
Chr1Z93: Nope, I did mean 0.11 😅

Chr1Z93:
> This should be ready now! Any tips on how to implement the coroutine we discussed to reduce freezing/stuttering?

Is this (https://github.com/argonui/SCED/pull/533/files#diff-f105a0d8fbf00607c87d8f42b33a2c34b6e4c4d13b25bbbf9be9715b13ce4fbcR232) the only line that introduces freezing?
If so, then we can't really help it. If not, then you could use a coroutine to gather all the data (with some delays between each step).

src/accessories/CampaignImporterExporter.ttslua
```diff
@@ -47,91 +51,120 @@ local campaignTokenData = {
...
+    end
+  end
+
   -- destroy existing campaign log and load saved campaign log
   findUniqueObjectWithTag("CampaignLog").destruct()
-  spawnObjectData({ data = importData["log"] })
-  
+  if coin.type == "Bag" then
+    local newLog = coin.takeObject({index = 0, position = importData["log"]})
+    Wait.time(function() newLog.setLock(true) end, 1)
```
Chr1Z93, line 160: This should be a callback function to the `takeObject` function above

src/accessories/CampaignImporterExporter.ttslua
```diff
@@ -47,91 +51,120 @@ local campaignTokenData = {
...
 -- After content is placed on table, conducts all the other import operations
-function createCampaignFromToken(importData)
+function createCampaignFromToken(importData, coin)
+
+  checkWarning = false
+  if importData["additionalIndex"] then
+    findUniqueObjectWithTag("AllCardsHotfix").destruct()
+    if coin.type == "Bag" then
+      local newBag = coin.takeObject({index = 0, position = importData["additionalIndex"]})
+      Wait.time(function() newBag.setLock(true) end, 1)
```
Chr1Z93, line 150: This should be a callback function to the `takeObject` function above

src/accessories/CampaignImporterExporter.ttslua
```diff
@@ -47,91 +51,120 @@ local campaignTokenData = {
...
     2,
     function() broadcastToAll("Error placing campaign box") end
   )
 end
 
 -- After content is placed on table, conducts all the other import operations
-function createCampaignFromToken(importData)
+function createCampaignFromToken(importData, coin)
+
+  checkWarning = false
```
Chr1Z93, line 145: Not a big fan of the frequent enabling / disabling of this

src/accessories/CampaignImporterExporter.ttslua
```diff
@@ -196,26 +228,25 @@ function createCampaignToken(_, playerColor, _)
...
     return
   end
 
-  local traumaValues = {
-    0, 0, 0, 0,
-    0, 0, 0, 0
-  }
-  local counterData = campaignLog.getVar("ref_buttonData")
-  if counterData ~= nil then
+  additionalIndex = findUniqueObjectWithTag("AllCardsHotfix")
```
Chr1Z93, line 237: There can technically be multiple objects with this tag (keep in mind the original usecase for this tag was correcting / adding cards).

src/accessories/CampaignImporterExporter.ttslua
```diff
@@ -224,20 +255,38 @@ function createCampaignToken(_, playerColor, _)
...
+  local logData = campaignLog.getData()
+  local indexData = additionalIndex.getData()
+  campaignLog.setLock(false)
+  additionalIndex.setLock(false)
+  checkWarning = false
+  obj.putObject(additionalIndex)
+  obj.putObject(campaignLog)
+  checkWarning = true
+  spawnObjectData({data = indexData})
+  spawnObjectData({data = logData})
```
Chr1Z93, line 289: Instead of removing the objects and then respawning them, it would be better to use `getData()` and then add that data to the `ContainedObjects` of the coin

Chr1Z93 (changes requested):

src/accessories/CampaignImporterExporter.ttslua
```diff
@@ -47,91 +51,120 @@ local campaignTokenData = {
...
+end
+
+function onObjectEnterContainer(container, object)
+  if container.hasTag("ImporterToken") and checkWarning then
+    broadcastToAll(
+        "Adding objects to the Save Coin bag will break functionality.  Please remove the objects.",
+        Color.Yellow
+    )
+  end
+end
```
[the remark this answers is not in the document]
Entrox-Licher: Yes!  The importing code relies on objects being in an exact order.  Technically I could add some longer checks to look for specific tags or object names, but this seemed simpler

src/accessories/CampaignImporterExporter.ttslua
```diff
@@ -47,91 +51,120 @@ local campaignTokenData = {
...
-    scale = { 0.5, 1, 0.5 },
-  })
   self.createButton({
     click_function = "createCampaignToken",
     function_owner = self,
     label = "Export",
-    tooltip = "Create a campaign save token!\n\n(Ensure all chaos tokens have been unsealed!)",
-    position = { x = 1, y = 0.2, z = 0 },
+    tooltip = "Create a campaign save token!",
+    position = { x = -1, y = 0.21, z = 0 },
```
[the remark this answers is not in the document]
Entrox-Licher: Nope!  This causes the button to be inside the object

src/accessories/CampaignImporterExporter.ttslua
```diff
@@ -47,91 +51,120 @@ local campaignTokenData = {
...
 end
 
+function onObjectLeaveContainer(container, object)
+  if container.hasTag("ImporterToken") and checkWarning then
+    broadcastToAll(
+        "Removing objects from the Save Coin bag will break functionality.  Please replace the objects in the same order they were removed.",
+        Color.Yellow
+    )
+  end
+end
```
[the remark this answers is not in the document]
Entrox-Licher: Maybe?  I just don't think the odds are very high that anyone will edit it outside of the main save, and I prefer to have all relevant code in one place for ease of editing/updating.

(Yellow doesn't work! Tried it, it comes out as just white, has to be Color.Yellow)

src/accessories/CampaignImporterExporter.ttslua
```diff
@@ -47,91 +51,120 @@ local campaignTokenData = {
...
-  local coin = nil
-  local coinObjects = getObjectsWithTag("ImporterToken")
-  if #coinObjects == 0 then
-    broadcastToAll("Could not find importer token", Color.Red)
-  elseif #coinObjects > 1 then
-    broadcastToAll("More than 1 importer token found. Please delete all but 1 importer token", Color.Yellow)
-  else
-    coin = coinObjects[1]
+function onCollisionEnter(info)
+  local objectData = info.collision_object.getData()
```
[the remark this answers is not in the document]
Entrox-Licher: Yup!  Good catch!

src/accessories/CampaignImporterExporter.ttslua
```diff
@@ -47,91 +51,120 @@ local campaignTokenData = {
...
+    function() createCampaignFromToken(importData, coin) end,
     function() return findUniqueObjectWithTag("CampaignLog") ~= nil end,
     2,
     function() broadcastToAll("Error placing campaign box") end
   )
 end
 
 -- After content is placed on table, conducts all the other import operations
-function createCampaignFromToken(importData)
+function createCampaignFromToken(importData, coin)
```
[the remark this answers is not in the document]
Entrox-Licher: Love it, agree, the name wasn't great haha

src/accessories/CampaignImporterExporter.ttslua
Entrox-Licher: It's not really frequent though?  It only happens when importing or exporting

src/accessories/CampaignImporterExporter.ttslua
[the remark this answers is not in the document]
Chr1Z93: Alright let's just ignore this then

src/accessories/CampaignImporterExporter.ttslua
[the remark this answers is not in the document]
Chr1Z93: Ah, I guess the short strings only work for native colors (of the seats) in TTS. TIL!

Keeping the code here is fine.

src/accessories/CampaignImporterExporter.ttslua
Chr1Z93: Yeah, it's at multiple points though. Why not just disable it at the start and keep it disabled for the whole process instead of toggling it?

Entrox-Licher: ??? It's only disabled once (on this line), then reenabled shortly thereafter.  Don't see the issue?

Chr1Z93 (approved):

src/accessories/CampaignImporterExporter.ttslua
Chr1Z93: Okay I must have been looking at an old version when I made that comment? Or I was just sleepy.. Either way this works nicely

Issue #534, opened 2024-01-09
Request: Multihanded gameplay improvements


Chr1Z93:
- add hotkey to swap seats (maybe clockwise and counter-clockwise around the table?)
- add handling for card drawing via number typing to correct hand for player decks

PR #537, opened 2024-01-09
Multihand improvements

Chr1Z93:
Closes https://github.com/argonui/SCED/issues/534

### DIFF
--- a/src/core/GameKeyHandler.ttslua
+++ b/src/core/GameKeyHandler.ttslua
@@ -1,5 +1,6 @@
 local blessCurseManagerApi = require("chaosbag/BlessCurseManagerApi")
 local guidReferenceApi = require("core/GUIDReferenceApi")
+local navigationOverlayApi = require("core/NavigationOverlayApi")
 local optionPanelApi = require("core/OptionPanelApi")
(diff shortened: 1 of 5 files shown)


Chr1Z93:
> Fantastic changes!
> 
> I'd recommend making it so the hotkey also adjusts your camera angle as done by the Nav Overlay

While that would be possible, I think multihanded gameplay will in most cases be done with the white and orange seat. When using these new hotkeys to swap seats, TTS loads the camera 0 by default. If this camera is just manually set to cover both seats (and then kept mostly static), that feels better to me than zooming in on each mat.

Keep in mind that people might be sitting in one seat for most rounds and then just swap to use the clue hotkey and immediately swap back 😅

Entrox-Licher:
All good points, but my main concern is it behaving inconsistently to the 'main' way we have of supporting multi-handed solo.  Shouldn't we want both of these to behave the same?

Chr1Z93:
> All good points, but my main concern is it behaving inconsistently to the 'main' way we have of supporting multi-handed solo. Shouldn't we want both of these to behave the same?

You do have a point. Updated it

Entrox-Licher (approved):

Issue #516, opened 2023-12-20
Improve Bless/Curse Manager


Entrox-Licher:
The Bless/Curse Manager has several areas that I feel could be improved upon, and given the upcoming influx of bless/curse cards, it will be good to tackle these before Feast releases (on the mod that is).

- Tracking of tokens is handled manually through the buttons.  This can lead to situations where the manager thinks there are different amounts of tokens in play than there actually are.  I'd recommend tracking this automatically by tracking when tokens enter/exit the chaos bag or the individual token storage bags (the ones next to the chaos bag, which also aren't being used by the manager)
- Sealing could be improved.  Currently for cards that don't have context menu support, the Take/Return option exists, but this works entirely sequentially, which will often result in returning the wrong token when returning (especially when sealing tokens to enemies or other scenario cards).  I suggest adding context menu support to the tokens themselves, allowing the ability to right click and return them to the bag or supply, and properly update the Manager accordingly.  This could also replace large chunks of the `CardsThatSealsTokens` logic too.  Alternatively, having events in place so the player can manually deal with the tokens (returning to player bag or supply or deleted or put into trash).  Most of this would be handled automatically if the first bullet were implemented.
- Tracking Bless/Curse tokens in play is a bit cumbersome.  It is good we have a GameKey for it, but with how little information is needed to be displayed, I don't think it would be hard to add that as a small section to the Manager itself, so it is always visible at a glance.
- Finally, the Manager is located in a really awkward position, away from all other objects that interact with the chaos bag.  I would recommend moving it either up near the chaos bag itself or near the Token Arranger & Chaos Bag Manager

These are all the main issues that come to mind at the moment, I'll update this issue thread with more if they occur to me.  Cheers!

Chr1Z93:
1) I don't really understand the first point, except that there is not tracking of manually edits to the chaos bag. My takeaway would be to remove the additional bless / curse containers since they are obsolete due to the manager and it's so long in use that it's highly likely to not break.

2) I agree that a context menu to return a token directly on the tokens would be useful. The "Take/Return" option is mostly for legacy support - I don't see any reason to improve this since no one should be using this at all.

3) Very good point! I think we could just have a small XML UI that is either always visible or visible onHover of the manager.

4) I agree that the position is kind of awkward, but the main advantage is that this is central. I don't like moving it outside of the main playarea.

Glitchy-Tozier:
[...]

PR #538, opened 2024-01-09
Update to blurse manager

Chr1Z93:
- removed separate bless / curse tokens (bless / curse manager does all the handling - chaos token reserve still exists for "emergencies")
- removed the "take / remove" button since all cards have proper context menus ("Wendy Menu" exists for emeergencies)
- add token count labels
- added missing "Return" option to Nephthys
- renamed Wendy's menu
- added tracking for manually returned tokens
- added releasing when discarding cards
- some code cleanup


Entrox-Licher:
Excellent changes!  They are all working as intended.

This seems like a good opportunity to finish addressing the issues mentioned in #516 . Namely:

1. To add context menu support to the individual tokens for releasing seals.  This is essential as manual edits to the bag (be it manually removing or manually returning) full break this tool.  I still think tracking the tokens dynamically is best to avoid this edge case, but in absence of that (which would require a bigger rework), we should remove as many instances as possible that would prompt a user to manually interact with the bag, and this is the top offender

2. Make Wendy's Menu more obvious to users, for the same reason as (1).  Likely a good idea to rename it so it's function is more obvious (there are more corner cases than just Parallel Wendy), and definitely mention it explicitly in the manager's description.

3. If we are going to change the position of it, now would be the time to do that too (not necessary if we are opting to keep it where it is).  Maybe a copy of it below the chaos bag could make sense as best of both worlds talked about in the issue?

Chr1Z93:
> Excellent changes! They are all working as intended.
> 
> This seems like a good opportunity to finish addressing the issues mentioned in #516 . Namely:
> 
> 1. To add context menu support to the individual tokens for releasing seals.  This is essential as manual edits to the bag (be it manually removing or manually returning) full break this tool.  I still think tracking the tokens dynamically is best to avoid this edge case, but in absence of that (which would require a bigger rework), we should remove as many instances as possible that would prompt a user to manually interact with the bag, and this is the top offender
> 2. Make Wendy's Menu more obvious to users, for the same reason as (1).  Likely a good idea to rename it so it's function is more obvious (there are more corner cases than just Parallel Wendy), and definitely mention it explicitly in the manager's description.
> 3. If we are going to change the position of it, now would be the time to do that too (not necessary if we are opting to keep it where it is).  Maybe a copy of it below the chaos bag could make sense as best of both worlds talked about in the issue?

1. I don't think this is as urgent as you make it out to be. There won't be any "full breaking", since you can use the existing context menu function to reset it's state (and dynamically get where tokens are at the moment).

2. For which cases would you need it outside of Parallel Wendy? Using custom card without scripting?

3. I'm still in the XML camp for this and might eventually have a small global window for it. Supporting multiple copies would require a bunch of code changes.

Entrox-Licher:
> > Excellent changes! They are all working as intended.
> > This seems like a good opportunity to finish addressing the issues mentioned in #516 . Namely:
> > 
> > 1. To add context menu support to the individual tokens for releasing seals.  This is essential as manual edits to the bag (be it manually removing or manually returning) full break this tool.  I still think tracking the tokens dynamically is best to avoid this edge case, but in absence of that (which would require a bigger rework), we should remove as many instances as possible that would prompt a user to manually interact with the bag, and this is the top offender
> > 2. Make Wendy's Menu more obvious to users, for the same reason as (1).  Likely a good idea to rename it so it's function is more obvious (there are more corner cases than just Parallel Wendy), and definitely mention it explicitly in the manager's description.
> > 3. If we are going to change the position of it, now would be the time to do that too (not necessary if we are opting to keep it where it is).  Maybe a copy of it below the chaos bag could make sense as best of both worlds talked about in the issue?
> 
> 1. I don't think this is as urgent as you make it out to be. There won't be any "full breaking", since you can use the existing context menu function to reset it's state (and dynamically get where tokens are at the moment).
> 2. For which cases would you need it outside of Parallel Wendy? Using custom card without scripting?
> 3. I'm still in the XML camp for this and might eventually have a small global window for it. Supporting multiple copies would require a bunch of code changes.

1. I'll admit the reset function is more robust than I originally thought, with the inclusion of the dynamic token detection.  There are still a lot of ways I can get it to break by just manually adding or removing tokens though (requiring multiple 'resets' in order to restore equilibrium, which is not very intuitive).  Removing all tokens does work in all cases I was able to test, but obviously this is somewhat of a nuclear option in actual play, where tokens will be sealed on all sorts of different cards.  I still think it's worth adding the functionality, since it is very likely to come up in play (Nephthys, Favor of the Sun/Moon, etc). 
2.  Custom cards are in fact the biggest scenario I was thinking of.  In general it is a good fallback option though, in order to prevent cards we may miss or support scenarios that could have similar effects.
3. I agree with this logic!  Let's keep it as is and we can revisit later if we do anything fancy XML-wise for the chaos bag.

4. Another bug I discovered: Wendy's Menu interacts very weirdly with cards that already have sealing options.  We should either make compatibility for that or (preferable I think) disable the interaction altogether (check for tag before adding Wendy's Menu)

Entrox-Licher (approved):

PR #544, opened 2024-01-11
added button to reshuffle encounter discard pile into deck

dscarpac:
XML button provides functionality

### DIFF
--- a/src/core/MythosArea.ttslua
+++ b/src/core/MythosArea.ttslua
@@ -201,6 +201,27 @@ function reshuffleEncounterDeck(params)
   Wait.time(function() isReshuffling = false end, 1)
 end
 
+function discardIntoDeck()
+  --flag to avoid multiple calls
+  if isReshuffling then return end
+  isReshuffling = true
+
+  -- flip deck and then shuffle
+  local discardPile = searchLib.atPosition(ENCOUNTER_DISCARD_POSITION, "isCardOrDeck")
+  if #discardPile > 0 then
+    local deck = discardPile[1]
+    if not deck.is_face_down then deck.flip() end
+    deck.setPositionSmooth(Vector(ENCOUNTER_DECK_POS) + Vector(0, 2, 0), false, true)
+    --wait until deck is in place .75 seconds, then shuffle
+    broadcastToAll("Shuffled encounter discard into deck.", "White")
+    Wait.time(function() searchLib.atPosition(ENCOUNTER_DECK_POS, "isCardOrDeck")[1].shuffle() end, .75)
+  else
+    broadcastToAll("Encounter discard pile is already empty.", "Red")
+  end
+
+  -- disable flag
+  Wait.time(function() isReshuffling = false end, 1)
+end
 ---------------------------------------------------------
 -- helper functions
 ---------------------------------------------------------
--- a/src/core/MythosArea.ttslua
+++ b/src/core/MythosArea.ttslua
@@ -156,7 +156,6 @@ function drawEncounterCard(params)
     -- nothing here, time to reshuffle
     if reshuffledAlready == true then
       reshuffledAlready = false
-      -- print something?
       return
     end
     reshuffleEncounterDeck() -- if there is no discard pile either, reshuffleEncounterDeck will give an error message already
@@ -182,7 +181,7 @@ function actualEncounterCardDraw(card, params)
 end
 
 function reshuffleEncounterDeck()
-  --flag to avoid multiple calls
+  -- flag to avoid multiple calls
   if isReshuffling then return end
   isReshuffling = true
   local encounterDeck = getEncounterDeck()
@@ -190,7 +189,9 @@ function reshuffleEncounterDeck()
     
   if #discardPile > 0 then
     local discardDeck = discardPile[1]
-    if not discardDeck.is_face_down then discardDeck.setRotation({0, -90, 180}) end --flips discard pile
+    if not discardDeck.is_face_down then --flips discard pile
+      discardDeck.setRotation({0, -90, 180}) 
+    end 
     if encounterDeck == nil then
       discardDeck.setPosition(Vector(ENCOUNTER_DECK_POS) + Vector({0, 1, 0}))
       discardDeck.shuffle()
@@ -198,7 +199,7 @@ function reshuffleEncounterDeck()
       encounterDeck.putObject(discardDeck)
       encounterDeck.shuffle()
     end
-      broadcastToAll("Shuffled encounter discard into deck.", "White")
+    broadcastToAll("Shuffled encounter discard into deck.", "White")
   else
     broadcastToAll("Encounter discard pile is already empty.", "Red")
   end
--- a/src/core/MythosArea.ttslua
+++ b/src/core/MythosArea.ttslua
@@ -134,13 +134,6 @@ end
 function getEncounterDeck()
   local searchResult = searchLib.atPosition(ENCOUNTER_DECK_POS, "isCardOrDeck")
 
-  for _, obj in ipairs(searchResult) do
-    if obj.type == 'Deck' then
-      return obj
-    end
-  end
-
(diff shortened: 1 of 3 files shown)


src/core/MythosArea.ttslua
```diff
@@ -201,6 +201,27 @@ function reshuffleEncounterDeck(params)
   Wait.time(function() isReshuffling = false end, 1)
 end
 
+function discardIntoDeck()
+  --flag to avoid multiple calls
+  if isReshuffling then return end
+  isReshuffling = true
+
+  -- flip deck and then shuffle
```
Entrox-Licher, line 209: You can just combine the deck objects directly using the `putObject` [function](https://api.tabletopsimulator.com/object/#putobject).  It should drastically simplify the logic here.

Entrox-Licher (dismissed):
Should clean up the function a bit there.  

I also wonder how much this feature is really needed?  Since the encounter discard already shuffles automatically if empty when drawing encounter cards, the only real use case for this is when specifically instructed to shuffle the encounter discard into the deck (which on most scenarios will happen once, maybe twice?).  Seems a bit niche to merit a spot on the mat

src/core/MythosArea.ttslua
dscarpac: Yeah, @Chr1Z93 has already mentioned this. I'm going to work on this a bit more and clean up the other encounter draw functions in the same script so they all match.

Re: nicheness -- the other use case is when the encounter deck runs out on its own. It only reshuffles when you draw an encounter card, but the encounter deck has to reshuffle after the last card is drawn and resolved. So, e.g. if the last card drawn is an enemy, if you defeat it, it goes into the discard after the encounter deck has already reshuffled. If there's another spot for it, I'm all ears, but IMO it's an important enough function to not just be a hotkey.

Entrox-Licher: Oh wow that's super interesting, I didn't know that's how the rules worked!  Definitely been playing that one wrong XD. 
 Too much muscle memory from other card games (even this one!  Horror/reshuffle of player cards only triggers on the next instance of draw, not when empty.  What an odd deviation).  In this case, I think it would be better to just change the existing script to check if the encounter deck is empty _after_ drawing, rather than before, and trigger the reshuffle, in accordance to the rules, as normal there.  

Then that leaves this button with just the first niche use case, which I don't think justifies its existence (as a button, hotkey or anything else).  We have very specifically strayed away from automating scenario-specific effects (aside from the odd set-up helper), so I don't really see a reason to start now, especially for such a basic action.

dscarpac: I may have communicated this poorly -- the encounter card draw script definitely should stay as it is -- we can't predict when a card has finished resolving, so the encounter deck reshuffle has to be triggered manually.

Entrox-Licher: Aaaaaah I see what you're saying now.  Hmmm I'm starting to come around to the use case then.  I still think we want to add 2 broadcast messages then: one when the deck is empty, to remind people to click the button to reshuffle, and one if the old reshuffle script is triggered, to 'chastise' them for it a bit and make sure they're playing it correctly.  Or maybe even disable the default reshuffle altogether honestly, to avoid that play pattern trap.

That being said, if it is going to be much more frequently called, I would consider also making a hotkey for it.  I'd let @Chr1Z93 weigh in on that decision though.

dscarpac: I'm also rewriting some of the rest of the MythosArea script under @Chr1Z93's tutelage -- some of this script makes parts of the other scripts redundant, whether or not we have a button to call it. You'll see some more soon :-D

Glitchy-Tozier:
I do think this comes up at least once or twice each scenario (mostly when the agenda flips).

_However_, i also agree that this functionality might not deserve a spot. Currently, space is reserved for functionality that does one of the following:
+ comes up very often and saves you a small amout of work (discard encounter card)
+ comes up seldomly but saves you a lot of work (remove all doom/clues from play)
+ (comes up often and saves you a lot of work (upkeep button))

This new button seems to be in the fourth category: It comes up rarely AND saves you only a minimal amout of work.

Chr1Z93:
1) Imo a button to reshuffle the discard pile would be useful enough to merit its existence. Agendas (or even acts) ask you to do this at least every other game (probably way more on average).

2) As @dscarpac correctly pointed out, the reshuffling when drawing a new card is too late. So I do like this addition. There should probably be a reminder somewhere to trigger this (maybe directly on the MythosArea, below the encounter deck?).

src/core/MythosArea.ttslua
```diff
@@ -160,7 +154,14 @@ function drawEncounterCard(params)
     actualEncounterCardDraw(card, params)
   else
     -- nothing here, time to reshuffle
-    reshuffleEncounterDeck(params)
+    if reshuffledAlready == true then
+      reshuffledAlready = false
+      -- print something?
```
Chr1Z93, line 159: Should replace this with an actual message 🤣

src/core/MythosArea.ttslua
```diff
@@ -180,27 +181,31 @@ function actualEncounterCardDraw(card, params)
   deckLib.placeOrMergeIntoDeck(card, pos, { 0, rotY, faceUpRotation })
 end
 
-function reshuffleEncounterDeck(params)
-  -- flag to avoid multiple calls
+function reshuffleEncounterDeck()
+  --flag to avoid multiple calls
```
Chr1Z93, line 185: Nitpick: Don't remove the space

src/core/MythosArea.ttslua
```diff
@@ -180,27 +181,31 @@ function actualEncounterCardDraw(card, params)
...
-    if not deck.is_face_down then deck.flip() end
-    deck.shuffle()
-    deck.setPositionSmooth(Vector(ENCOUNTER_DECK_POS) + Vector(0, 2, 0), false, true)
-    Wait.time(function() actualEncounterCardDraw(deck.takeObject({ index = 0 }), params) end, 0.5)
+  local encounterDeck = getEncounterDeck()
+  local discardPile = searchLib.atPosition(ENCOUNTER_DISCARD_POSITION, "isCardOrDeck")
+    
+  if #discardPile > 0 then
+    local discardDeck = discardPile[1]
+    if not discardDeck.is_face_down then discardDeck.setRotation({0, -90, 180}) end --flips discard pile
```
Chr1Z93, line 193: Nitpick: Split this into 3 lines

src/core/MythosArea.ttslua
```diff
@@ -180,27 +181,31 @@ function actualEncounterCardDraw(card, params)
...
+    if encounterDeck == nil then
+      discardDeck.setPosition(Vector(ENCOUNTER_DECK_POS) + Vector({0, 1, 0}))
+      discardDeck.shuffle()
+    else
+      encounterDeck.putObject(discardDeck)
+      encounterDeck.shuffle()
+    end
+      broadcastToAll("Shuffled encounter discard into deck.", "White")
```
Chr1Z93, line 201: Nitpick: Fix indentation

xml/MythosArea.xml
```diff
@@ -0,0 +1,10 @@
+<Panel 
+  position="124 40 -13"
+  rotation="0 0 180"
+  height="74" width="74">
+
+
+    <Button scale="0.1 0.1 1"  fontsize="62" onClick="reshuffleEncounterDeck">➡</Button>
+
+
+</Panel>
```
Chr1Z93, line 10: Formatting is all over the place. I can recommend this VS Code extension for formatting: https://marketplace.visualstudio.com/items?itemName=DotJoshJohnson.xml

Chr1Z93 (changes requested):

dscarpac:
Cleaned up the button, added mouseover and buttonpress colors:
image
image

Chr1Z93 (approved):
looks good to me!

PR #551, opened 2024-01-18
Added simple detection for disabled physics

Chr1Z93:
This will notify users when their physics aren't set to full.

closes https://github.com/argonui/SCED/issues/549

### DIFF
--- a/src/core/PhysicsDetector.ttslua
+++ b/src/core/PhysicsDetector.ttslua
@@ -2,6 +2,6 @@
 
 -- this event should only fire if physics aren't fully enabled
 function onCollisionExit()
-  broadcastToAll("Physics disabled?", "Orange")
+  broadcastToAll("Physics disabled? Check chat log", "Orange")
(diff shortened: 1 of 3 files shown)


src/core/PhysicsDetector.ttslua
```diff
@@ -0,0 +1,7 @@
+-- will notify the user to enable physics if it appears to not be fully enabled
+
+-- this event should only fire if physics aren't fully enabled
+function onCollisionExit()
+  broadcastToAll("Physics disabled?", "Orange")
```
Entrox-Licher, line 5: I'd add a little bit to the broadcast like "(Check chat log)" or something just to make sure people see the longer message

Entrox-Licher (changes requested):

src/core/PhysicsDetector.ttslua
Chr1Z93: Fair point, will add that

PR #562, opened 2024-01-23
discard hotkey from hand

dscarpac:
uses discard button from closest playermat, not triggering player when card is in a player's hand

### DIFF
--- a/src/core/GameKeyHandler.ttslua
+++ b/src/core/GameKeyHandler.ttslua
@@ -120,6 +120,8 @@ function getColorToDiscardFor(hoveredObject, playerColor)
   local discardForMatColor
   if inArea(pos, areaNearPlaymat) then
     return closestMatColor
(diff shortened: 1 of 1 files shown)


src/core/GameKeyHandler.ttslua
```diff
@@ -120,6 +120,8 @@ function getColorToDiscardFor(hoveredObject, playerColor)
   local discardForMatColor
   if inArea(pos, areaNearPlaymat) then
     return closestMatColor
+  elseif pos.y > 3 then -- discard to closest mat if card is in a hand
```
Chr1Z93, line 123: If you don't mind, could you use the actual value of a handzone instead of hardcoding this?
```lua
Player[playerColor].getHandTransform().position.y
```
This should do the trick (You can use any `handColor` since they'll all be the same, so I'm just using the triggering player's).

Chr1Z93 (changes requested):

Chr1Z93 (approved):

PR #588, opened 2024-02-13
Updated clue counter for Kate

Chr1Z93:
Now counts all clues on the playermat


Entrox-Licher (approved):

Buhallin (commented):
Seems like this will count clues on Cover Up, or on enemies in the threat area?

Chr1Z93:
> Seems like this will count clues on Cover Up, or on enemies in the threat area?

You are totally right! Threat area should get excluded. Thanks for catching that.

PR #592, opened 2024-02-15
Added 3.6.0 Update Notes

BootleggerFinn:
Also included a pair of cards with the new High Quality card backs next to the update notes.

Did not have the space for a special thanks, but I think that's okay this time.


Chr1Z93 (approved):
Great idea with the blank cards!

PR #596, opened 2024-02-16
Import Instructions: Bugfixes and QoL

Chr1Z93:
- trying to remove a non-existent card does no longer stop the instructions after that
- removing a card will now only remove the related mini-card if the quantity got to 0
- additional copies of the instruction header will now ignored instead of ending the execution
- instruction generator will now put each instruction on a new line
- instruction generator will now broadcast an error if it couldn't get an ID for a card

TODO:
- increase inputfield size (gonna wait with this until we get a new image for it)


Entrox-Licher:
While you're at it, the Hemlock Vale index card is upside down

BootleggerFinn (approved):

PR #630, opened 2024-03-04
Performance improvements

Chr1Z93:
Loading the game locks the UI for about 4 seconds due to the AllCardsBag indexing 1.700 cards. This aims to improve the load time (currently down to about 2 seconds).


src/playercards/AllCardsBag.ttslua
Chr1Z93, line 1: The changes to this file are the only non-cosmetic changes in this PR. Gist of it is utilization of the more efficient `json-parse()` method. It is not as robust as `JSON.encode()`, hence the needed changes to some unicode characters in the card metadata of the upgradesheets.

BootleggerFinn (approved):

src/playercards/AllCardsBag.ttslua
BootleggerFinn: Makes sense! Thanks for the context.

PR #639, opened 2024-03-27
Update to AllCardsBag & Deck Importer

Entrox-Licher:


src/arkhamdb/ArkhamDb.ttslua
```diff
@@ -141,29 +141,49 @@ do
...
+    ArkhamDb.adjustDeck(deckData, callback)
+  end
 
-    internal.maybeModifyDeckFromDescription(slots, deck.description_md, playerColor)
+  ArkhamDb.adjustDeck = function(deckData, callback)
+    -- Initialize deck slot table and perform common transformations.  The order of these should not
+    -- be changed, as later steps may act on cards added in each.  For example, a random weakness or
+    -- investigator may have bonded cards or taboo entries, and should be present
+    local slots = deckData.slots
+    internal.maybeDrawRandomWeakness(slots, playerColor)
```
Chr1Z93, line 175: Looks like `playerColor` is nil?

src/arkhamdb/ArkhamDb.ttslua
```diff
@@ -141,29 +141,49 @@ do
...
-    internal.checkTaboos(deck.taboo_id, slots, playerColor)
+    internal.checkTaboos(deckData.taboo_id, slots, playerColor)
     internal.maybeAddUpgradeSheets(slots)
 
     -- get upgrades for customizable cards
-    local customizations = {}
-    if deck.meta then
-      customizations = JSON.decode(deck.meta)
-    end
+
```
Chr1Z93, line 184: Comment seems out of place

src/arkhamdb/DeckImporterMain.ttslua
```diff
@@ -391,6 +391,10 @@ end
 function handleCustomizableUpgrades(cardList, customizations)
   for _, card in ipairs(cardList) do
     if card.metadata.type == "UpgradeSheet" then
+      if customizations["isFormatted"] then
+        card.data["LuaScriptState"] = customizations[card.metadata.id]
+      end
```
Chr1Z93, line 396: Needs comments to explain what it does

src/playercards/AllCardsBag.ttslua
```diff
@@ -1,8 +1,165 @@
...
+agilityIconsIndex = { }
+wildIconsIndex = { }
+typeIndex = { }
+slotIndex = { }
+costIndex = { }
+permanentIndex = { }
+usesIndex = { }
+specialCriteriaIndex = { }
+
+local specialCriteriaManualOverride = {
```
Chr1Z93, line 21: Now that we have 3 special criterias already (and might get more down the line), what do you think about including them in the metadata as a single `|` separated field (like `class`)?
Example:
`"specialCriteria": "healsHorror|parley"`

src/playercards/AllCardsBag.ttslua
```diff
@@ -27,20 +184,49 @@ end
...
+  typeIndex = { }
+  slotIndex = { }
+  slotIndex["arcane"] = { }
+  slotIndex["arcane x2"] = { }
+  slotIndex["ally"] = { }
+  slotIndex["accessory"] = { }
+  slotIndex["body"] = { }
+  slotIndex["tarot"] = { }
+  slotIndex["hand"] = { }
+  slotIndex["hand x2"] = { }
```
Chr1Z93, line 222: For easier maintainability in the future, how about not specifying the subtables of the slotIndex and instead initializing them when the first card with that slot is encountered?
I doubt that we'll get a new slot in the near future, but maybe we get more Masks or Footwears and then it would be neat to be able to search by that if we update the metadata accordingly?

src/playercards/AllCardsBag.ttslua
```diff
@@ -27,20 +184,49 @@ end
...
+  slotIndex[1] = { }
+  slotIndex[2] = { }
+  slotIndex[3] = { }
```
Chr1Z93, line 225: It's bad practice to have number and string keys for the same table. This should be split into a separate table.

src/playercards/AllCardsBag.ttslua
```diff
@@ -147,89 +333,216 @@ function buildSupplementalIndexes()
...
         end
-        if (isMystic) then
-          table.insert(classAndLevelIndex["Mystic"..upgradeKey], cardMetadata.id)
+        cycleIndex[cycleName] = cycleIndex[cycleName] or {}
+        table.insert(cycleIndex[cycleName], cardMetadata.id)
+      end
+
+      local bonded = cardMetadata.bonded
+
+      local checkedTraits = {}
```
Chr1Z93, line 368: What is `checkedTraits` used for?

src/playercards/AllCardsBag.ttslua
```diff
@@ -147,89 +333,216 @@ function buildSupplementalIndexes()
...
+          if traitsIndex[trait] == nil then
+            traitsIndex[trait] = { }
+          end
+          table.insert(traitsIndex[trait], cardMetadata.id)
         end
-        if (isRogue) then
-          table.insert(classAndLevelIndex["Rogue"..upgradeKey], cardMetadata.id)
+      end
+
+      if cardMetadata.willpowerIcons then
```
Chr1Z93, line 380: Instead of repeating the same code 4 times, you could use a loop for this:
```lua
for _, symbol in ipairs({"combat", "intellect", "willpower", "agility", "wild") do
  local index = _G[symbol  .. "IconsIndex"]
  local iconCount = cardMetadata[symbol .. "Icons"] or 0
  index[iconCount] = index[iconCount] or { }
  table.insert(index[iconCount ], cardMetadata.id)
end
```

src/playercards/AllCardsBag.ttslua
```diff
@@ -147,89 +333,216 @@ function buildSupplementalIndexes()
...
+      --         usesIndex[bondCardMetadata.uses[1].count] = { }
+      --       end
+      --       table.insert(usesIndex[bondCardMetadata.uses[1].count], cardMetadata.id)
+      --       if not usesIndex[string.lower(bondCardMetadata.uses[1].type)] then
+      --         usesIndex[string.lower(bondCardMetadata.uses[1].type)] = { }
+      --       end
+      --       table.insert(usesIndex[string.lower(bondCardMetadata.uses[1].type)], cardMetadata.id)
+      --     end
+      --   end
+      -- end
```
Chr1Z93, line 481: Should be removed

src/playercards/AllCardsBag.ttslua
```diff
@@ -147,89 +333,216 @@ function buildSupplementalIndexes()
...
+  if card1Class ~= card2Class then
+    return card1Class < card2Class
   end
   if (card1.data.Nickname ~= card2.data.Nickname) then
     return card1.data.Nickname < card2.data.Nickname
   end
   return card1.data.Description < card2.data.Description
 end
 
+function sortCardsTable (cardsTable)
```
Chr1Z93, line 542: Superfluous space after function name

src/playercards/AllCardsBag.ttslua
```diff
@@ -147,89 +333,216 @@ function buildSupplementalIndexes()
...
+  return table.sort(cardsTable, cardComparator)
```
Chr1Z93, line 543: I don't think the `return` is needed

src/playercards/AllCardsBag.ttslua
```diff
@@ -147,89 +333,216 @@ function buildSupplementalIndexes()
...
+  }
+  card1Class = cardClassOrder[card1Class]
+  card2Class = cardClassOrder[card2Class]
 
-  if (card1.metadata.level ~= card2.metadata.level) then
-    return card1.metadata.level < card2.metadata.level
+  if (card1Level ~= card2Level) then
+    return card1Level < card2Level
+  end
+  if card1Class ~= card2Class then
```
Chr1Z93, line 533: Could use `elseif` to make this shorter

src/playercards/AllCardsBag.ttslua
```diff
@@ -249,34 +562,80 @@ function getCardById(params)
...
     broadcastToAll("Still loading player cards, please try again in a few seconds", {0.9, 0.2, 0.2})
     return { }
   end
-  local upgradeKey
-  if (params.upgraded) then
-    upgradeKey = "-upgrade"
-  else
-    upgradeKey = "-level0"
+  -- Clean up string input
+  local value = nil
```
Chr1Z93, line 577: I think `tableKey` would be a better name than `value` considering it can be a string or number

src/playercards/AllCardsBagApi.ttslua
```diff
@@ -51,24 +51,34 @@ do
...
-  AllCardsBagApi.getCardsByClassAndLevel = function(class, upgraded)
-    return getAllCardsBag().call("getCardsByClassAndLevel", {class = class, upgraded = upgraded})
-  end
-
-  AllCardsBagApi.getCardsByCycle = function(cycle)
-    return getAllCardsBag().call("getCardsByCycle", cycle)
+  -- Used to index into the supplemental card indices 
+  ---@param indexName exact string of the index variable to be referenced, minus the 'Index' part
+  ---@param includeOthers boolean of whether or not to include non-player cards
+  ---@param value key used to index into the dictionary
```
Chr1Z93, line 57: Nit: keep comments in the correct order

Chr1Z93 (changes requested):

src/playercards/AllCardsBag.ttslua
Entrox-Licher: I've got no problem with this inherently (we already do that for things like cost), but it would mean adding yet another extra check to the already lengthy loop.  I doubt it would have a noticeable impact, but still seems worth mentioning.

If we still think that's fine, happy to make that change

src/playercards/AllCardsBag.ttslua
Entrox-Licher: Ooo very smart to inline the ipairs call!  Making this change now!

src/playercards/AllCardsBag.ttslua
Entrox-Licher: Whoops yeah, good catch (Still bitter though XD)

src/playercards/AllCardsBag.ttslua
Entrox-Licher: I agree it's a bit messy, but since this isn't really something that's going to be touched by any other function we can hopefully leave it as is?  If this becomes a separate table, the Filter Card Search would need an entirely separate operator to track them (because I use `_G` for it in the function below).

Alternatively, if it is too intolerable, I could maybe switch to just strings of the numbers instead?

src/playercards/AllCardsBag.ttslua
Entrox-Licher: Yeah...this is probably a good idea.  I will probably make a separate PR for that though, because it will be a _lot_ of metadata changes

src/playercards/AllCardsBag.ttslua
Chr1Z93: Good point. I agree that we should prioritize performance in this case even if the difference won't be big.

src/arkhamdb/ArkhamDb.ttslua
Entrox-Licher: Good catch!  Should be `deckData.playerColor`

src/arkhamdb/DeckImporterMain.ttslua
Entrox-Licher: This is specific to the deckbuilder, I'll just scrap it for now

src/playercards/AllCardsBag.ttslua
Entrox-Licher: It lets you inline the sorting of a table, which I think can be useful (used in some places in the filter search & deck builder, but removable if you disagree)

src/playercards/AllCardsBag.ttslua
Entrox-Licher: 100% agree, changed!

src/playercards/AllCardsBag.ttslua
Entrox-Licher: Making the change, but how does it make it shorter?  Isn't it the same number of lines & same execution time, just semantically different?

src/playercards/AllCardsBag.ttslua
Entrox-Licher: Uuuuuuh no clue haha, must be a remnant of an older sorting method.  Removed!

src/playercards/AllCardsBag.ttslua
Chr1Z93: Lua already passed tables by value, so you don't need to return the reference is what I meant :)

src/playercards/AllCardsBag.ttslua
Chr1Z93: You save an `end` statement for each `elseif`:

```lua
if card1Level ~= card2Level then
    return card1Level < card2Level
elseif card1Class ~= card2Class then
    return card1Class < card2Class
elseif card1.data.Nickname ~= card2.data.Nickname then
    return card1.data.Nickname < card2.data.Nickname
else
   return card1.data.Description < card2.data.Description
end
```

Chr1Z93 (changes requested):

src/playercards/AllCardsBag.ttslua
Entrox-Licher: For clarity, currently the code for the index looks like this:
```lua
local filterOptions = {
  y = "cycle",
  f = "class",
  k = "traits",
  o = "cost",
  w = "willpowerIcons",
  c = "combatIcons",
  i = "intellectIcons",
  a = "agilityIcons",
  d = "wildIcons",
  t = "type",
  p = "level",
  z = "slot",
  v = "permanent",
  u = "uses",
  q = "specialCriteria"
}
...
matchingCards = allCardsBagApi.getCardsByData(filterOptions[filter], value, includeOthers)
```


So the issue here really is just the namespace of the variables for the tables in `AllCardsBag` gets very ugly if we make this change, because they have to be consistent in order to be usable with `_G`.  So that means all the tables that use numeric indices would have to be appended with `_number` or something, even the ones that have exclusively numeric indices like the icon ones.  

Personally I don't see some mixed indices in the tables as a particularly big problem, since in practical terms the only limitation _I think_ it presents is not being able to use `ipairs`?  I guess I'm just not seeing the downside?

Chr1Z93: Isn't this just mixing two indices in one?
`slotIndex` should be holding all cards that take up a specific slot in each subtable.
The "1, 2 and 3" subtables are technically an additional index for the slot count.
I don't see why we have to combine that into a single index.

Chr1Z93:
Closed for now, but on my radar for the near future: https://github.com/argonui/SCED/issues/762

PR #640, opened 2024-03-27
Updated Extra Action Functionality & Added More Investigators

Entrox-Licher:
Made the extra action token update more dynamic; the extra action (if any) is now defined in the investigator metadata, rather than hard coded into the Playmat code.  

Changed default behaviour (when no extra action token corresponds) to be leaving the token as neutral, instead of changing it to class color, to better indicate when a character is using the extra token.

Additionally, added more investigators who get the extra action token, to better track once per round abilities (e.g. Yorick and Roland)


Entrox-Licher:
New investigators with token associated with them, token type in parenthesis (feel free to suggest more or changing/removing one of these):

- Roland Normal, PB & Alt Art (Investigate)
- Skids Normal, PB & Alt Art (Rogue)
- Agnes Normal PB & Alt Art (Spell)
- Zoey || & PF (Spell)
- Rex Taboo (Investigate)
- Jim || & PF (Spell)
- Ashcan all versions (Item)
- Mark (Guardian Fight)
- Leo (Parley)
- Diana (Spell)
- Rita (Evade)
- Marie (Spell) (We were missing this one XD)
- Mandy Normal & Taboo (Tome)
- Trish Normal & Taboo (Investigate)
- Dexter Normal & Alt Art (Item)
- Silas Normal & Alt Art (Tome)  (No great matches here?)
- Norman Normal & Alt Art (Tome)
- Monterey || & PF (Item)
- Carson (Parley)
- Amina (Item)
- Wilson (Item)
- Alessandra (Parley) (Also missing this one haha)


Specifically, I'm unsure whether the 'once per phase' effects (e.g. Agnes & || Zoey) deserve a token or not, since that could be confusing (since it requires manual flips).

Chr1Z93:
For the sake of visibility, I'll post my concerns here too:
- Giving investigators without additional actions specific tokens could lead to confusion (e.g. Amina getting the "Play item" action)
- Mixing classes doesn't look nice (Amina's "Play item" action will be a survivor token since only Bob has it atm)

BootleggerFinn:
I agree that if investigators are getting bespoke extra action tokens, then those without extra actions probably shouldn't have extra action tokens at all. On the other hand, it wouldn't do to have no easy way of adding an extra action token if a played asset grants one. Maybe those without a match just get a neutral just-in-case token?

BootleggerFinn:
What if we had a 21st token state that showed a reaction icon, with no class color bias? While it's not _technically_ an extra action, it's meaning would be very clear, I think.

Chr1Z93:
@Entrox-Licher Would you mind reverting the additional action type changes from this PR? I'd still like to have the extra token type in the metadata, but as we already discussed stay away from the additional changes until we have proper tokens (or a dynamic one!) for that.

On that note, what do you think about adding `none` as an option? This would remove the additional token and would be useful for investigators like Calvin, Carolyn, Daniela, Darrell etc.

Also I think it might be more clear to call this field `extraToken` since we are not just using it for actions.

Chr1Z93 (changes requested):
See conversation above

Chr1Z93:
Closing this PR to revisit it here: https://github.com/argonui/SCED/pull/703

PR #645, opened 2024-04-11
Added deck-checking for "Forced Learning"

Chr1Z93:
Playmat will now check decks for forced learning (e.g. the set-aside pile of permanents)

### DIFF
--- a/objects/AllPlayerCards.15bb07/ForcedLearning.fa06f9.gmnotes
+++ b/objects/AllPlayerCards.15bb07/ForcedLearning.fa06f9.gmnotes
@@ -2,6 +2,7 @@
   "id": "08031",
   "type": "Asset",
   "class": "Seeker",
+  "startsInPlay": true,
   "level": 0,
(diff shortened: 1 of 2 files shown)


Entrox-Licher:
This was an intentional omission on my part from the original PR.  With slots already at a premium, I don't love implying the need for this to take one up to track something the mod already does automatically.  I think the more elegant solution would be to 'fan out' rather than stack the permanents on the side of the player mat.

Chr1Z93:
> This was an intentional omission on my part from the original PR. With slots already at a premium, I don't love implying the need for this to take one up to track something the mod already does automatically. I think the more elegant solution would be to 'fan out' rather than stack the permanents on the side of the player mat.

I see your point. If we don't let it start in play then the upkeep script should just check decks too. I don't like spreading the permanents because there could be a lot of them.

Chr1Z93:
@Entrox-Licher I updated the script to now check decks too.

Entrox-Licher (approved):

PR #657, opened 2024-04-18
Nkosi Mabati helper / redrawing token tools

dscarpac:
image

Updated helper as originally described in the Discord post. Notable improvements:
(1) Allows for any symbol to be chosen as a sigil, including custom tokens
(2) Replacement token is redrawn into the same position on the playermat as the old token
(3) Some more initial support for token redrawing to seed ideas for Wendy, Custom Modifications, etc. into the Global script, including removing the replaced token from the token stats list.


src/chaosbag/ChaosBagApi.ttslua
```diff
@@ -71,8 +71,9 @@ do
   ---@param drawAdditional boolean Controls whether additional tokens should be drawn
   ---@param tokenType? string Name of token (e.g. "Bless") to be drawn from the bag
   ---@param guidToBeResolved? string GUID of the sealed token to be resolved instead of drawing a token from the bag
-  ChaosBagApi.drawChaosToken = function(mat, drawAdditional, tokenType, guidToBeResolved)
-    return Global.call("drawChaosToken", {mat = mat, drawAdditional = drawAdditional, tokenType = tokenType, guidToBeResolved = guidToBeResolved})
+  ---@param returnedToken tts__Object Token to be replaced with newly drawn token
```
Chr1Z93, line 74: Add a `?` at the end of the parameter name to indicate that it is optional

src/core/Global.ttslua
```diff
@@ -295,28 +303,42 @@ function drawChaosToken(params)
     chaosBag.shuffle()
 
     -- add the token to the list, compute new position based on list length
-    tokenOffset[1] = tokenOffset[1] + (0.17 * #chaosTokens)
+    if params.returnedToken then
+      trackChaosToken(params.returnedToken.getName(), matGUID, "subtract")
+      indexOfReturnedToken = getTokenIndex(params.returnedToken)
+      tokenPosition = params.returnedToken.getPosition()
```
Chr1Z93, line 309: Should make these variables local (declare above so that they still work as intended)

src/core/Global.ttslua
```diff
@@ -353,13 +380,19 @@ end
 -- chaos token stat tracker
 ---------------------------------------------------------
 
-function trackChaosToken(tokenName, matGUID)
+function trackChaosToken(tokenName, matGUID, addOrSubtract)
```
Chr1Z93, line 383: Imo a boolean variable would be better for `addOrSubtract`. Could call it `subtract` and then default to adding if it is false or nil.

src/playercards/cards/NkosiMabati3.ttslua
Chr1Z93, line 1: Indentation doesn't match the default of "2 spaces"

src/playercards/cards/NkosiMabati3.ttslua
```diff
@@ -0,0 +1,185 @@
...
+)
+end 
+
+-- Create dialog window to choose sigil and create sigil-drawing button
+function chooseSigil(playerColor)
+    self.clearContextMenu()
+    self.addContextMenuItem("Clear Helper", deleteButtons)
+    Player[playerColor].showOptionsDialog("Choose Sigil", tokenNames, 1, 
+        function(chosenToken)
+            if chosenToken == "Custom Token" then
```
Chr1Z93, line 78: I think it would be nice if you had two small helper functions to convert from "token names in bag" to "human readable token names" (and vice versa). These could be used instead of these "if"-statements that happen in multiple places.

Chr1Z93 (changes requested):

src/core/Global.ttslua
```diff
@@ -1188,13 +1208,14 @@ function contentDownloadCallback(request, params)
...
-    spawnTable.rotation = { 0, 270, 0}
+    spawnTable.rotation = { 0, 270, 0 }
 
     local pos = getValidSpawnPosition()
     if pos then
       spawnTable.position = pos
     else
-      broadcastToAll("Please make space in the area below the tentacle stand in the upper middle of the table and try again.", "Red")
+      broadcastToAll(
+      "Please make space in the area below the tentacle stand in the upper middle of the table and try again.", "Red")
```
Chr1Z93, line 1218: Would prefer to keep this in a single line

src/core/Global.ttslua
```diff
@@ -1352,7 +1373,9 @@ function playermatRemovalSelected(player, selectedIndex, id)
   if mat then
     -- confirmation dialog about deletion
     player.pingTable(mat.getPosition())
-    player.showConfirmDialog("Do you really want to remove " .. matColor .. "'s playermat and related objects? This can't be reversed.", function() removePlayermat(matColor) end)
+    player.showConfirmDialog(
+    "Do you really want to remove " .. matColor .. "'s playermat and related objects? This can't be reversed.",
+      function() removePlayermat(matColor) end)
```
Chr1Z93, line 1378: Would also keep this in a single line

Chr1Z93 (changes requested):

Chr1Z93 (approved):

PR #662, opened 2024-04-19
Added uses to EotE weaknesses

Chr1Z93:


BootleggerFinn (approved):
Interesting idea. I think I'll steal this for my IRL games!

PR #672, opened 2024-05-01
Added numpad hotkey to spawn action token

dscarpac:
Numpad (1) now spawns an action token. It detects the nearest investigator card and changes color to that class.

### DIFF
--- a/src/util/TokenSpawnTool.ttslua
+++ b/src/util/TokenSpawnTool.ttslua
@@ -90,6 +90,7 @@ function onScriptingButtonDown(index, playerColor)
   elseif tokenType == "actionToken" then
     local matColor = playmatApi.getMatColorByPosition(position)
     local result = playmatApi.searchAroundPlaymat(matColor, "isCard")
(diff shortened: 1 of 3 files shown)


src/util/TokenSpawnTool.ttslua
```diff
@@ -74,6 +86,21 @@ function onScriptingButtonDown(index, playerColor)
         end
       end
     end
+  -- check for nearest investigator card and change action token state to its class 
+  elseif tokenType == "actionToken" then
+    local matColor = playmatApi.getMatColorByPosition(position)
+    local result = playmatApi.searchAroundPlaymat(matColor, "isCard")
+    for j, card in ipairs(result) do
```
Chr1Z93, line 93: Should add `local investigatorClass = "Neutral"` before the loop to
a) not define it globally
b) cover the case of not finding an investigator card (e.g. because metadata is missing)

Chr1Z93 (changes requested):

Chr1Z93 (approved):

PR #681, opened 2024-05-10
New hotkey to take cards into threat area

dscarpac:
Can be used to (a) pick up exhausted or unengaged enemies in the playarea, maintaining the ready/exhausted state
(b) engage enemies off of other players threat areas into your own
(c) move a newly drawn encounter card into an empty slot on the right.


src/core/GameKeyHandler.ttslua
```diff
@@ -57,6 +58,80 @@ function addDoomToAgenda()
   doomCounter.call("addVal", 1)
 end
 
+-- move the hovered object to the nearest empty slot on the playermat
+function takeCardIntoThreatArea(playerColor, hoveredObject)
+  -- only continue if an unlocked card
```
Chr1Z93, line 63: I don't think you can even hover a locked object but I might be wrong (at least you don't get the yellow outline)

src/core/GameKeyHandler.ttslua
```diff
@@ -57,6 +58,80 @@ function addDoomToAgenda()
...
+      or hoveredObject.hasTag("Location")
+      or hoveredObject.locked then
+    broadcastToColor("Hover a non-location and try again.", playerColor, "Yellow")
+    return
+  end
+
+  local matColor = playmatApi.getMatColor(playerColor)
+  local mat = guidReferenceApi.getObjectByOwnerAndType(matColor, "Playermat")
+  -- do not continue if the threat area is already full
+  if playmatApi.getEncounterCardDrawPosition(matColor, false) == playmatApi.getEncounterCardDrawPosition(matColor, true) then
```
Chr1Z93, line 75: I think you actually need to compare the coordinates since Lua does only check if they are the same table reference if you do it like this (and that will always be false).

src/core/GameKeyHandler.ttslua
```diff
@@ -57,6 +58,80 @@ function addDoomToAgenda()
...
+  
+  -- initialize list of objects to move
+  local moveTheseObjects = {}
+  
+  for _, obj in ipairs(searchLib.onObject(hoveredObject, "isTileOrToken")) do
+    table.insert(moveTheseObjects, obj)
+  end
+  
+  -- find out if the original card is on the green or red playmats
+  local result
```
Chr1Z93, line 89: Superfluous line

src/core/GameKeyHandler.ttslua
```diff
@@ -57,6 +58,80 @@ function addDoomToAgenda()
...
+  end
+  -- determine modifiers for the playmats
+  local modifierY
+  if originalMatColor == "Red" then
+    modifierY = 90
+  elseif originalMatColor == "Green" then
+    modifierY = -90
+  else
+    modifierY = 0
+  end
```
Chr1Z93, line 114: Okay this needs to be more dynamic 😅

Chr1Z93 (changes requested):

src/core/GameKeyHandler.ttslua
dscarpac: I might be doing something wrong here, or it was because of something else, but in testing without this line, the scripted saw a card as being on the green playmat even when it was on the center playmat, as long as that card was previously on the green playmat.

src/core/GameKeyHandler.ttslua
dscarpac: Suggestions?

src/core/GameKeyHandler.ttslua
Chr1Z93: I'm talking about the double `local result`

src/core/GameKeyHandler.ttslua
Chr1Z93: Nvm this works

src/core/GameKeyHandler.ttslua
Chr1Z93: Made something for you: https://github.com/argonui/SCED/pull/682

Chr1Z93 (approved):

PR #699, opened 2024-05-25
Redraw the playmat slot XML when a player joins

Chr1Z93:
This should fix the issue of new clients seeing letters instead of symbols


dscarpac:
Do we also need this for other uses of custom fonts, like the usage of Teutonic on the Reshuffle button on the MythosArea or on XML buttons like the Tekeli-li helper, etc.?

dscarpac (approved):

PR #716, opened 2024-06-14
Adds helpers for automatic token redrawing and replacement

dscarpac:
Extends Nkosi Mabati functionality to addition cards that return and redraw tokens. Future cards can be easily added based on if they only redraw specific tokens.

- Wendy Adams
- False Covenant (additionally returns Curse to pool)
- Heavy Furs / Claypool's Furs
- Custom Modifications

Additionally Nkosi script has been updated to share scripts with the above cards.

### DIFF
--- a/src/chaosbag/BlessCurseManagerApi.ttslua
+++ b/src/chaosbag/BlessCurseManagerApi.ttslua
@@ -24,8 +24,9 @@ do
   -- updates the internal count (called by cards that seal bless/curse tokens)
   ---@param type string Type of chaos token ("Bless" or "Curse")
   ---@param guid string GUID of the token
-  BlessCurseManagerApi.releasedToken = function(type, guid)
-    getManager().call("releasedToken", { type = type, guid = guid })
+  ---@param fromDrawn? boolean Whether or not token was just drawn from the chaos bag
+  BlessCurseManagerApi.releasedToken = function(type, guid, fromDrawn)
+    getManager().call("releasedToken", { type = type, guid = guid, fromDrawn = fromDrawn })
   end
 
   -- updates the internal count (called by cards that seal bless/curse tokens)
@@ -52,6 +53,18 @@ do
   ---@param hoveredObject tts__Object Hovered object
   BlessCurseManagerApi.addBlurseSealingMenu = function(playerColor, hoveredObject)
     getManager().call("addMenuOptions", { playerColor = playerColor, hoveredObject = hoveredObject })
+  end
+  
+  -- adds bless / curse to the chaos bag
+  -- @param type string Type of chaos token ("Bless" or "Curse")
+  BlessCurseManagerApi.addToken = function(type)
+    getManager().call("addToken", type)
+  end
+
+  -- removes bless / curse from the chaos bag
+  -- @param type string Type of chaos token ("Bless" or "Curse")
+  BlessCurseManagerApi.removeToken = function(type)
+    getManager().call("removeToken", type)
   end
   
   return BlessCurseManagerApi
--- a/src/chaosbag/BlessCurseManagerApi.ttslua
+++ b/src/chaosbag/BlessCurseManagerApi.ttslua
@@ -24,9 +24,9 @@ do
   -- updates the internal count (called by cards that seal bless/curse tokens)
   ---@param type string Type of chaos token ("Bless" or "Curse")
   ---@param guid string GUID of the token
-  ---@param fromDrawn? boolean Whether or not token was just drawn from the chaos bag
-  BlessCurseManagerApi.releasedToken = function(type, guid, fromDrawn)
-    getManager().call("releasedToken", { type = type, guid = guid, fromDrawn = fromDrawn })
+  ---@param fromBag? boolean Whether or not token was just drawn from the chaos bag
+  BlessCurseManagerApi.releasedToken = function(type, guid, fromBag)
+    getManager().call("releasedToken", { type = type, guid = guid, fromBag = fromBag })
   end
 
   -- updates the internal count (called by cards that seal bless/curse tokens)
--- a/src/chaosbag/BlessCurseManagerApi.ttslua
+++ b/src/chaosbag/BlessCurseManagerApi.ttslua
@@ -56,13 +56,13 @@ do
   end
   
   -- adds bless / curse to the chaos bag
-  -- @param type string Type of chaos token ("Bless" or "Curse")
+  ---@param type string Type of chaos token ("Bless" or "Curse")
   BlessCurseManagerApi.addToken = function(type)
     getManager().call("addToken", type)
   end
 
   -- removes bless / curse from the chaos bag
-  -- @param type string Type of chaos token ("Bless" or "Curse")
+  ---@param type string Type of chaos token ("Bless" or "Curse")
   BlessCurseManagerApi.removeToken = function(type)
     getManager().call("removeToken", type)
   end
--- a/src/chaosbag/ChaosBagApi.ttslua
+++ b/src/chaosbag/ChaosBagApi.ttslua
@@ -71,9 +71,9 @@ do
   ---@param drawAdditional boolean Controls whether additional tokens should be drawn
   ---@param tokenType? string Name of token (e.g. "Bless") to be drawn from the bag
   ---@param guidToBeResolved? string GUID of the sealed token to be resolved instead of drawing a token from the bag
-  ---@param returnedToken? tts__Object Token to be replaced with newly drawn token
+  ---@param takeParameters? table Position and rotation of the location where the new token should be drawn to, usually to replace a returned token
   ChaosBagApi.drawChaosToken = function(mat, drawAdditional, tokenType, guidToBeResolved, returnedToken)
-    return Global.call("drawChaosToken", {mat = mat, drawAdditional = drawAdditional, tokenType = tokenType, guidToBeResolved = guidToBeResolved, returnedToken = returnedToken})
+    return Global.call("drawChaosToken", {mat = mat, drawAdditional = drawAdditional, tokenType = tokenType, guidToBeResolved = guidToBeResolved, takeParameters = takeParameters})
   end
 
   -- returns a Table List of chaos token ids in the current chaos bag
--- a/src/chaosbag/ChaosBagApi.ttslua
+++ b/src/chaosbag/ChaosBagApi.ttslua
@@ -72,7 +72,7 @@ do
   ---@param tokenType? string Name of token (e.g. "Bless") to be drawn from the bag
   ---@param guidToBeResolved? string GUID of the sealed token to be resolved instead of drawing a token from the bag
   ---@param takeParameters? table Position and rotation of the location where the new token should be drawn to, usually to replace a returned token
-  ChaosBagApi.drawChaosToken = function(mat, drawAdditional, tokenType, guidToBeResolved, returnedToken)
+  ChaosBagApi.drawChaosToken = function(mat, drawAdditional, tokenType, guidToBeResolved)
     return Global.call("drawChaosToken", {mat = mat, drawAdditional = drawAdditional, tokenType = tokenType, guidToBeResolved = guidToBeResolved, takeParameters = takeParameters})
   end
 
--- a/src/core/Global.ttslua
+++ b/src/core/Global.ttslua
@@ -286,6 +286,7 @@ end
 
 function activeRedrawEffect(originParams)
   dataFromLastReplacementEffect = originParams
+  makeButtonsToRedraw()
 end
 
 function makeButtonsToRedraw()
@@ -294,8 +295,7 @@ function makeButtonsToRedraw()
     return
   end
 
-  local tokensInPlay = getChaosTokensinPlay()
-  if #tokensInPlay == 0 then
+  if #chaosTokens == 0 then
     broadcastToAll("No tokens found in play area", "Red")
     return
   end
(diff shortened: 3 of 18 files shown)


src/core/Global.ttslua
```diff
@@ -329,34 +453,28 @@ function drawChaosToken(params)
...
+    if params.takeParameters then
+      takeParameters.position = params.takeParameters.position
+      takeParameters.rotation = params.takeParameters.rotation
     else
       tokenOffset[1] = tokenOffset[1] + (0.17 * #chaosTokens)
       takeParameters.position = params.mat.positionToWorld(tokenOffset)
       takeParameters.rotation = params.mat.getRotation()
     end
 
+    chaosBag.shuffle()
```
Chr1Z93, line 475: What's the reason for moving the shuffling down here?

src/core/Global.ttslua
```diff
@@ -283,6 +283,130 @@ function getTokenIndex(token)
...
+    takeParameters.rotation = mat.getRotation() + Vector(0, 0, -8)
+  else
+    takeParameters.rotation = returnedToken.getRotation()
+  end
+
+  returnChaosTokenToBag(returnedToken)
+  if string.find("all|eligible|ineligible|falseCovenant", id) then
+     token = drawChaosToken({mat = mat, drawAdditional = true, takeParameters = takeParameters})
+  else
+    token = drawChaosToken({mat = mat, drawAdditional = true, takeParameters = takeParameters, tokenType = id})
```
Chr1Z93, line 397: What's the "id" used for?

Chr1Z93 (changes requested):

src/core/Global.ttslua
```diff
@@ -283,6 +284,134 @@ function getTokenIndex(token)
...
+  dataFromLastReplacementEffect = originParams
+end
+
+function makeButtonsToRedraw()
+  if isTokenXMLActive == true then
+    broadcastToAll("Clear already active buttons first, then try again", "Red")
+    return
+  end
+
+  local tokensInPlay = getChaosTokensinPlay()
```
Chr1Z93, line 297: This should use the existing `chaosTokens` table instead of the function that's only needed for outside access.

src/core/Global.ttslua
```diff
@@ -283,6 +284,134 @@ function getTokenIndex(token)
...
+  if #tokensInPlay == 0 then
+    broadcastToAll("No tokens found in play area", "Red")
+    return
+  end
+
+  local chaosbag = findChaosBag()
+  local matchingTokensInPlay = {}
+  -- determine if only some tokens are able to be returned to the bag
+  if dataFromLastReplacementEffect.redrawEligibilty == "ineligible" then
```
Chr1Z93, line 306: As talked about, I'd rather implement a system that supports both "valid" and "invalid" tokens similar to the "CardsThatSealTokens"

src/core/Global.ttslua
```diff
@@ -283,6 +284,134 @@ function getTokenIndex(token)
...
+    end
+    isTokenXMLActive = true
+  -- no need to make buttons if there is only one eligible token to return and redraw
+  else
+    returnAndRedraw(_, matchingTokensInPlay[1].getGUID())
+  end
+end
+
+function returnAndRedraw(_, tokenGUID)
+  local indexOfReturnedToken
```
Chr1Z93, line 378: Don't need this line, just add the `local` to the declaration below

src/core/Global.ttslua
```diff
@@ -283,6 +284,134 @@ function getTokenIndex(token)
...
+    takeParameters.rotation = mat.getRotation() + Vector(0, 0, -8)
+  else
+    takeParameters.rotation = returnedToken.getRotation()
+  end
+
+  returnChaosTokenToBag(returnedToken)
+  if dataFromLastReplacementEffect.redrawnTokenType == "random" then
+    token = drawChaosToken({mat = mat, drawAdditional = true, takeParameters = takeParameters})
+  else
+    token = drawChaosToken({mat = mat, drawAdditional = true, takeParameters = takeParameters, tokenType = dataFromLastReplacementEffect.redrawnTokenType})
```
Chr1Z93, line 399: If you change `redrawnTokenType` to a boolean field like `drawRandomReplacement = true / false` then you don't need this whole if statement

src/core/Global.ttslua
```diff
@@ -329,34 +458,28 @@ function drawChaosToken(params)
...
   chaosTokensLastMatGUID = matGUID
-
+  
   -- if we have left clicked and have no tokens OR if we have right clicked
   if params.drawAdditional or #chaosTokens == 0 then
     local chaosBag = findChaosBag()
     if #chaosBag.getObjects() == 0 then return end
-    chaosBag.shuffle()
 
     local indexOfReturnedToken
```
Chr1Z93, line 467: Looks like we don't need this anymore?

src/playercards/CardsThatRedrawTokens.ttslua
```diff
@@ -0,0 +1,61 @@
...
+      self.addContextMenuItem("Clear Helper", deleteButton)
+    end
+  end
+end
+
+function makeXMLButton()
+  turnOnHelper = true
+  self.clearContextMenu()
+  self.addContextMenuItem("Clear Helper", deleteButton)
+  log(color)
```
Chr1Z93, line 24: Leftover log statement

src/playercards/CardsThatRedrawTokens.ttslua
```diff
@@ -0,0 +1,61 @@
...
+      xmlTable[1].attributes.iconWidth = "400"
+      xmlTable[1].attributes.iconAlignment = "Right"
+      xmlTable[1].attributes.icon = icon
+  end
+  self.UI.setXmlTable(xmlTable)
+end
+
+function triggerXMLTokenLabelCreation()
+    Global.call("activeRedrawEffect", originParams)
+    Global.call("makeButtonsToRedraw")
```
Chr1Z93, line 52: I assume these functions are always gonna be used together? In that case they should get combined (or at least let the first one call the second one)

src/playercards/CardsThatRedrawTokens.ttslua
```diff
@@ -0,0 +1,61 @@
...
+
+function onLoad(savedData)
+  self.addContextMenuItem("Enable Helper", makeXMLButton)
+
+  if savedData and savedData ~= "" then
+    turnOnHelper = JSON.decode(savedData)
+    if turnOnHelper == true then
+      makeXMLButton()
+      self.clearContextMenu()
+      self.addContextMenuItem("Clear Helper", deleteButton)
```
Chr1Z93, line 15: Both of these lines are in the `makeXMLButton()` function already

Chr1Z93 (changes requested):

src/core/Global.ttslua
```diff
@@ -283,6 +284,127 @@ function getTokenIndex(token)
...
+    broadcastToAll("Clear already active buttons first, then try again", "Red")
+    return
+  end
+
+  if #chaosTokens == 0 then
+    broadcastToAll("No tokens found in play area", "Red")
+    return
+  end
+
+  local chaosbag = findChaosBag()
```
Chr1Z93, line 303: What's this for?

src/core/Global.ttslua
```diff
@@ -283,6 +284,127 @@ function getTokenIndex(token)
...
+  local matchingTokensInPlay = {}
+
+  -- determine if only some tokens are able to be returned to the bag
+  for _, token in ipairs(chaosTokens) do
+    local tokenName = getReadableTokenName(token.getName())
+    -- nil handling
+    redrawData.VALID_TOKENS = redrawData.VALID_TOKENS or {}
+    redrawData.INVALID_TOKENS = redrawData.INVALID_TOKENS or {}
```
Chr1Z93, line 311: Should be outside of the loop, only need it once

src/core/Global.ttslua
```diff
@@ -283,6 +284,127 @@ function getTokenIndex(token)
...
+  
+    -- allow valid tokens or not invalid tokens, also allow any token if both lists empty
+    if (redrawData.VALID_TOKENS[tokenName] and #redrawData.INVALID_TOKENS == 0) ~= nil or (isTableEmpty(redrawData.VALID_TOKENS) and not redrawData.INVALID_TOKENS[tokenName]) or
```
Chr1Z93, line 314: Should use the `isTableEmpty()` function for the first condition too (and I think the nil comparision should just be for the first part):

```lua
if (redrawData.VALID_TOKENS[tokenName] ~= nil and isTableEmpty(redrawData.INVALID_TOKENS)) ...
```

src/core/Global.ttslua
```diff
@@ -283,6 +284,127 @@ function getTokenIndex(token)
...
+
+  if #chaosTokens > indexOfReturnedToken then
+    takeParameters.rotation = mat.getRotation() + Vector(0, 0, -8)
+  else
+    takeParameters.rotation = returnedToken.getRotation()
+  end
+
+  returnChaosTokenToBag(returnedToken)
+  if redrawData.redrawnTokenType == "random" then
+    token = drawChaosToken({mat = mat, drawAdditional = true, takeParameters = takeParameters})
```
Chr1Z93, line 391: Should use `chaosTokens[indexOfReturnedToken] = ...` instead of leaking a global variable (`token`)

src/core/Global.ttslua
```diff
@@ -283,6 +284,127 @@ function getTokenIndex(token)
...
+  else
+    token = drawChaosToken({mat = mat, drawAdditional = true, takeParameters = takeParameters, tokenType = redrawData.redrawnTokenType})
+  end
+  
+  chaosTokens[indexOfReturnedToken] = token
+  if redrawData.triggeringCard == "FalseCovenant" then
```
Chr1Z93, line 397: Should use the name of the returned token instead of relying on the triggering card

src/playercards/CardsThatRedrawTokens.ttslua
```diff
@@ -0,0 +1,59 @@
...
+      xmlTable[1].attributes.iconWidth = "400"
+      xmlTable[1].attributes.iconAlignment = "Right"
+      xmlTable[1].attributes.icon = buttonIcon
+  end
+  self.UI.setXmlTable(xmlTable)
+end
+
+function triggerXMLTokenLabelCreation()
+  -- needs to be its own function in order to pass originParams as a table
+    log(originParams)
```
Chr1Z93, line 49: Leftover log statement

src/playercards/CardsThatRedrawTokens.ttslua
```diff
@@ -0,0 +1,59 @@
...
+  Global.call("activeRedrawEffect", originParams)
+end
+
+-- Delete button
+function deleteButton()
+  self.clearContextMenu()
+  self.addContextMenuItem("Enable Helper", makeXMLButton)
+  self.UI.setXml("")
```
Chr1Z93, line 57: How about showing / hiding the XML instead of recreating it? (e.g. create it once in onload (but hidden) and then only show/hide subsequently)

src/playercards/cards/FalseCovenant.ttslua
```diff
@@ -0,0 +1,14 @@
+buttonValue = "Cancel"
+buttonIcon = "token-curse"
+buttonColor = "#633A84E6"
+buttonFontSize = 300
+
+originParams = {
+  triggeringCard = "FalseCovenant",
+  redrawnTokenType = "random",
```
Chr1Z93, line 8: Will this ever not be "random" except for Nkosi?

src/playercards/cards/WendyAdams.ttslua
```diff
@@ -0,0 +1,14 @@
...
+buttonWidth = "1100"
+buttonPosition = "70 -70 -22"
+buttonFontSize = 200
+buttonRotation = "0 0 90"
+
+originParams = {
+  triggeringCard = "Wendy",
+  redrawnTokenType = "random",
+  VALID_TOKENS = {},
+  INVALID_TOKENS = {}
```
Chr1Z93, line 11: Don't need the empty tables because we added nil handling

Chr1Z93 (changes requested):

src/core/Global.ttslua
dscarpac: leftover from Nkosi, probably

src/playercards/cards/FalseCovenant.ttslua
dscarpac: Leaving it open for future expandability.

src/core/Global.ttslua
```diff
@@ -283,6 +284,128 @@ function getTokenIndex(token)
...
+  end
+
+  returnChaosTokenToBag(returnedToken)
+  if redrawData.redrawnTokenType == "random" then
+    chaosTokens[indexOfReturnedToken] = drawChaosToken({mat = mat, drawAdditional = true, takeParameters = takeParameters})
+  else
+    chaosTokens[indexOfReturnedToken] = drawChaosToken({mat = mat, drawAdditional = true, takeParameters = takeParameters, tokenType = redrawData.redrawnTokenType})
+  end
+  
+  chaosTokens[indexOfReturnedToken] = token
```
Chr1Z93, line 397: You could condense this a little like this:
```lua
local params = { mat = mat,  drawAdditional = true,  takeParameters = takeParameters }
if redrawData.redrawnTokenType ~= "random" then
  params.tokenType = redrawData.redrawnTokenType
end
chaosTokens[indexOfReturnedToken] = drawChaosToken(params)
```

src/chaosbag/BlessCurseManagerApi.ttslua
```diff
@@ -54,5 +55,17 @@ do
     getManager().call("addMenuOptions", { playerColor = playerColor, hoveredObject = hoveredObject })
   end
   
+  -- adds bless / curse to the chaos bag
+  -- @param type string Type of chaos token ("Bless" or "Curse")
```
Chr1Z93, line 59: Missing a `-` (should be 3 for annotations) - also for the function below

src/chaosbag/ChaosBagApi.ttslua
```diff
@@ -71,9 +71,9 @@ do
   ---@param drawAdditional boolean Controls whether additional tokens should be drawn
   ---@param tokenType? string Name of token (e.g. "Bless") to be drawn from the bag
   ---@param guidToBeResolved? string GUID of the sealed token to be resolved instead of drawing a token from the bag
-  ---@param returnedToken? tts__Object Token to be replaced with newly drawn token
+  ---@param takeParameters? table Position and rotation of the location where the new token should be drawn to, usually to replace a returned token
```
Chr1Z93, line 74: Missing annotation for the new return value (object reference to the token?)

src/core/Global.ttslua
```diff
@@ -378,18 +494,18 @@ function drawChaosToken(params)
...
     local name = token.getName()
     local tokenData = mythosAreaApi.returnTokenData().tokenData or {}
     local specificData = tokenData[name] or {}
     token.setDescription(specificData.description or "")
     trackChaosToken(name, matGUID)
 
-    if params.returnedToken then
-      chaosTokens[indexOfReturnedToken] = token
+    if params.takeParameters then
+      return token
```
Chr1Z93, line 506: Should probably always return this instead of only for the parameters

Chr1Z93 (changes requested):
Could you also add a few more comments?

src/chaosbag/ChaosBagApi.ttslua
dscarpac: Looks like that's unused now with the new redraw function taking its place.

src/playercards/CardsThatRedrawTokens.ttslua
```diff
@@ -1,3 +1,62 @@
...
+
+> originParams = {
+>   triggeringCard = "ClaypoolsFurs",
+>   redrawnTokenType = "random",
+>   VALID_TOKENS = {
+>     ["Frost"] = true
+>   }
+> }
+
+require("playercards/CardsThatRedrawTokens")
```
Chr1Z93: This should give you an error because of the circular reference (that's the reason it's not spelled out in the other file)

Chr1Z93: Ah you already caught it with the next commit

Chr1Z93 (approved):

PR #722, opened 2024-06-25
New option: Enable all card helpers

Chr1Z93:
This option will cause all card helpers (cards with XML / script buttons) to be enabled by default.
Additionally, introduces a new library file for such cards.


BootleggerFinn (approved):
Looks great. What a good idea and big undertaking!

PR #751, opened 2024-07-06
Added features to Search Assistant

dscarpac:
(1) Creates an X where search targets should be dropped if you started with 0 cards in hand
(2) Allows for initiation of search on pressing a number key while hovering the object.

EDIT: Remaining issues -
(1) Snap point does not seem to be deleted via scripting (unlike the X)
(2) Rotation of X could be mathematically derived
(3) Color of X could be tied to player hand color if seated, else matColor

### DIFF
--- a/src/accessories/SearchAssistant.ttslua
+++ b/src/accessories/SearchAssistant.ttslua
@@ -34,6 +34,7 @@ inputParameters.validation      = 2
 
 function onLoad()
   normalView()
+  self.max_typed_number = 9999
 end
 
 -- regular view with search box
@@ -89,6 +90,10 @@ function updateSearchNumber(_, _, input)
   inputParameters.value = tonumber(input)
 end
 
+function onNumberTyped(player_color, number)
+  startSearch(player_color, number)
+end
+
 -- starts the search with the number from the input field
 function searchCustom(_, messageColor)
   local number = inputParameters.value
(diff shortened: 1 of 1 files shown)


src/accessories/SearchAssistant.ttslua
```diff
@@ -130,7 +135,31 @@ function startSearch(messageColor, number)
...
+  else
+    oldVectorLines = Global.getVectorLines()
+    lines = {}
+    local pos = setAsidePosition
+    if matColor == "Green" or matColor == "Red" then
+      pointsLine1 = { { pos.x - 0.5, pos.y, pos.z + 1 }, { pos.x + 0.5, pos.y, pos.z - 1 } }
+      pointsLine2 = { { pos.x - 0.5, pos.y, pos.z - 1 }, { pos.x + 0.5, pos.y, pos.z + 1 } }
+    else
+      pointsLine1 = { { pos.x - 1, pos.y, pos.z + 0.5 }, { pos.x + 1, pos.y, pos.z - 0.5 } }
+      pointsLine2 = { { pos.x - 1, pos.y, pos.z - 0.5 }, { pos.x + 1, pos.y, pos.z + 0.5 } }
```
Chr1Z93, line 149: This can very likely be simplified by using setAsideRotation

Chr1Z93 (changes requested):
Cool upgrade!

src/accessories/SearchAssistant.ttslua
dscarpac: I tried this first, but I don't think the vector lines rotate in quite the same way as the existing variable.

src/accessories/SearchAssistant.ttslua
```diff
@@ -130,7 +135,36 @@ function startSearch(messageColor, number)
   setAsideRotation = { handData.rotation.x, handData.rotation.y + 180, 180 }
 
   -- place hand cards set aside
-  deckLib.placeOrMergeIntoDeck(handCards, setAsidePosition, setAsideRotation)
+  if #handCards > 0 then
+    deckLib.placeOrMergeIntoDeck(handCards, setAsidePosition, setAsideRotation)
+  else
+    oldSnapPoints = Global.getSnapPoints()
+    newSnapPoints = oldSnapPoints
```
Chr1Z93, line 142: I don't think this works because Lua passes tables by reference, not by value

dscarpac: yeah, see my EDIT comments above -- I just left this in here as a reminder to me/you to figure out how to make this work ;)

PR #771, opened 2024-07-14
Added clue replenishing to hotkey

Chr1Z93:
- now properly places clue tokens that are spawned with numpad 8
- updated resource spawning for numpad 9

### DIFF
--- a/src/core/GameKeyHandler.ttslua
+++ b/src/core/GameKeyHandler.ttslua
@@ -276,6 +276,12 @@ function removeOneUse(playerColor, hoveredObject)
     end
   end
 
+  -- error handling
+  if not targetObject then
+    broadcastToColor("No tokens found!", playerColor, "Yellow")
+    return
+  end
+
   -- release sealed token if card has one and no uses
   if tokenChecker.isChaosToken(targetObject) and hoveredObject.hasTag("CardThatSeals") then
     local func = hoveredObject.getVar("releaseOneToken") -- check if function exists
@@ -283,12 +289,6 @@ function removeOneUse(playerColor, hoveredObject)
       hoveredObject.call("releaseOneToken", playerColor)
       return
     end
-  end
-
-  -- error handling
-  if not targetObject then
-    broadcastToColor("No tokens found!", playerColor, "Yellow")
-    return
   end
 
   -- handling for stacked tokens
--- a/src/core/PlayArea.ttslua
+++ b/src/core/PlayArea.ttslua
@@ -464,7 +464,7 @@ end
 
 -- Count victory points from locations in play area
 ---@param highlightOff boolean True if highlighting should be enabled
----@return. Returns the total amount of VP found in the play area
+---@return number totalVP Total amount of VP found in the play area
 function countVP(highlightOff)
   local totalVP = 0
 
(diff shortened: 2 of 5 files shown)


dscarpac:
Two issues I ran into:
(1a) With a clickable resource counter already on a card, moving a resource onto the card manually, and then hitting "9" removes the resource token but does not add to the count: e.g. "14" plus one resource token, then hitting "9" gives "15" on the counter, when 16 is expected. (Note that without the resource counter, this gives the expected behavior when manually adding a resource token.
(1b) When the use type is something else, like "charge", manually putting another charge onto a card that already has a clickable counter does _not_ remove the charge token, unlike (1a)

(2) Adding more than ~16 clue tokens to a location causes all of the clues to vibrate and dance around on top of the card. Any more than 20 clues spawns them at the bottom left of the card instead. Perhaps we can convert these to clickable clue counters, just like the resource tokens?

Chr1Z93:
> (1a) With a clickable resource counter already on a card, moving a resource onto the card manually, and then hitting "9" removes the resource token but does not add to the count: e.g. "14" plus one resource token, then hitting "9" gives "15" on the counter, when 16 is expected. (Note that without the resource counter, this gives the expected behavior when manually adding a resource token.
> (1b) When the use type is something else, like "charge", manually putting another charge onto a card that already has a clickable counter does _not_ remove the charge token, unlike (1a)

I don't think it should be intended to combine existing resource tokens with a clickable counter. Imo the existing resource token should not get deleted, because the counter takes precedence. (Should be easy to add that)

> (2) Adding more than ~16 clue tokens to a location causes all of the clues to vibrate and dance around on top of the card. Any more than 20 clues spawns them at the bottom left of the card instead. Perhaps we can convert these to clickable clue counters, just like the resource tokens?

I'd like to avoid having clickable clue counters on locations since that will effect a lot of other code (like "Take a clue" hotkey and VP counting. How about spawning >16 clues on top of the existing ones instead? Here's something related: https://github.com/argonui/SCED/issues/390

dscarpac:
> > (1a) With a clickable resource counter already on a card, moving a resource onto the card manually, and then hitting "9" removes the resource token but does not add to the count: e.g. "14" plus one resource token, then hitting "9" gives "15" on the counter, when 16 is expected. (Note that without the resource counter, this gives the expected behavior when manually adding a resource token.
> > (1b) When the use type is something else, like "charge", manually putting another charge onto a card that already has a clickable counter does _not_ remove the charge token, unlike (1a)
> 
> I don't think it should be intended to combine existing resource tokens with a clickable counter. Imo the existing resource token should not get deleted, because the counter takes precedence. (Should be easy to add that)
> 
> > (2) Adding more than ~16 clue tokens to a location causes all of the clues to vibrate and dance around on top of the card. Any more than 20 clues spawns them at the bottom left of the card instead. Perhaps we can convert these to clickable clue counters, just like the resource tokens?
> 
> I'd like to avoid having clickable clue counters on locations since that will effect a lot of other code (like "Take a clue" hotkey and VP counting). How about spawning >16 clues on top of the existing ones instead? Here's something related: #390

I'm fine with your solution to (2), maybe wrapping up #390 in the process.

For (1), as long as these are consistent, that's much better, but I'm also thinking about ways to merge the clickable resource counters and tokens that get spawned in other ways, not just a manually placed resource or charge. For example, dealing with the tokens (or the single clickable resource counter) that spawn with Extra Ammunition and the like. If you move those tokens over to a weapon that already has tons of tokens on it, it'd be nice to accommodate that without the extra mess.

dscarpac (approved):

PR #774, opened 2024-07-19
Added arkham.build support

Chr1Z93:
Also features some reformatting


dscarpac (changes requested):
(1) Deck Importer should give mouseover text to indicate Arkham.build support in the text fields
(2) Deck IDs are too long to fit in the text field
image

dscarpac (approved):

PR #779, opened 2024-07-25
DeckLib: Flip cards instantly

Chr1Z93:
This way the cards are rotated before they leave the hand


dscarpac:
Was this fixing some weird behavior or just inconsistencies in the code?

Chr1Z93:
> Was this fixing some weird behavior or just inconsistencies in the code?

Without this fix other players can briefly see cards that the search assistant sets aside because they don't rotate before leaving the hand

dscarpac (approved):

PR #781, opened 2024-07-26
Bugfixes for "Remove One Use" hotkey

Chr1Z93:
- various bugfixes relating to empty / non-matching "uses"
- updated feedback message
- also updated feedback message logic for "Take a clue"
- fixed a bug for spawning a resource on cards without "uses"


dscarpac (changes requested):
The wording on the new message is a bit awkward. "Removed a Charge" is OK, but "Removed an Ammo" isn't quite right in English, it'll need to be "an Ammo token" or something similar. Not sure what was wrong with the old way, which is a little less awkward sounding between the game language and English.

Another one: the name of the resource counter is "Resourcecounter" without a space.

Chr1Z93:
> The wording on the new message is a bit awkward. "Removed a Charge" is OK, but "Removed an Ammo" isn't quite right in English, it'll need to be "an Ammo token" or something similar. Not sure what was wrong with the old way, which is a little less awkward sounding between the game language and English.
> 
> Another one: the name of the resource counter is "Resourcecounter" without a space.

There was nothing wrong with it but it didn't allow adding the location easily.
I've updated it.

dscarpac (approved):

PR #782, opened 2024-07-26
Added Playermat-specific Options

Chr1Z93:
This adds a new playermat-specific option panel to control some things:
- texture specifying
- slot editing
- hand color choosing

The "old" slot editing button was turned into the "Discard 1" button from the Hand Helper

To-Do:
- [x] update code to allow for more than one XML
- [x] design the option panel itself
- [x] dynamically create the option panel
- [x] save / load functionality


dscarpac:
Might be reasonable to add some options for Multi-Handed in here -- something like "make this player's hand visible" so people don't have to go into the TTS options menu. We sometimes get requests for 2x 2-Handed as well, so maybe we could restrict the visibility of the hand to only some players. (e.g. make this hand visible to "Orange" or a player's steam name or something)

Chr1Z93:
> Might be reasonable to add some options for Multi-Handed in here -- something like "make this player's hand visible" so people don't have to go into the TTS options menu. We sometimes get requests for 2x 2-Handed as well, so maybe we could restrict the visibility of the hand to only some players. (e.g. make this hand visible to "Orange" or a player's steam name or something)

That's a good idea but we don't have control over hand visibility.

dscarpac:
> > Might be reasonable to add some options for Multi-Handed in here -- something like "make this player's hand visible" so people don't have to go into the TTS options menu. We sometimes get requests for 2x 2-Handed as well, so maybe we could restrict the visibility of the hand to only some players. (e.g. make this hand visible to "Orange" or a player's steam name or something)
> 
> That's a good idea but we don't have control over hand visibility.

Rats.

dscarpac (approved):

Issue #764, opened 2024-07-08
Request: Update Nkosi Mabati helper


Chr1Z93:
It needs a "choose sigil" button that enables together with other helpers.

PR #787, opened 2024-07-31
Added helper to Nkosi

dscarpac:
Puts card behavior in line with other helpers using CardsWithHelper functions.

#764

### DIFF
--- a/src/playercards/cards/NkosiMabati3.ttslua
+++ b/src/playercards/cards/NkosiMabati3.ttslua
@@ -1,4 +1,9 @@
 require("playercards/CardsWithHelper")
+
+-- intentionally global
+hasXML                     = false
+isHelperEnabled            = false
+
 local chaosBagApi = require("chaosbag/ChaosBagApi")
 
 -- XML background color for each token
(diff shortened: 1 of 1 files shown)


src/playercards/cards/NkosiMabati3.ttslua
```diff
@@ -14,17 +19,22 @@ local tokenColor  = {
...
+    isHelperEnabled = isHelperEnabled,
+    sigil = sigil
+  })
 end
 
 function onLoad(savedData)
-  self.addContextMenuItem("Enable Helper", chooseSigil)
-  sigil = JSON.decode(savedData)
-  if sigil and sigil ~= nil then
+  syncDisplayWithOptionPanel()
```
Chr1Z93, line 30: The sync should be called at the end of onLoad()

src/playercards/cards/NkosiMabati3.ttslua
```diff
@@ -59,11 +69,35 @@ function makeXMLButton()
   )
 end
 
+-- Create XML button to prompt choosing a sigil; acts as this card's helper
+function initialize()
+  self.UI.setXmlTable({
```
Chr1Z93, line 74: There should be a condition: If there's already a sigil defined (e.g. because it was saved and loaded), this should skip the "choose sigil" XML

Chr1Z93 (changes requested):

Chr1Z93 (approved):
Looks good to me!

PR #800, opened 2024-08-03
Added conditional hand hiding

Chr1Z93:
There's now an option in the playermat options to toggle hand hiding for you for that mat


BootleggerFinn:
Tested this, it works fantastic!

BootleggerFinn (approved):

PR #817, opened 2024-08-15
Increased scale of counters on playermats

dscarpac:
image
image


Chr1Z93:
Good idea with the increased size, but I'd like to keep them just slightly smaller. How about scale of 0.3 (instead of 0.32)? Clue Counter seems fine at 0.45

Chr1Z93 (approved):

PR #827, opened 2024-08-16
Moved side scenario spawn position one row higher

Chr1Z93:
Now you can spawn the side scenarios while a campaign is placed.
Some campaigns might have stuff placed up there (for example the map in TSK), which we might want to adjust.


dscarpac (approved):
Merging, but would like to see what the new layout looks like for e.g. the map in TSK.

PR #833, opened 2024-08-17
Moved DES and Well Connected to XML helpers

dscarpac:
Removed counting from Hand Helper and moved it to a helper on the card itself. Converted Well Connected to XML.

image

image


Chr1Z93 (changes requested):
The code looks good to me! Just the visuals of the XML could be better imo. Very good work so far though :)

1) Well Connected: This should be bigger and better readable

2) DES: I think this should be smaller and a different color, the panel looks like fog on the card

Chr1Z93 (approved):

PR #836, opened 2024-08-17
Updated damage and horror tokens

Chr1Z93:
- removed states in favor of an XML button / label
- enabled clickability for locked tokens
- updated Tokenmanager to spawn tokens directly with the correct script state instead of callbacks


dscarpac:
I like these but
(1) the button area should be a little smaller because it's still easy to click the button area when you want to move them
(2) mixed feelings about no "1" state

Chr1Z93:
> I like these but (1) the button area should be a little smaller because it's still easy to click the button area when you want to move them

Did you notice that the button will be unclickable after a 2s hover? Don't think the button should be smaller

> (2) mixed feelings about no "1" state

I'm open about changing it, but this way it's gonna be a less disruptive change since most of the time single tokens will always be in use first (and then you don't directly run into the button)

dscarpac:
> > I like these but (1) the button area should be a little smaller because it's still easy to click the button area when you want to move them
> 
> Did you notice that the button will be unclickable after a 2s hover? Don't think the button should be smaller
> 

I did notice this after posting, is this what causes there to sometime be a green outline? Regardless, I think 2 seconds is way too long if this hover-wait-move is the intended behavior to move the tokens. Might take some more playtesting / beta channel help but as it is I found this to be a bit frustrating.

Chr1Z93:
> > > I like these but (1) the button area should be a little smaller because it's still easy to click the button area when you want to move them
> > 
> > 
> > Did you notice that the button will be unclickable after a 2s hover? Don't think the button should be smaller
> 
> I did notice this after posting, is this what causes there to sometime be a green outline? Regardless, I think 2 seconds is way too long if this hover-wait-move is the intended behavior to move the tokens. Might take some more playtesting / beta channel help but as it is I found this to be a bit frustrating.

This is indeed the intended behaviour. How often do you need to move a damage token on count 2 or above?

dscarpac:
> > > > I like these but (1) the button area should be a little smaller because it's still easy to click the button area when you want to move them
> > > 
> > > 
> > > Did you notice that the button will be unclickable after a 2s hover? Don't think the button should be smaller
> > 
> > 
> > I did notice this after posting, is this what causes there to sometime be a green outline? Regardless, I think 2 seconds is way too long if this hover-wait-move is the intended behavior to move the tokens. Might take some more playtesting / beta channel help but as it is I found this to be a bit frustrating.
> 
> This is indeed the intended behaviour. How often do you need to move a damage token on count 2 or above?

Not a ton, especially if you use the discard buttons/hotkeys, but certainly not all of our users do. There are plenty of "heal 2" effects that happen and I am sure most users just drop those tokens in the trash. I see your point and agree that these can be something like the resource counters, so there is precedent, but I think this is a big enough change to the simple idea that damage/horror have been moveable tokens forever that I need to see some more feedback from others and playtesting of actual scenarios myself before I can support these exact parameters. (I want to try this out with 1s of hover instead of 2 to see if that feels better).

dscarpac (approved):

PR #843, opened 2024-08-18
Added images to image folder

Chr1Z93:
These are no directly part of the mod, but just in case I get hit by a bus it's better to have this available ;)


Chr1Z93:
Now that I think about it, this should probably be a separate repo

Chr1Z93:
https://github.com/Chr1Z93/SCED-tools

PR #856, opened 2024-09-09
Added story assets from TMG to index

dscarpac:
Additionally adds the Jewel of Sarnath to set aside area on deck import (must be added to a deck first)


Chr1Z93 (commented):
Didn't check everything again, but it looks good from spot checking

PR #868, opened 2024-09-13
Add The Drowned City

Chr1Z93:
- [x] add campaign placeholder
- [x] update table layout for additional box
- [x] add button to PlayerCard panel
- [x] add data for PlayerCard panel

Closes: https://github.com/argonui/SCED/issues/858

updated layout:
image


dscarpac:
(1) Add an error message for downloading The Drowned City campaign since we're so far out that says something like "this product has not been released" to avoid confusion
(2) Move Innsmouth to the empty spot next to Dream Eaters, so the row below contains all four campaigns with "big" boxes: EOTE, TSK, FOHV, and TDC.

Chr1Z93:
> (1) Add an error message for downloading The Drowned City campaign since we're so far out that says something like "this product has not been released" to avoid confusion (2) Move Innsmouth to the empty spot next to Dream Eaters, so the row below contains all four campaigns with "big" boxes: EOTE, TSK, FOHV, and TDC.

Both done

dscarpac (approved):

PR #869, opened 2024-09-14
Added new font and flippable capability to damage / horror tokens

dscarpac:
image

Default orientation of tokens matches the investigator and player cards with health / sanity.


objects/Damagetokens.480bda/Damage.cd2a02.json
Chr1Z93, line 1: Since they are flippable, we don't need to change the URL for these

objects/Horrortokens.7bd2a0.json
```diff
@@ -18,7 +18,7 @@
     "CastShadows": true,
     "ColliderURL": "",
     "Convex": true,
-    "DiffuseURL": "https://steamusercontent-a.akamaihd.net/ugc/1758068501357109766/6D35F3D98A4C89696895FACA788F54005B1688D3/",
+    "DiffuseURL": "https://steamusercontent-a.akamaihd.net/ugc/2424714310447293403/23463082E4C79C0428E2B63C6DC4B6F46CDC9B7B/",
```
Chr1Z93, line 21: Why is this image rotated like that?

src/core/DamageHorrorCounter.ttslua
```diff
@@ -77,12 +77,19 @@ function createXml()
   self.UI.setXmlTable(xml)
 end
 
--- rotate to face-up when dropped
-function onDrop()
-  self.setRotation(self.getRotation():setAt("z", 0))
+function onRotate(_, flip, _, _, oldFlip)
```
Chr1Z93, line 80: How about just putting the button on both sides?

src/core/DamageHorrorCounter.ttslua
```diff
@@ -97,6 +104,11 @@ end
 
 function updateLabel()
   local uiColor, uiVal = getUiColorAndVal()
+  if val > 9 then
+    self.UI.setAttribute("xmlTxt", "fontSize", "350")
+  else
+    self.UI.setAttribute("xmlTxt", "fontSize", "700")
+  end
```
Chr1Z93, line 111: Imo this should be part of the function that gathers the data for the UI (`getUiColorAndVal()`). Would of course need a rename then.

Chr1Z93 (changes requested):
Also, is it intentional that the health + sanity on playermats stays as is?

Chr1Z93 (approved):

PR #872, opened 2024-09-16
Bugfix: Fixed visibility of playermat options panel on load

dscarpac:


src/playermat/Playermat.ttslua
```diff
@@ -916,6 +916,7 @@ end
 
 function onClick_hideOrShowOptions(player)
   GlobalApi.changeWindowVisibility(player.color, "optionPanelMain", _, self)
+  Wait.time(function() updateSave() end, 1)
```
Chr1Z93, line 919: I think we want to narrow this down further and use frames as delay (afaik 3 is sufficient)

dscarpac: fine with me

Chr1Z93 (approved):

PR #877, opened 2024-09-18
Added a "shuffle discard into deck" button to playermats

Chr1Z93:
image


objects/Playermat1White.8b081b.json
```diff
@@ -342,28 +342,6 @@
     "ImageURL": "https://steamusercontent-a.akamaihd.net/ugc/2462982115659543571/5D778EA4BC682DAE97E8F59A991BCF8CB3979B04/",
     "WidthScale": 0
   },
-  "CustomUIAsset": [
```
Chr1Z93, line 345: Turns out we don't need to have these double declared (they are also in the global assets and for some reason that's enough?)

dscarpac (approved):

Issue #854, opened 2024-09-03
Bug: Tour doesn't display correctly on some screen resolutions


Chr1Z93:
Might be able to turn the positioning of elements into relative values instead of hardcoded offsets

PR #878, opened 2024-09-19
Adapted tour and controls for unusual resolutions

Chr1Z93:
Tour has been updated to use relative coordinates, control bar was increased in roughly 10% size

closes https://github.com/argonui/SCED/issues/854


dscarpac (changes requested):
I see this PR is mostly about fixing double spaces :-D.

We need to change a few things in the tour for 4.0, probably, but before I forget, let's change Daisy's script from talking about a non-existent newspaper. Preston can also mention clicking on damage/horror tokens to increase their value.

Chr1Z93:
> I see this PR is mostly about fixing double spaces :-D.
> 
> We need to change a few things in the tour for 4.0, probably, but before I forget, let's change Daisy's script from talking about a non-existent newspaper. Preston can also mention clicking on damage/horror tokens to increase their value.

It's mostly about changing the positioning of XML elements relative to screen resolution instead of hard-coding it.

Would prefer to not mix changes like that - this PR is about the functionality of the tour, not the "tour script".

dscarpac (approved):

PR #881, opened 2024-09-23
Added Return All Sealed Tokens option

dscarpac:
Specific to Radiant Smite


src/playercards/CardsThatSealTokens.ttslua
```diff
@@ -289,20 +297,30 @@ function releaseAllTokens(playerColor)
...
 
+-- returns all sealed tokens to the token pool
+function returnAllTokens(playerColor)
+  printToColor("Returning " .. #sealedTokens .. " tokens to the token pool", playerColor)
+  for i = 1, #sealedTokens do
+    returnToken(table.remove(sealedTokens))
+  end
+  updateSave()
+  Player[playerColor].clearSelectedObjects()
+end
```
Chr1Z93, line 322: Can you explain why we need a new function (why is `returnMultipleTokens` not sufficient?)

Chr1Z93 (approved):

Issue #518, opened 2023-12-22
Request: Update the tour


Chr1Z93:
Some recent additions / changes should be pointed out by the tour.

Chr1Z93:
Final tour card (decoupled from the rest of the tour) with investigator specific information,
e.g. Wendy: "There is a gamekey for adding sealing options to cards"

Chr1Z93:
Investigator specific things have been opened as separate issue here: https://github.com/argonui/SCED/issues/862

PR #882, opened 2024-09-24
Tour: updated script and added back button

Chr1Z93:
"Back" button goes back to the previous card in case you missed something.

Closes https://github.com/argonui/SCED/issues/518

### DIFF
--- a/modsettings/CustomUIAssets.json
+++ b/modsettings/CustomUIAssets.json
@@ -97,7 +97,12 @@
   {
     "Name": "NextArrow",
     "Type": 0,
-    "URL": "https://i.imgur.com/MztSQis.png"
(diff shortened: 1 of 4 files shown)


dscarpac (changes requested):
Back arrow should be referenced in the tour: "Use the arrows to move forward and backward"

dscarpac:
> > Back arrow should be referenced in the tour: "Use the arrows to move forward and backward"
> 
> Are you sure? I think they are pretty intuitive

In that case no arrows should be mentioned at all? I'm just referring to the Roland text which was clearly written before there were two arrows.

Chr1Z93:
> > > Back arrow should be referenced in the tour: "Use the arrows to move forward and backward"
> > 
> > 
> > Are you sure? I think they are pretty intuitive
> 
> In that case no arrows should be mentioned at all? I'm just referring to the Roland text which was clearly written before there were two arrows.

Sorry, I missed that Roland references them. Updated it

dscarpac (approved):

Issue #870, opened 2024-09-15
Request: XML pop-out for Token Manager(s)


dscarpac:
Have a button on screen that loads the Bless / Curse manager GUI and possibly the Token Manager.

Chr1Z93:
Bless / Curse Manager UI is added via linked PR.
Imo the Token Arranger could eventually be reworked, but it's gonna be a tremendous amount of work to condense it somehow. So for now that will not be on my radar.

PR #883, opened 2024-09-27
Added a global XML UI for the Bless/Curse Manager

Chr1Z93:
This is in addition (and in sync) with the existing Bless / Curse Manager.
We could remove it at a later point, but I am not sure yet if we want to force the global XML on everyone.
It features an "accordion" button to collapse / expand it.

Closes https://github.com/argonui/SCED/issues/870

image

### DIFF
--- a/xml/Global/BottomBar.xml
+++ b/xml/Global/BottomBar.xml
@@ -1,20 +1,28 @@
 <Defaults>
   <Button class="navbar"
+    tooltipOffset="-135"
     tooltipPosition="Left"
     tooltipBackgroundColor="rgba(0,0,0,1)"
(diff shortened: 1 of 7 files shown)


xml/Global/BottomBar.xml
```diff
@@ -1,20 +1,28 @@
 <Defaults>
   <Button class="navbar"
+    tooltipOffset="-135"
```
dscarpac, line 3: Looks like this is too far away (if necessary at all?) compared to the position of the buttons.

dscarpac (changes requested):

dscarpac (approved):

Issue #862, opened 2024-09-10
Spawn investigator specific things


Chr1Z93:
Example: 
- Darrell - bag with evidence tokens
- Tony - bag with bounty tokens
- parallel Wendy - notecard with hotkey explanation 
etc.

PR #887, opened 2024-09-30
Added an investigator placing callback

Chr1Z93:
This allows us to send specific instructions for some investigators or even spawn a specialized tool.

At this point the following investigators get something:
- parallel Wendy: Message
- Norman: Message
- Suzi: Helper spawned
- Darrell: Evidence bag spawned
- Tony: Bounty bag spawned

PR includes:
- bugfix for Upkeep function (stopped on first card in investigator area)
- new function for Token Manager to generate data to spawn an infinite bag for a specific token
- new "SetAside7" zone which is below the regular set aside area (used for the investigator specific stuff)
- Suzi helper was moved to option panel source (despite not being spawnable via the option panel, maybe we should rename it to just "source bag")
- update to the "messageColor" feature of the playermat to make sure it'll always be able to get a color of a seated player

Closes: https://github.com/argonui/SCED/issues/862


src/playermat/Playermat.ttslua
```diff
@@ -341,19 +348,14 @@ function doUpkeep(_, clickedByColor, isRightClick)
...
-    elseif obj.type == "Card" then      
-      if obj.hasTag("DoInUpkeep") then
-        obj.call("doInUpkeep")
-      end
-      -- do not rotate, replenish, etc. on cards in investigator card area
-      if inArea(self.positionToLocal(obj.getPosition()), INVESTIGATOR_AREA) then
-        break
-      end
-      
+    elseif obj.hasTag("DoInUpkeep") and obj.is_face_down then
```
dscarpac, line 355: I kept this from being a separate elseif in case of a future card that has the DoInUpkeep tag and still has uses to replenish, to be rotated, etc.

dscarpac (changes requested):

src/playermat/Playermat.ttslua
Chr1Z93: That's fair, I'll update it later!

dscarpac (approved):

PR #891, opened 2024-10-02
Investigator Callback: Added 2nd zone for token bags

Chr1Z93:
This new zone is directly above set aside 1, in the same "row" as the Hand Helper / Search Assistant / Stat Tracker.


dscarpac:
The zones for the lateral seats is a little bit too close to the playmat; e.g.
image

These bags go away with the clean-up helper since they are detected as being on the playmat.

Chr1Z93:
Updated the position

image

dscarpac (approved):

PR #922, opened 2024-10-24
Updated Chaos Token stat tracker and added VS Code build task

Chr1Z93:
New: Middle click to just print your own stats
image

New: description since XML tooltips suck
image

New: history for the last 5 tokens (with the color of the currently seated player at that mat)
image

New: build task for VS Code to compile the current state of the local mod repo (requires cloning the TTS Mod Manager - could be adapted to use the .exe too)
image


dscarpac:
Someone asked in the Discord about having this print out the last drawn token in case you click too quickly or if you forget? Maybe the last three drawn for each player?

Chr1Z93:
> Someone asked in the Discord about having this print out the last drawn token in case you click too quickly or if you forget? Maybe the last three drawn for each player?

That could indeed be useful. That's unrelated to this change though, so it should be a separate PR.

dscarpac (approved):

PR #933, opened 2024-10-28
Added Parallel Mateo

Chr1Z93:
Rudimentary sealing options for now, can be improved later


dscarpac:
I realize this is last minute addition to the mod and requires other changes than co-opting the || Wendy functions, but it'd be nice to use the ability to resolve the sealed tokens on investigator cards. Advanced Serpents should also be able to seal tokens from other cards directly rather than releasing them on those cards and then sealing from the bag.

At the very least for now, let's fix the rotation on the sealed Bless token so it matches the other sealed cards.

image

Chr1Z93:
> I realize this is last minute addition to the mod and requires other changes than co-opting the || Wendy functions, but it'd be nice to use the ability to resolve the sealed tokens on investigator cards. Advanced Serpents should also be able to seal tokens from other cards directly rather than releasing them on those cards and then sealing from the bag.
> 
> At the very least for now, let's fix the rotation on the sealed Bless token so it matches the other sealed cards.

Updated the rotation, but don't really feel like doing more in this PR. Feel free to pick it up though

dscarpac (approved):

PR #939, opened 2024-10-29
Fixed card drawing with reshuffle

Chr1Z93:


src/core/Global.ttslua
```diff
@@ -225,7 +225,11 @@ function onObjectEnterZone(zone, object)
   elseif zone.type == "Hand" and object.type == "Card" then
     -- make sure the card is face-up
     if object.is_face_down then
-      object.flip()
+      if object.held_by_color then
+        object.held_flip_index = 0
+      else
+        object.flip()
+      end
```
Chr1Z93, line 232: Cards moved to the hand still flip, but if they are held by a player, they keep it held instead of losing control.

src/playermat/Playermat.ttslua
```diff
@@ -252,13 +252,7 @@ end
...
 function searchDeckAndDiscardArea(filter)
   local pos = self.positionToWorld(DECK_DISCARD_AREA.center)
-  local scale = self.getScale()
-  local size = {
-    x = DECK_DISCARD_AREA.size.x * scale.x,
-    y = DECK_DISCARD_AREA.size.y,
-    z = DECK_DISCARD_AREA.size.z * scale.z
-  }
-  return searchArea(pos, size, filter)
+  return searchArea(pos, DECK_DISCARD_AREA.size * self.getScale(), filter)
```
Chr1Z93, line 255: No functional change, just implemented Vector multiplication to make it more concise

src/playermat/Playermat.ttslua
```diff
@@ -595,11 +604,11 @@ function getDeckAreaObjects()
 end
 
 -- draws the specified number of cards (reshuffling of discard pile is handled separately)
+---@param cardOrDeck tts__Object Card/Deck to draw from
```
Chr1Z93, line 607: Re-using the existing reference since this is only called when the deck area objects are already determined

src/playermat/Playermat.ttslua
```diff
@@ -651,8 +664,9 @@ end
 function flipTopCardFromDeck()
   Wait.time(function()
     local deckAreaObjects = getDeckAreaObjects()
-    if deckAreaObjects.topCard then
-    elseif deckAreaObjects.draw then
+    if deckAreaObjects.topCard ~= nil then return end
+
+    if deckAreaObjects.draw ~= nil then
```
Chr1Z93, line 669: No functional change, just made the conditions clearer

src/playermat/Playermat.ttslua
```diff
@@ -1711,3 +1725,18 @@ end
...
+  end
+end
+
+-- pauses the current coroutine for 'seconds' seconds
+function coWaitSeconds(seconds)
+  local startTime = os.clock()
+  while os.clock() - startTime < seconds do
+    coroutine.yield(0)
+  end
+end
```
Chr1Z93, line 1742: Helper functions for coroutine pausing

src/core/Global.ttslua
dscarpac: Not sure about this change. Did some testing and dragging a card to hand doesn't flip the card when I was hoping/expecting it. What's the case where this is an improvement?

Chr1Z93: I don't understand? It's still just flipping the card, just that you don't lose control

dscarpac: I'll try to make a video later today/tomorrow and send it to you, but it's only flipping about 75%-80% of the time.

dscarpac (approved):

PR #965, opened 2024-11-05
Fix player hand zone being upside down after using the rotate player mat function

max-torch:
fixes the issue that was reported here
https://discord.com/channels/704806269923491941/704806269923491944/1302241548611620976

### DIFF
--- a/src/playermat/PlayermatApi.ttslua
+++ b/src/playermat/PlayermatApi.ttslua
@@ -410,7 +410,7 @@ do
       -- offset the rotation by 180 degrees if the guid matches the player hand zones
       local finalRotationY = rotationY
       if obj.type == "Hand" then
-        newRotationY = rotationY - 180
+        finalRotationY = rotationY + 180
       end
 
       obj.setRotation(obj.getRotation():setAt("y", finalRotationY))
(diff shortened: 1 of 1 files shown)


src/playermat/PlayermatApi.ttslua
```diff
@@ -406,7 +406,14 @@ do
     for guid, pos in pairs(storedPositions) do
       local obj = getObjectFromGUID(guid)
       obj.setPosition(mat.positionToWorld(pos))
-      obj.setRotation(obj.getRotation():setAt("y", rotationY))
+
+      -- offset the rotation by 180 degrees if the guid matches the player hand zones
+      local finalRotationY = rotationY
+      if guid == "0285cc" or guid == "a70eee" or guid == "5fe087" or guid == "be2f17" then
```
Chr1Z93, line 412:
```suggestion
      if obj.type == "Hand" then
```
Would not rely on GUIDs - I think checking the type is better

src/playermat/PlayermatApi.ttslua
```diff
@@ -406,7 +406,14 @@ do
...
+        newRotationY = rotationY - 180
```
Chr1Z93, line 413:
```suggestion
        finalRotationY = rotationY + 180
```
Seems like you used the wrong variable name? Also I am not sure if TTS handles negative rotations well 😅

Chr1Z93 (changes requested):

Chr1Z93 (approved):
Looks good to me!

PR #993, opened 2024-11-16
Bug Fixes to the Victory Display

Entrox-Licher:
- Fixed Victory Display not updating when cards deleted from the Victory Display
- Fixed new cards getting added to Victory Points not being highlighted if highlighting was on
- Fixed cards getting removed from Victory Display not getting highlighting turned off

### DIFF
--- a/src/core/VictoryDisplay.ttslua
+++ b/src/core/VictoryDisplay.ttslua
@@ -77,6 +77,7 @@ end
 
 -- removing an object from the victory display
 function onCollisionExit()
+  log("Stopped Colliding!")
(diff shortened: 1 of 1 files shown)


src/core/VictoryDisplay.ttslua
```diff
@@ -77,6 +77,7 @@ end
 
 -- removing an object from the victory display
 function onCollisionExit()
+  log("Stopped Colliding!")
```
dscarpac, line 80: Log statement left in?

Chr1Z93 (approved):

PR #994, opened 2024-11-17
Fixed 'Discard 1' Button Flickering

Entrox-Licher:
The 'Discard 1' button on the green and red playermats was flickering because it was partially clipping into the object at certain camera angles.  Just slightly increased the button height to fix the issue

### DIFF
--- a/src/playermat/Playermat.ttslua
+++ b/src/playermat/Playermat.ttslua
@@ -210,6 +210,7 @@ function onLoad(savedData)
   buttonParameters.label = "Discard 1"
   buttonParameters.click_function = "doDiscardOne"
   buttonParameters.tooltip = "Discard one random card from hand (hidden cards are excluded)."
+  buttonParameters.position.y = 0.15
   buttonParameters.position.z = 0.92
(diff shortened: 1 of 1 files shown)


Chr1Z93:
This is very weird, all buttons should be outside of the playermat. And then even if this was an issue, we should increase it for all of the buttons and not just the discard button.

Entrox-Licher:
This is a recurring TTS issue I encounter, at it seems to depend on machine settings or hardware specifications, in addition to view angles and the like.  In my experience it is _usually_ buttons near the bottom of objects for some reason (likely a view angle thing). 

I did a sweep through all the buttons in mod and didn't find any others with issues, and I didn't want to do a big change if I couldn't reproduce issues.  That said, in my other mods I do tend to default all my buttons' y position to 0.2 for this reason

Chr1Z93:
Oh that's interesting! Haven't heard of this before. I'd still prefer to change this for the shared parameters (LOC 97).

Chr1Z93:
Fixed here: https://github.com/argonui/SCED/pull/1018

PR #1010, opened 2024-11-19
Added "Token Draw Dummys" - typing numbers to draw x tokens

Chr1Z93:
They are hidden beneath the "Draw Chaos Token" buttons and accept any number from 1-9 (and basically right-click that many times).


dscarpac:
Slick! Should we add a tool tip to the draw chaos token button or no?

Also, should add this to the version notes.

dscarpac (approved):

PR #1014, opened 2024-11-20
handled fixing token position on card flip for moving with tokens

YumiWhellie:
Tokens will now be on top of the card after you move a card with tokens on it


Chr1Z93:
Closing this for now (might revisit later)

PR #1029, opened 2024-11-23
Implemented gallery as custom UI assets

Chr1Z93:
Need to test in MP, but this could fix the caching issue for the gallery


Chr1Z93:
Doesn't work well in MP, so I'm gonna close this

PR #1034, opened 2024-12-01
Added handling for discarding cloned chaos tokens

Chr1Z93:
This shouldn't happen normally, but if someone uses the discard hotkey on a temporary chaos token, this destroys the token instead of returning it to the chaos bag.

### DIFF
--- a/src/core/GameKeyHandler.ttslua
+++ b/src/core/GameKeyHandler.ttslua
@@ -134,23 +134,29 @@ function discardObject(playerColor, hoveredObject)
 end
 
 -- actually performs the discarding of the object and tokens / tiles on it
-function performDiscard(playerColor, hoveredObject)
+function performDiscard(playerColor, targetObject)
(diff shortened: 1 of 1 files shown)


src/core/GameKeyHandler.ttslua
```diff
@@ -134,23 +134,29 @@ function discardObject(playerColor, hoveredObject)
 end
 
 -- actually performs the discarding of the object and tokens / tiles on it
-function performDiscard(playerColor, hoveredObject)
+function performDiscard(playerColor, targetObject)
+  -- destroy temporary chaos tokens instead
```
dscarpac, line 138: is this going to trigger just for chaos tokens, or temporary action tokens, too?

Chr1Z93: Just for chaos tokens to stop them from entering the chaos bag this way

dscarpac (approved):

PR #1053, opened 2024-12-15
Updated scenario card bag to replace cards inside decks

Chr1Z93:


dscarpac (approved):
I'll trust you on this one.

PR #1072, opened 2025-01-11
Bugfix: Upkeep button should ignore investigator cards

dscarpac:
Prevents e.g. Hank's set aside bonded cards from rotating.

### DIFF
--- a/src/playermat/Playermat.ttslua
+++ b/src/playermat/Playermat.ttslua
@@ -394,8 +394,8 @@ function doUpkeep(_, clickedByColor, isRightClick)
       obj.call("doInUpkeep")
     end
 
-    if obj.type == "Card" and not inArea(self.positionToLocal(obj.getPosition()), INVESTIGATOR_AREA) and not obj.hasTag("Investigator") then
-      -- do not continue for cards in investigator card area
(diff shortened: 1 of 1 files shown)


src/playermat/Playermat.ttslua
```diff
@@ -394,7 +394,7 @@ function doUpkeep(_, clickedByColor, isRightClick)
       obj.call("doInUpkeep")
     end
 
-    if obj.type == "Card" and not inArea(self.positionToLocal(obj.getPosition()), INVESTIGATOR_AREA) then
+    if obj.type == "Card" and not inArea(self.positionToLocal(obj.getPosition()), INVESTIGATOR_AREA) and not obj.hasTag("Investigator") then
```
Chr1Z93, line 397: Should removed the investigator area check now (it's surpassed by the tag check)

Chr1Z93 (changes requested):

src/playermat/Playermat.ttslua
Chr1Z93:
```suggestion
    if obj.type == "Card" and not obj.hasTag("Investigator") then
```

PR #1082, opened 2025-01-12
Updated patch notes

Chr1Z93:
image


dscarpac:
Let's come up with something better than "drop shitty on snap points"

PR #1085, opened 2025-01-12
Updated player card panel for homunculus

Chr1Z93:


dscarpac:
Hmmm, I'm not sure given that you can't choose it as a starting investigator. I get that it needs to be found somehow, but that also reminds me that The Great Work instructs you flip it, and given that the other side of it in the mod is a player card, we need something to tell people what that really should be (yes, I'm aware it would be strange if someone puts the card in their deck not knowing what it does). Does importing The Great Work give you The Homunculus as a bonded card?

dscarpac (approved):

PR #1101, opened 2025-01-24
Various Improvements to Mythos Area

dscarpac:
(1) Moves the two snap points from the Mythos Area object to Global
(2) Re-aligns snap points for Encounter Deck and Discard pile to match
(3) Clarifies broadcast message for the doom in play addition/subtraction
(4) Moves Agenda position to Global (Constants)


Chr1Z93:
> (1) Moves the two snap points from the Mythos Area object to Global
I don't see how this is an improvement, can you explain?

Chr1Z93:
> (4) Moves Agenda position to Global (Constants)

I think we should turn this into a local position if anything

dscarpac:
> > (1) Moves the two snap points from the Mythos Area object to Global
> 
> I don't see how this is an improvement, can you explain?

I couldn't find an explanation for why only two snap points (the set aside cards box, and the encounter deck) were snap points on the MythosArea object, and all of the other snap points (e.g. discard pile, agenda/act/scenario reference, chaos bag) were snap points in the other global file. This just clears up the inconsistency and codifies the location of these objects globally.

dscarpac:
> > (4) Moves Agenda position to Global (Constants)
> 
> I think we should turn this into a local position if anything

The only reason to do this here was to match (and codify) the snap point positions, for easier lookup for anyone creating new scenarios or trying to "get" the agenda position from an outside script, no different from how the encounter deck and discard position already are, just for consistency. If all three of these should be local positions, that's fine, but that makes it harder to know where the bags should put these objects (e.g. the PR in the other repo)

PR #1127, opened 2025-02-06
Scripted Gift of Nodens

bankey1443:
Adds scripting to Gift of Nodens that will list all of the possible targets (survivor skills in the discard pile) in the context menu by name. Once a context menu item is clicked, that card is moved near Gift of Nodens.

nodens

image


Chr1Z93 (approved):
Looks good to me! Thanks for the contribution

PR #1132, opened 2025-02-08
Added doom spawning as replenish via hotkey

dscarpac:
Allows for neater organization of doom on cards. Press "7" over cards that do not have "uses" in their metadata or have "doom" in their metadata to perform the replenish ability.

Could allow for clickable counter or dynamic token in the future.

Tokens with "ignore doom" tag are left alone.

image


Chr1Z93 (commented):
I think it would be nice if the doom ignore tag was respected

Chr1Z93 (approved):

PR #1139, opened 2025-02-14
Kohaku helper

dvrooman:
### Updates to Kohaku's Helper
- Flips reaction token on Investigator Card when Bless/Curse/Action buttons are pressed
- Adds "if reaction token is face down" to logic for enabling buttons
- Fixes reversed error messages for Bless/Curse already at 10 in bag
- Separates logic for enabling Elder Sign (available if _either_ Bless or Curse is < 10)

### DIFF
--- a/src/playercards/cards/KohakuNarukami.ttslua
+++ b/src/playercards/cards/KohakuNarukami.ttslua
@@ -106,11 +106,13 @@ function maybeUpdateButtonState()
 
   if numInBag.Bless <= numInBag.Curse and numInBag.Bless < 10 then
     state.Bless = true
-    state.ElderSign = true
   end
 
   if numInBag.Curse <= numInBag.Bless and numInBag.Curse < 10 then
     state.Curse = true
+  end
+
+  if numInBag.Bless < 10 or numInBag.Curse < 10 then
     state.ElderSign = true
   end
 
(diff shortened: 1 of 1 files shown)


src/playercards/cards/KohakuNarukami.ttslua
```diff
@@ -134,14 +148,45 @@ function setUiState(params)
...
+  local pos = zones.getZonePosition(matColor, "Investigator")
+  local searchResult = searchLib.atPosition(pos, "isCardOrDeck")
+  if #searchResult == 1 then
+    local reaction = searchLib.onObject(searchResult[1], "isTileOrToken")
+    if #reaction == 1 then
+      if string.sub(reaction[1].getData().Nickname, -8, -1) == "Reaction" then
+        reactionToken = reaction[1]
+      end
+    end
+  end
```
Chr1Z93, line 182:
```suggestion
  local rot = playermatApi.returnRotation(matColor)
  local size = Vector(2.3, 2, 3.4)
  local searchResult = searchLib.inArea(pos, rot, , "isUniversalToken")
  for _, obj in ipairs(searchResult) do
      if string.sub(obj.getName(), -8, -1) == "Reaction" then
        reactionToken = obj
        break
      end
  end
```
Untested, but something like this would be better to get the ability token imo (one search instead of 2)

src/playercards/cards/KohakuNarukami.ttslua
```diff
@@ -134,14 +148,45 @@ function setUiState(params)
 end
 
 function errorMessage(_, _, triggeringButton)
+  local reactionToken = getReactionToken()
+  local actionSpent = false
```
Chr1Z93, line 152: Would use a different variable name here, maybe `isTokenFaceDown`?

dvrooman:
Incorporated suggestions from Chr1Z

Chr1Z93 (approved):
Thanks for the bugfix & additions!

PR #1183, opened 2025-04-04
Added a hotkey to set a virtual color for the next hotkey

Chr1Z93:


dscarpac:
Not sure about this yet -- in practice, if you are getting more than one clue from a location (e.g. Deduction), it requires you to hit the hotkey, press the hotkey for getting a clue, then moving your cursor back to your playmat, pressing the hotkey again, then moving back to the location, then hitting the clue hotkey again.

Chr1Z93:
> Not sure about this yet -- in practice, if you are getting more than one clue from a location (e.g. Deduction), it requires you to hit the hotkey, press the hotkey for getting a clue, then moving your cursor back to your playmat, pressing the hotkey again, then moving back to the location, then hitting the clue hotkey again.

Good point. We add a timer to it as grace period. E.g. after it was used the first time it persists as the virtual color for 0.5s (and the timer resets if used again). That would allow you to grab multiple clues.

Chr1Z93:
Updated it with a 1s grace period
grafik

dscarpac (approved):

PR #1185, opened 2025-04-04
Display an XML count of clue stacks on top

Chr1Z93:
Closes: https://github.com/argonui/SCED/issues/390

grafik

### DIFF
--- a/config.json
+++ b/config.json
@@ -26,7 +26,7 @@
     "HandTrigger.be2f17",
     "HandTrigger.0285cc",
     "HandTrigger.a70eee",
-    "ScriptingTrigger.a2f932",
+    "PlayAreaZone.a2f932",
     "FogOfWarTrigger.3aab97",
     "TableLegBottomRight.afc863",
     "TableLegBottomLeft.c8edca",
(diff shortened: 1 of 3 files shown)


dscarpac:
image

(1) Let's block this for when the stacks are on the clue-counting area of the playermat. I find the extra numbers to be too distracting, and not particularly useful. (If you need to spend a limited number of clues, hovering over tells you how big the stack is anyways)
(2) The XML label should be centered on the top token in the stack, not on the stack as a whole, since you sometimes will get the number printed way too close to the edge of the token on top.

Chr1Z93:
> (1) Let's block this for when the stacks are on the clue-counting area of the playermat. I find the extra numbers to be too distracting, and not particularly useful. (If you need to spend a limited number of clues, hovering over tells you how big the stack is anyways)
Good point, updated it to limit it to the playarea

> (2) The XML label should be centered on the top token in the stack, not on the stack as a whole, since you sometimes will get the number printed way too close to the edge of the token on top.
I noticed that too but I don't think it's possible. Tokens combine to a single "stack" object and I don't see a way to get the dimensions of just part of it.

dscarpac (approved):

PR #1187, opened 2025-04-05
Update BookofLivingMyths.xml

dvrooman:
tablelayout z height causing issues with clickability of buttons

### DIFF
--- a/xml/playercards/BookofLivingMyths.xml
+++ b/xml/playercards/BookofLivingMyths.xml
@@ -13,7 +13,7 @@
     onClick="resolveToken"
     textColor="white"
     active="false"/>
-  <TableLayout position="0 -55 -22"
+  <TableLayout position="0 -55 -40"
     rotation="0 0 180"
     height="900"
     width="1400"


dvrooman:
I wonder if this is an issue with all cards that have xml ui on top of the card? I'm currently playing Kōhaku, so I noticed it on him.

I haven't done any testing yet to see if TTS settings/resolution make a difference. I've noticed issues on xml buttons that are too close to geometry on other projects though.

Chr1Z93:
Could you provide a screenshot of the issue that this is supposed to fix? I am not aware of it

dvrooman:
https://github.com/user-attachments/assets/9a9f0eba-f206-4edd-b4b8-ece7eb003c2d

The button flashes in and out of existence. You need to be in very specific areas for the button to be clickable.

I don't remember this happening in previous builds, and looking through the commit history, I don't think this value has changed. It might be a setting in TTS I changed on my machine that is causing the issue.

As I said earlier, I've witnessed similar behavior in other projects when I try to get XML buttons too close to geometry. The specific value of -40 I chose was trial and error: I was too lazy to do the scaling math. -40 seems to result in consistent button functionality without having the button appear to float above the card.

Chr1Z93:
Wow that is weird indeed and I can not reproduce it 😅
I wonder what causes this. Probably effects more code then.

dvrooman:
I think I figured it out! Seems to occur when I'm using a non-native resolution. I have a 27" 5k monitor, and TTS is notoriously unoptimized for macOS. To avoid hiccups and my fans constantly spinning up, I was running the game at a lower resolution. Once I switched to the monitor's native resolution, the issue went away.

In all likelihood, this is a macOS specific bug, but I have no way of testing that. Either way, I don't believe my change will affect functionality other than improving it for my particular edge case.

dvrooman:
> Probably effects more code then.

Most likely. Now that I’m thinking of it, one of the players in my group has issues clicking buttons a lot. I know he’s running an older MacBook Pro, so probably has his resolution lowered…

Chr1Z93 (approved):

PR #1188, opened 2025-04-06
Updated Search-A-Card

Chr1Z93:
- Added a "no limit" mode that allows you to spawn all matches (usually limited to 100 for performance reasons)
- Removed search string length limitation
- Slightly updated the code

### DIFF
--- a/src/playercards/Search-A-Card.ttslua
+++ b/src/playercards/Search-A-Card.ttslua
@@ -2,20 +2,23 @@ require("playercards/PlayerCardSpawner")
 
 local allCardsBagApi           = require("playercards/AllCardsBagApi")
 
-local BUTTON_LABELS            = {}
-BUTTON_LABELS["spawn"]         = {}
-BUTTON_LABELS["spawn"][true]   = "All matching cards"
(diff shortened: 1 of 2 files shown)


src/playercards/Search-A-Card.ttslua
```diff
@@ -45,24 +54,38 @@ function onLoad(savedData)
...
   buttonParameters.height         = 600
   buttonParameters.width          = 2800
   self.createButton(buttonParameters)
+
+  createContextMenu()
+end
+
+function createContextMenu()
+  self.clearContextMenu()
+  self.addContextMenuItem((noLimit and "Dis" or "En") .. "able NoLimit", toggleNoLimit)
```
dscarpac, line 81: did an extra space in between "No" and "Limit" cause a line break?

dscarpac (changes requested):

src/playercards/Search-A-Card.ttslua
Chr1Z93: It was really close to a line break, but we can fit the space - updated

dscarpac (approved):

PR #1199, opened 2025-04-11
Bugfixes for Layouts

Chr1Z93:
- update the position of playermat objects if they exist already
- fixed names in index so bags don't get spawned twice


dscarpac:
image

My machine might not be fast enough, but when going from Classic to Sides (4p), the action tokens remain in situ.

Chr1Z93:
> My machine might not be fast enough, but when going from Classic to Sides (4p), the action tokens remain in situ.

Thanks, that was a specific issue when two mats temporarily overlapped! Should be fixed now.

dscarpac (approved):

PR #1205, opened 2025-04-17
Increased area for "Doom in play" calculation

Chr1Z93:
Needed change to support rotated playermats

### DIFF
--- a/src/mythos/DoomInPlayCounter.ttslua
+++ b/src/mythos/DoomInPlayCounter.ttslua
@@ -4,8 +4,8 @@ local playermatApi     = require("playermat/PlayermatApi")
 local ZONE, TRASH
 local IGNORE_TAG       = "DoomCounter_ignore"
 local TOTAL_PLAY_AREA  = {
-  upperLeft = { x = -9, z = -35 },
-  lowerRight = { x = -60, z = 35 }
+  upperLeft = { x = -9, z = -50 },
+  lowerRight = { x = -60, z = 50 }
 }
 local subtractDoom     = false
 


dscarpac:
Is there another way to do this? I'd rather that this only look on playermats (like what we do with clues?) in addition to the main playmat area. As is, if you don't have the 4-player side setup, it will still count doom in play if a card is off in a corner somewhere with doom on it.

Chr1Z93:
> Is there another way to do this? I'd rather that this only look on playermats (like what we do with clues?) in addition to the main playmat area. As is, if you don't have the 4-player side setup, it will still count doom in play if a card is off in a corner somewhere with doom on it. Another potential solution is to change the search area if the other player mat arrangement is selected.

That's a good point. Only searching on playermats would be turning this search into up to 6x Physics.cast() though:
- 4x Playermats
- 1x Play Area
- 1x Mythos Area

While probably not too much of an issue performance wise, it gets a bit iffy because we are already searching the playermats continuously for clues.
So in that case we probably want to change our approach and combine these searches.
Long story short: Possible, but that is a larger change than this 2-line-fix, so should be in a separate PR and isn't super high priority to me personally.

dscarpac:
I guess maybe it's not urgent, but I'm thinking of cases where you put doom on an enemy and it's temporarily out of play, like in Depths of Yoth or Twisted Hollow. The doom shouldn't be counted, but there's no good place for the cards to go under this new calculation.

Chr1Z93:
> I guess maybe it's not urgent, but I'm thinking of cases where you put doom on an enemy and it's temporarily out of play, like in Depths of Yoth or Twisted Hollow. The doom shouldn't be counted, but there's no good place for the cards to go under this new calculation.

Usually, cards lose all tokens when they leave play. Is there some special wording for that? I don't exactly remember it.

dscarpac:
> > I guess maybe it's not urgent, but I'm thinking of cases where you put doom on an enemy and it's temporarily out of play, like in Depths of Yoth or Twisted Hollow. The doom shouldn't be counted, but there's no good place for the cards to go under this new calculation.
> 
> Usually, cards lose all tokens when they leave play. Is there some special wording for that? I don't exactly remember it.

Yes, these scenarios have campaign guide instructions, e.g.:
image

Chr1Z93:
> > > I guess maybe it's not urgent, but I'm thinking of cases where you put doom on an enemy and it's temporarily out of play, like in Depths of Yoth or Twisted Hollow. The doom shouldn't be counted, but there's no good place for the cards to go under this new calculation.
> > 
> > Usually, cards lose all tokens when they leave play. Is there some special wording for that? I don't exactly remember it.
> 
> Yes, these scenarios have campaign guide instructions, e.g.:
> image
> 

Okay then we might want to look into giving these zones some special handling (maybe using the special cards from Antimarkovnikov?).
Either way, I don't think that should be a blocker for this PR since that simply fixes the current issue.

dscarpac (approved):

PR #1220, opened 2025-04-28
Updated tour

Chr1Z93:
- updated speech bubble (higher res)
- increased scale (about 40% and additional 20% for text)
- adjusted script of some tour cards
- added new tour card for hotkeys

grafik


Chr1Z93:
> Do we need the section about "going back to third person mode" anymore since the tour does that at the end automatically?

That's fair. Updated Darrell with more useful information:

grafik

dscarpac (approved):

PR #1226, opened 2025-04-30
Added table setup support to option panel state loading

Chr1Z93:


src/core/Global.ttslua
```diff
@@ -2949,7 +2953,6 @@ end
 function loadOptionPanelState(newOptions)
   for id, state in pairs(newOptions) do
     if optionPanel[id] ~= state then
-      optionPanel[id] = state
```
Chr1Z93, line 2952: This was removed because it is already part of the `applyOptionPanelChange()` function

src/core/Constants.ttslua
Chr1Z93, line 1: There is just an alphabetic sorting of the presets and alignment of equal signs in this file

src/core/Global.ttslua
```diff
@@ -2865,6 +2865,10 @@ function applyOptionPanelChange(id, state)
     -- option: Show displacement tool
   elseif id == "showDisplacementTool" then
     spawnOrRemoveHelper(state, "Displacement Tool", { -68, 1.53, 57 })
+
+    -- this will only be called by the "loadOptionPanelState" function
+  elseif id == "tableSetup" then
+    updateTableSetup(state, nil)
```
Chr1Z93, line 2871: This is the only "real" change

dscarpac (approved):

PR #1261, opened 2025-05-21
New Chaos Token images

Chr1Z93:
Intentionally no screenshot here, please check them out ingame (updated in all places via search & replace)


Chr1Z93:
Re-opening with some edits

PR #1267, opened 2025-05-22
Updates trackers based on token cancellation

dscarpac:
Flipping a chaos token removes it from the tracks (both on automatic counters and the chaos bag stat tracker


Chr1Z93 (changes requested):
This will untrack the token each time it's flipped. It's also not accounting for chaos tokens that are used as keys.

Chr1Z93 (approved):
lgtm

PR #1279, opened 2025-06-01
Key of Solomon helper

dvrooman:
Using a Bless/Curse exhausts the card. I couldn't find any API to do this, so I did it manually. Sorry if I missed something already built into SCED.

It would be nice to use the Player's Rotation Degrees setting, but I couldn't find a way to access that…

### DIFF
--- a/src/playercards/cards/KeyofSolomon.ttslua
+++ b/src/playercards/cards/KeyofSolomon.ttslua
@@ -1,0 +1,109 @@
+require("playercards/CardsWithHelper")
+local blessCurseManagerApi     = require("chaosbag/BlessCurseManagerApi")
+local playermatApi             = require("playermat/PlayermatApi")
+
+-- intentionally global
+hasXML                         = true
+isHelperEnabled                = false
+local updated, loopId
+
+function updateSave()
+  self.script_state = JSON.encode({
+    isHelperEnabled = isHelperEnabled,
+    loopId = loopId
+  })
+end
+
+function onLoad(savedData)
+  if savedData and savedData ~= "" then
+    local loadedData = JSON.decode(savedData)
+    isHelperEnabled = loadedData.isHelperEnabled
+    loopId = loadedData.loopId
+  end
+  if isHelperEnabled then updateDisplay() end
+end
+
+function shutOff()
+  if loopId then
+    Wait.stop(loopId)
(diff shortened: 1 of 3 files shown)


src/playercards/cards/KeyofSolomon.ttslua
```diff
@@ -0,0 +1,109 @@
...
+    local investigatorName = playermatApi.getInvestigatorName(matColor)
+    broadcastToAll("Heal up to 2 damage/horror from an Investigator or Ally at " .. investigatorName .. "'s location.", "Yellow")
+  end
+  
+  updated = false
+  Wait.frames(maybeUpdateButtonState, 2)
+  
+  -- exhaust card
+  local rot = self.getRotation()
+  rot.y = rot.y + 45 -- it would be nice to get the player's preferred Rotation Degrees here
```
Chr1Z93, line 64: Unfortunately, there is no way to get the player's local setting.
That said, I would like to default this to "playermat rotation + 90°".

Chr1Z93 (changes requested):

src/playercards/cards/KeyofSolomon.ttslua
dvrooman: Should be updated. My brain is officially pudding for the evening, so hopefully the new stuff added inspired by your request isn't riddled with bugs ;)

src/playercards/cards/KeyofSolomon.ttslua
```diff
@@ -0,0 +1,135 @@
...
+  else
+    matColor = playermatApi.getMatColor(player.color)
+  end
+
+  blessCurseManagerApi.removeToken(tokenType, player.color)
+
+  if tokenType == "Curse" then
+    local actionResources = 2
+    playermatApi.updateCounter(matColor, "ResourceCounter", nil, actionResources)
+    printToColor("Added " .. actionResources .. " resources to " .. matColor .. "'s resource pool.", player.color)
```
Chr1Z93, line 53: I think this should use `GlobalApi.getColoredName(playerColor)`.
And also probably print it for both the triggering player and the effected player (if different).

Chr1Z93 (approved):
Merging the current state and finalizing it myself with Dave's permission

PR #1281, opened 2025-06-02
Gabriel Carillo Helper

dvrooman:


src/playercards/cards/GabrielCarillo1.ttslua
```diff
@@ -0,0 +1,79 @@
...
+
+  local matColor
+  if player.color == "Black" then
+    matColor = playermatApi.getMatColorByPosition(self.getPosition())
+  else
+    matColor = playermatApi.getMatColor(player.color)
+  end
+
+  blessCurseManagerApi.addToken("Curse", player.color)
+  tokenAdded = true
```
Chr1Z93, line 49: Considering that there are ways to ready cards outside of upkeep, I would not track whether the ability was used already

Chr1Z93 (approved):
Merging the current state and finalizing it myself with Dave's permission

PR #1319, opened 2025-06-26
Adds a hotkey to rotate the alt preview by 90°

Chr1Z93:
Can be useful to fix falsely imported cards or other things.


Chr1Z93:
Less useful than I thought - editing in the JSON is much easier

PR #1323, opened 2025-06-28
add width parameter to summoned servitor

Shestel12:
Add the option to change width of summoned servitor buttons - should help with localization, since those buttons are of different size in each language
image


Chr1Z93 (commented):
Looks good, but what is the background for the formula "x * 1000 + 5"?

Shestel12:
> Looks good, but what is the background for the formula "x * 1000 + 5"?

* 1000 is to translate from the type of value that vector expects (0.05) to the one that is used in button width (50). + 5 is to add a bit more to account for the border size (it always has the same thickness).

PR #1324, opened 2025-06-28
Update ru.po

lordhx:
- Fixed an issues with newline
- Updated a few existing translations to fit UI or sound more native
- Added a couple of missing translations


locales/ru.po
```diff
@@ -247,7 +253,7 @@ msgstr "По сторонам (4И)"
 
 # Dropdown Option: Table Setup 5
 msgid "OP_tableSetup_option_solo"
-msgstr "Настоящее соло (1И)"
+msgstr "True solo (1И)"
```
Chr1Z93, line 256: Is there no proper translation for this?

PR #1365, opened 2025-07-16
feat: init ko translations (UI)

derornos:


Chr1Z93 (approved):
Thanks for the contribution!

PR #1379, opened 2025-07-23
Updated Token splash

Chr1Z93:
Implemented custom fade function to allow seamless drawing of multiple tokens.

Potential future improvement: When stopping the ongoing animation, continue with the current color progress instead of restarting the fade animation.


dscarpac:
One more thing I've noticed -- seems the fade out is very slow (or delayed?) if you return the token (by clicking again) before the initial fade *in* has completed

PR #1382, opened 2025-07-24
Add personal last five chaos token tracker and helper function.

cchenchen:
Adds a tracker for each mat color for "last five" chaos token tracker.

Also moves the "last five" chaos token output from the main handler to a helper function `printLastFiveChaosTokens(playerColor)` that can be called in multiple places.

For consideration - while creating the edit I noticed inconsistency between use of `matColor` and `playerColor`. Are they always identical? This may have impact on tables keyed off of one or the other. Suggest refactoring to use one or the other for consistency.

### DIFF
--- a/src/Global/Global.ttslua
+++ b/src/Global/Global.ttslua
@@ -1097,6 +1097,9 @@ function trackChaosToken(tokenName, matColor, subtract)
     -- initialize table
     tokenDrawingStats[key]               = tokenDrawingStats[key] or {}
 
+    -- initialize per-player table for last drawn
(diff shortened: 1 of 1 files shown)


cchenchen:
That makes sense, that the tracker uses `matColor` as that is where the chaos tokens spawn, while the button press uses `playerColor` as that's who is pressing the button. Does lua throw errors if `#<setname>` is empty or if there is no key for `<setname>[<index>]`? If so I'll put a sanity check for the niche case where a player starts on one board and then swaps colors before clicking the button.

PR #1499, opened 2025-10-13
Adds option to revert back to standard resource images

dscarpac:
Adds option to both playermat and global option panels to use custom images for the resource tokens (the status quo, turned on by default).
When off, the "state" is still maintained to show if the token "actually" represents ammo, charges, etc.


src/Global/Global.ttslua
```diff
@@ -3413,6 +3413,9 @@ function applyOptionPanelChange(id, state)
     Wait.frames(function() spawnOrRemoveHelperForPlayermats("Clickable Clue Counter", state) end, state and 0 or 20)
     Wait.frames(function() PlayermatApi.clickableClues(state, "All") end, 10)
 
+    -- option: Custom Resource Images
+  elseif id == "useCustomResourceImages" then
+
```
Chr1Z93: This doesn't do anything, should be removed.

src/Global/Global.ttslua
```diff
@@ -4031,14 +4050,18 @@ function TokenManager.spawnMultipleTokens(params)
...
     tokenType = string.lower(subType)
+  else
+    callbackName = "addUseName"
+    callbackParams = { useName = subType }
   end
 
-  -- special handling for some tokens like action/ability tokens
-  local callbackName = nil
-  local callbackParams = nil
   if tokenType == "universalActionAbility" then
```
Chr1Z93:
```suggestion
  end

  if tokenType == "resource" then
      callbackName = "addUseName"
      callbackParams = { useName = subType }
  elseif tokenType == "universalActionAbility" then
```

I think this is the cleanest solution to naming regular resources too

PR #1508, opened 2025-10-20
Phase Tracker verbosity mode

MartinMoFe:
The phase tracker has a second level of verbosity that displays in the chat how many resources and cards the characters have at the begining of the inv phase. It also reminds you in the chat if there are enemies in the play area and in the threat area of playermats. There is a third level of verbosity that does all that and, additionally, moves the camera and shows clearly the position of the investigator minicards and highlights enemies in the enemy phase. I've been using these phase tracker in my games and it helps me follow the game, since the chat messages integrate well with messages like "removed charge" or "spent clues".

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
-.vscode/settings.json
 .mo
+.get-lua*
+.LUAcoding*.txt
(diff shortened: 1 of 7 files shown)


Chr1Z93 (changes requested):
Left a few comments and then noticed that this has a lot of the exact same things I already commented on in your other PR. Please take a look at my comments over there.

PR #1513, opened 2025-10-25
Updated Tarot Card Deck functions

dscarpac:
* Additional scripting for Observed / Damned
* Adds scripting to Observed / Damned to spawn scripted tarot card deck
* Allows for positioning of scripted tarot card readings to match orientation of the scripted bag
* Adds ability to double tarot card size for increased visibility
* Uses existing card rotation / alt-preview rotation functions to rotate cards for readings

### DIFF
--- a/src/playercards/ScriptedTarot.ttslua
+++ b/src/playercards/ScriptedTarot.ttslua
@@ -28,7 +28,12 @@ function chaos(color)
   self.takeObject({
     position = self.positionToWorld(BAG_OFFSET + CARD_OFFSET),
     smooth = true,
-    callback_function = function(card) if math.random(2) == 2 then card.call("rotateSelfAndPreview", color) else card.setRotation(self.getRotation()) end end
+    callback_function = function(card)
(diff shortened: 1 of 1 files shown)


src/playercards/ScriptedTarot.ttslua
```diff
@@ -10,30 +10,34 @@ READING = {
...
-function chaos()
+function chaos(color)
   self.shuffle()
   self.takeObject({
-    position = self.getPosition() + CARD_OFFSET,
-    rotation = ORIENTATIONS[math.random(2)],
-    smooth = true
+    position = self.positionToWorld(BAG_OFFSET + CARD_OFFSET),
+    smooth = true,
+    callback_function = function(card) if math.random(2) == 2 then card.call("rotateSelfAndPreview", color) else card.setRotation(self.getRotation()) end end
```
Chr1Z93: Looks like this is not getting the original bag rotation when it's rotated on its head.

PR #1537, opened 2025-11-02
Respect customized Mythos Area transform

Chr1Z93:


dscarpac:
I'm finding that if I change values a second time in the panel, and hit confirm, nothing happens.

dscarpac (approved):

PR #1543, opened 2025-11-05
PhaseTracker Upgrades

MartinMoFe:
Feature 1. Toggleable verbosity mode of the PhaseTracker.
In this feature, The phase tracker can be toggled on to display more information, like
·the resources and hand cards that the players have, to keep a more detailed record, at the Investigator Phase
· The number of enemies in the Playarea and in the Threat area in the Enemy Phase.
· Highlighting for 10 seconds the enemies
·More colourful messages
·It enables future messaging, like displaying forced effects that trigger at the end of any given phase

Feature 2. Toggleable verbosity mode *and* camera moving of the PhaseTracker
This feature does all the previous things *and*
·moves the camera to the mythos area when the Mythos Phase begins
·moves the camera to the investigator miniplayer card when the Investigator Phase begins
·moves the camera to the enemy in the playarea if there is a single one (never forget that enemy in the corner)

Changes: this change adds variables to the PhaseTracker, so it changes the json file. It adds two functions to the playermat, that return the elements in the Threat area.

Notes:
·The feature 1 and 2 could be untangled in future versions.
· The functions that detect enemies and things in general in the playarea could be global if used by other features in the future

### DIFF
--- a/src/playarea/PhaseTracker.ttslua
+++ b/src/playarea/PhaseTracker.ttslua
@@ -1,5 +1,7 @@
 local PlayermatApi = require("playermat/PlayermatApi")
 local MythosAreaApi = require("mythos/MythosAreaApi")
+local zones         = require("playermat/Zones")
+local SearchLib     = require("util/SearchLib")
 
 local PHASE_NAMES = {
   "I. Mythos Phase",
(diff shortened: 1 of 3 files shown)


Chr1Z93 (changes requested):
Let me preface this by saying that your code is fully functional and that is the most important thing.
I have to request a few changes though to make sure it fits in the existing ecosystem as smoothly as possible.

1) Utilize native TTS features when possible
Example: You have a helper function for getting player references instead of doing `Player[playerColor]`.

2) Unneeded log() statements
You seem to have some unused log() statements from development left in the code.
Example: `log(matColor)`

3) Threat Area position getting
We already have a file that maps all the local positions for the playermat -> playermat/Zones.ttslua
Please use that instead of expanding the PlayermatApi (I'll gladly help you with it).

MartinMoFe:
Notes taken. Developed cleaner in other branch.

PR #1582, opened 2025-12-11
Adds two "Advance" buttons // Clue Spending UI // Updated i18n

Chr1Z93:
This includes https://github.com/argonui/SCED/pull/1595 and https://github.com/argonui/SCED/pull/1613.

grafik

Main features:
- advance buttons for mythos area
- clue spending UI
- playermat localization

### DIFF
--- a/locales/de.po
+++ b/locales/de.po
@@ -658,28 +658,28 @@ msgid "PM_option_useResourceCounters_dropdown_4"
 msgstr "Deaktiviert"
 
 # Playermat: Buttons
-msgid "PM_discardMove_button"
+msgid "PMB_discardMove_button"
 msgstr "Ablegen / ➜"
 
-msgid "PM_discardMove_button+tooltip"
+msgid "PMB_discardMove_button+tooltip"
 msgstr "Rechtsklick um in die Bedrohungszone zu verschieben"
 
-msgid "PM_discard_button"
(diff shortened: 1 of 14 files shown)


Chr1Z93:
> (1) Should not having enough clues (bringing up the warning message) prevent the act card from flipping? Could have another part of the dialogue box here that allows for a skip, or just bring up the clue UI anyways and force users to cancel act advancement or skip.

This is an interesting question. I am not sure if we should gatekeep the advance button like that - if someone wants to advance it with unsufficient clues, let them do it.

> (2) Apologies -- might need a lighter touch on doom removal for agendas that tell you to move doom from this agenda onto the next, or leaving doom on enemies / locations in play unless we want to incorporate that into the metadata.

This is fully handled via the settings menu on the doom counter.

dscarpac:
> > (2) Apologies -- might need a lighter touch on doom removal for agendas that tell you to move doom from this agenda onto the next, or leaving doom on enemies / locations in play unless we want to incorporate that into the metadata.
> 
> This is fully handled via the settings menu on the doom counter.

I think this is only true if the user is anticipating what the advance button does and does not do and sets the menu properly. I guess it is possible to reconstruct the board state with proper messages but this would be a good thing to ask people to test.

PR #1593, opened 2026-01-02
Adds Elder Sign metadata to investigators and token splash  

dscarpac:
Additionally adds metadata for health and sanity for a future feature, and adjusts position of token splash text.

### DIFF
--- a/src/Global/Global.ttslua
+++ b/src/Global/Global.ttslua
@@ -1242,6 +1242,7 @@ function tokenSplash(matColor)
           else
             modifier = TokenArrangerApi.getCurrentModifier(tokenName) or TokenArrangerApi.getCurrentModifier("")
           end
+
           if modifier and math.abs(modifier) < 100 then
             local modifierText = "( "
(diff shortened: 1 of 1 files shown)


objects/AllPlayerCards.15bb07/AgnesBaker.25e2db/AgnesBaker.6797bb.gmnotes
```diff
@@ -10,6 +10,9 @@
   "intellectIcons": 2,
   "combatIcons": 2,
   "agilityIcons": 3,
+  "elderSignEffect": {
+    "description": "+1 for each horror on Agnes Baker."
+  },
```
Chr1Z93: a) Why has the health / sanity sneaked into this PR? 🤣 

b) Why is it not added for all investigators?

Chr1Z93 (commented):
I think we should keep the name for the metadata field instead of renaming it for the investigatorData

PR #1596, opened 2026-01-05
Attempt to fix HideWhenFaceDown in localizations

Shestel12:
Right now HideWhenFaceDown property seems to be ignored on cards that are replaced by the AllEncounterCardsBag - it always defaults to false. This PR attempts to fix it.

### DIFF
--- a/src/mythos/AllEncounterCardsBag.ttslua
+++ b/src/mythos/AllEncounterCardsBag.ttslua
@@ -385,14 +385,8 @@ function updateOriginalItemData(originalData, replaceData)
   end
 
   originalData["Tags"] = originalData["Tags"] or replaceData["Tags"] or {}
-  -- Try to set HideWhenFaceDown with following priority, depending on whether it is set or not:
-  -- Original -> Replacement -> true
   if originalData["HideWhenFaceDown"] == nil then
-    if replaceData["HideWhenFaceDown"] ~= nil then
-      originalData["HideWhenFaceDown"] = replaceData["HideWhenFaceDown"]
-    else
-      originalData["HideWhenFaceDown"] = true
-    end
+    originalData["HideWhenFaceDown"] = true
   end
   table.insert(originalData["Tags"], "Replaced")
 


src/mythos/AllEncounterCardsBag.ttslua
```diff
@@ -385,6 +385,15 @@ function updateOriginalItemData(originalData, replaceData)
   end
 
   originalData["Tags"] = originalData["Tags"] or replaceData["Tags"] or {}
+  -- Try to set HideWhenFaceDown with following priority, depending on whether it is set or not:
+  -- Original -> Replacement -> true
+  if originalData["HideWhenFaceDown"] == nil then
+    if replaceData["HideWhenFaceDown"] ~= nil then
+      originalData["HideWhenFaceDown"] = replaceData["HideWhenFaceDown"]
```
Chr1Z93: I don't think `HideWhenFaceDown` should be a field that can differ for original and localized files.

Can you try if this achieves the desired result too?

```lua
if originalData["HideWhenFaceDown"] == nil then
  originalData["HideWhenFaceDown"] = true
end
```

Shestel12: Yep, that works, thanks!

Shestel12: In almost all of the localization there are files where the HideWhenFaceDown property is present and differs between original and localized files...But these are most likely just bugs, due to how the files were copied/generated/etc., like you mentioned original should take priority.
Hidden Chamber, for example:
[original, correctly hidden](https://github.com/Chr1Z93/SCED-downloads/blob/e90110983528b7ee3557afab27de74af83e15daf/decomposed/campaign/The%20Dunwich%20Legacy/TheDunwichLegacy.2898f6/4BloodontheAltar.30684d/Set-aside.7a167a/TheHiddenChamber.cb0272.json#L18)
[translation, HideWhenFaceDown:false](https://github.com/Chr1Z93/SCED-downloads/blob/4179a8093485d110ec68ae4b6981a95e3e03bdc7/decomposed/language-pack/German%20-%20Campaigns/German-Campaigns.GermanC/DasVerm%C3%A4chtnisvonDunwich.3b7aa9/DieverborgeneKammer.1f5bbe.json#L17)

PR #1713, opened 2026-04-13
Adds investigator tokens from 2026 Core

dscarpac:
Right click function adds border highlighting in case two investigators are of the same class.

image

image


dscarpac:
Might need some color correction in the tokens (Mystic especially)

Chr1Z93 (approved):

PR #1724, opened 2026-05-04
Enhancements to Mythos Area

dscarpac:
(1) Added option to set Hard/Expert as the default side of the scenario reference card

(2) Encounter deck is shuffled automatically on new scenario placement.


locales/de.po
```diff
@@ -32,6 +32,14 @@ msgstr ""
 "Legt die Sprache für UI-Elemente,\n"
 "wie dieses Optionsmenü fest."
 
+# Option: set default difficulty to Hard/Expert
+msgid "OP_defaultHardExpert"
+msgstr "Set difficulty to Hard/Expert"
+
+# Tooltip: set default difficulty to Hard/Expert
+msgid "OP_defaultHardExpert_tooltip"
+msgstr ""Set the scenario reference card\nto Hard/Expert by default."
```
Chr1Z93: There's a double `"` at the start (should just be one) - same thing in the other languages

src/mythos/MythosArea.ttslua
```diff
@@ -291,6 +291,29 @@ function fireScenarioChangedEvent(md)
     end
   end
 
+  -- shuffle encounter deck if there is no player in the Black seat ("debug mode")
+
```
Chr1Z93:
```suggestion
local blackSeatUsed = false
```

src/mythos/MythosArea.ttslua
```diff
@@ -291,6 +291,29 @@ function fireScenarioChangedEvent(md)
...
+    end
+  end
+
+  -- flip scenario card to Hard/Expert
+  if GlobalApi.getOptionPanelState()['defaultHardExpert'] then
+    local rot = scenarioCard.getRotation()
+    scenarioCard.setRotation(rot:setAt("z", 180))
+    tokenData = md["tokens"]["back"]
+    fireTokenDataChangedEvent()
+  end
```
Chr1Z93: I think this should be in this place:
https://github.com/Chr1Z93/SCED/blob/90e0f711edbdbb4c70c543f4ca6b82a2c120aa59/src/mythos/MythosArea.ttslua#L218

Then we don't need to repeat the `tokenData` code and also don't fire the event twice.

Chr1Z93 (approved):

PR #1725, opened 2026-05-21
Add Chosen of Zburamoarte

Chr1Z93:
This adds the new reward card from Children of Blood and Blood tokens (including the necessary code changes to make them function).


dscarpac:
I guess this is all right even if no one can technically play with it yet. Will also need to add some relevant code to Nkosi. Should we add some modifiers to these to make something appear on the token splash, like for Bless/Curse? Later on we'll need to add functionality for sealing on investigator cards (or some equivalent like moon tokens).

Chr1Z93:
You're right, this is just the first stepping stone in preparation for the full "blood token support".
I think I already added a modifier, but I'll double-check that.

Chr1Z93:
Thanks for updating it!

PR #1728, opened 2026-05-29
Add a reminder card for "Starting"

Chr1Z93:
This will automatically be added to your deck if you have a "starting" card. It has a button to draw all starting cards.

grafik

### DIFF
--- a/objects/AllPlayerCards.15bb07.json
+++ b/objects/AllPlayerCards.15bb07.json
@@ -5,6 +5,7 @@
     "r": 0.70588
   },
   "ContainedObjects_order": [
+    "StartingReminder.starting",
     "ZoeysCross.66d811",
     "ZoeysCross.66d810",
     "ZoeySamarasParallelFront.98a0e3",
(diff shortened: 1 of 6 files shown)


dscarpac:
*Should use Teutonic font to match other helpers
*Should we eliminate this function from the Search Helper now?
*Add some kind of broadcast to the player to remind them to only choose one starting card to keep in their hand (if there are multiple). Or, have it use the extra search helper hand to make it clear what is going on.

Chr1Z93:
- Font: I'll fix that
- Search Assistant: I think we can keep it in its context menu for now since it doesn't do any harm and is accessible for in-mod deckbuilders
- Broadcast: There's a print message already, would you prefer to turn that into a broadcast?

dscarpac:
> * Font: I'll fix that
> * Search Assistant: I think we can keep it in its context menu for now since it doesn't do any harm and is accessible for in-mod deckbuilders
> * Broadcast: There's a print message already, would you prefer to turn that into a broadcast?

Ah, you're right, I must have missed the Print message (I usually expect broadcasts when some action is required)

Chr1Z93:
Thanks for updating it!

PR #1736, opened 2026-06-09
Update Constants.ttslua

a-constantin:
highly requested update


Chr1Z93:
Hi there! 
Usually, these presets are reserved for regulars, contributors or content creators. 
I'll let this one in since you figured it out on your own :)
Hopefully this means that there are more contributions coming from you ;)

Chr1Z93 (approved):

PR #1742, opened 2026-06-21
Remove Dynamic Taboo List Loading

Chr1Z93:
This change improves the loading time of the mod by a lot and ensures it properly works without web-connection.


Chr1Z93:
Merging this immediately for the performance boost

PR #1770, opened 2026-08-05
Added UI phase tracker

kigrup:
Added the phase tracker as a toggleable and draggable window. The click handler runs the PhaseTracker object changeState method so everything stays in sync.

image

This first version does not load the current phase correctly when loading a game, it always starts at the first phase and is synced after the first change.


src/Global/Global.ttslua
```diff
@@ -13,6 +13,7 @@ local MessageQueue                = require("util/MessageQueue")
 local MathLib                     = require("util/MathLib")
 local MythosAreaApi               = require("mythos/MythosAreaApi")
 local NavigationOverlayApi        = require("core/NavigationOverlayApi")
+local PhaseTracker                = require("playarea/PhaseTracker")
```
Chr1Z93: I don't think this is doing anything at the moment

modsettings/CustomUIAssets.json
kigrup, line 1: I "cloned" the phase images and uploaded them as UIAssets because the original images were PNGs thus showing an ugly white background when used as an Image element. These new images have a gray background matching the color of the UI Panel.

src/playarea/PhaseTracker.ttslua
```diff
@@ -115,4 +122,7 @@ function changeState(_, _, isRightClick)
     tracker.destruct()
     spawnObjectData({ data = data })
   end
+
+  -- update UI phase tracker image
+  UI.setAttribute("phaseTrackerUIImage", "image", PHASE_UI_IMAGES_NAMES[phaseId])
```
kigrup, line 127: @Chr1Z93 tried running this line in the onLoad method but it did not seem to work

src/Global/Global.ttslua
kigrup: At first I thought the PhaseTracker was static so I tried calling the changeState method through this and forgot to remove it.

src/playarea/PhaseTracker.ttslua
Chr1Z93: Yeah, the UI is being created onLoad() - you could ensure it gets initialized with the correct image though.
Right here: https://github.com/Chr1Z93/SCED/pull/1770/changes#diff-33ede2ce868c33d230e42f64ff1f63217e51f1bd25497c01ed079fc85e335032R2804

modsettings/CustomUIAssets.json
Chr1Z93: That's fine

src/playarea/PhaseTracker.ttslua
kigrup: I tried replacing that with something like
`local phaseTrackerObject = getObjectFromGUID("d0c8fa")`
...
`image = phaseTrackerObject.PHASE_UI_IMAGES_NAMES[phaseTrackerObject.phaseId]`
but an error says the variable "PHASE_UI_IMAGES_NAMES" is inaccessible even when I remove its "local" keyword

kigrup: Nevermind, instead of accessing the variables I made a getter method on the object and calling that works

Chr1Z93: Yeah, you need to create a function and call it or use `getVar()` / `getTable()`

kigrup:
The phase name on the UI looks a bit hard to read to me.

- Should I add a tooltip with the phase name (which would allow us to translate it)?
    - Did this but was not happy with how TTS updates the tooltip when clicking it. It sometimes shows the previous tooltip instead of the correct one.
- Should I add an an option in the Options menu to change the size of the Phase Window and the button/image?
   - For now I just made the button a bit bigger so the text is readable.

Chr1Z93 (approved):
Great job, thank you

PR #1771, opened 2026-08-06
Made ui action tracker clickable

kigrup:
The UI Action tracker window can now be used to flip the action tokens of the players. Left click consumes the action and right click restores it. The button takes effect when the token related to that button is "at rest" to fix a small bug.


Chr1Z93 (approved):
Thank you for the contribution!

PR #1799, opened 2026-09-15
GenericCounter with XML instead of createButton

a-constantin:
Hi,

I like using zoom_always in TTS and clearly I'm the only one, but it messes up with the counters, right clicking a counter just opens the context menu instead of decrementing the counter, I have to first left click and then I can right click. But with XML buttons it works just fine, and since that's what damage and horror tokens use, I made the genericcounter use code from DamageHorrorCounter.ttslua, which has the added benefit of the tokens and the counters now having the same font.

Hopefully it doesn't break anything, thank you for your time.


Chr1Z93:
Hi there!
Would you mind elaborating on "zoom_always"? I am not sure if I understand the issue you are solving yet :)

a-constantin:
In TTS, if you do "zoom_always on" and "zoom_follow_pointer off" in the console, the alt zoom display is always showing in a corner of the screen when you hover an item with your mouse instead of having to hold the alt key, which I find useful, but it breaks right clicking some buttons and this was the simplest way I found to fix it.

Chr1Z93:
Oh, that's interesting! I didn't even know these commands. I can replicate the issue.
Though I am not convinced that this is the proper solution, since there are a lot of other buttons that would malfunction in this mode.
As a super early fix while I am looking into this, let me note that you can also set the state of these counts by hovering an pressing a number.

